Issue #859, opened 2015-02-22
Implement account create/delete hooks.


Olink:
Requested here: https://tshock.co/xf/index.php?threads/to-welcome-the-new-forums-lets-have-some-fun.3427/#post-38024

PR #860, opened 2015-02-22
Add account create & delete hooks

nicatronTg:
This should implement #859.

Looking for someone to sanity check it and make sure it's good, just merge if it does.


Olink:
It might be prudent to load the full User object from the database before deleting it, as well as after creating it for the hooks.  Currently the hooks will be invoked with partial user objects that don't guarantee any sanity of data.

PR #869, opened 2015-02-25
Adds the ability to log to an SQL database

WhiteXZ:


TShockAPI/Commands.cs
```diff
@@ -893,23 +893,26 @@ private static void RegisterUser(CommandArgs args)
...
-                if (TShock.Users.GetUserByName(user.Name) == null && user.Name != TSServerPlayer.AccountName) // Cheap way of checking for existance of a user
+				if (TShock.Users.GetUserByName(user.Name) == null && user.Name != TSServerPlayer.AccountName)
+					// Cheap way of checking for existance of a user
 				{
 					args.Player.SendSuccessMessage("Account \"{0}\" has been registered.", user.Name);
 					args.Player.SendSuccessMessage("Your password is {0}.", user.Password);
 					TShock.Users.AddUser(user);
-					Log.ConsoleInfo("{0} registered an account: \"{1}\".", args.Player.Name, user.Name);
+					TShock.CharacterDB.SeedInitialData(TShock.Users.GetUser(user));
+					TShock.Log.ConsoleInfo("{0} registered an account: \"{1}\".", args.Player.Name, user.Name);
```
nicatronTg: This line is duplicated.

TShockAPI/TextLog.cs
```diff
@@ -0,0 +1,250 @@
...
+	/// </summary>
+	public class TextLog : ILog, IDisposable
+	{
+		private readonly StreamWriter _logWriter;
+		private readonly LogLevel _logLevel;
+
+		/// <summary>
+		/// Log file name
+		/// </summary>
+		public static string fileName;
```
nicatronTg: This should be:

```csharp
public static string fileName { get; private set; }
```

nicatronTg:
After what I've noted has been fixed, looks good to me :+1:.

Do not merge this via PR; I'll rebase it.

TShockAPI/Commands.cs
Olink: Which line?  If this is /register it shouldnt be seeding the database anymore...

nicatronTg: Yeah, that's a good point. Thanks for pointing that out.

PR #870, opened 2015-02-25
DO NOT MERGE Adds SQL logging from WhiteX's fork; closes #869

nicatronTg:
Waiting on yet another okay from @Olink; do not merge with git merge, wait for rebase. Kthx.

### DIFF
--- a/TShockAPI/ILog.cs
+++ b/TShockAPI/ILog.cs
@@ -42,11 +42,6 @@ namespace TShockAPI
         /// </summary>
 		string Name { get; }
 
-        /// <summary>
-        /// Returns true if the ILog is using SQL or false if not
-        /// </summary>
-        bool Sql { get; }
-
 		/// <summary>
 		/// Checks whether the log level contains the specified flag.
 		/// </summary>
--- a/TShockAPI/ILog.cs
+++ b/TShockAPI/ILog.cs
@@ -1,0 +1,140 @@
+﻿/*
+TShock, a server mod for Terraria
+Copyright (C) 2011-2015 Nyx Studios (fka. The TShock Team)
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+using System;
+
+namespace TShockAPI
+{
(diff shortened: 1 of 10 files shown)


TShockAPI/ILog.cs
```diff
@@ -0,0 +1,140 @@
...
+	{
+        /// <summary>
+        /// Log name
+        /// </summary>
+		string Name { get; }
+
+        /// <summary>
+        /// Returns true if the ILog is using SQL or false if not
+        /// </summary>
+        bool Sql { get; }
```
Olink: Remove this.  Its not used anywhere and doesn't make sense to have in a Log implementation.  A user of ILog should never need to know if its a SQL log or not, it should just tell it to do something.

TShockAPI/SqlLog.cs
```diff
@@ -0,0 +1,248 @@
...
+
+			var frame = new StackTrace().GetFrame(2);
+			if (frame != null)
+			{
+				var meth = frame.GetMethod();
+				if (meth != null && meth.DeclaringType != null)
+					caller = meth.DeclaringType.Name;
+			}
+
+			try
```
Olink: We dump 10 (or w/e config is configured to) log entries if sql goes down.  This is probably a bad idea, imo.

TShockAPI/TShock.cs
```diff
@@ -204,16 +188,18 @@ public override void Initialize()
...
+								hostport.Length > 1 ? hostport[1] : "3306",
+								Config.MySqlDbName,
+								Config.MySqlUsername,
+								Config.MySqlPassword
 								);
 					}
 					catch (MySqlException ex)
 					{
-						Log.Error(ex.ToString());
+						Console.ForegroundColor = ConsoleColor.Red;
```
Olink: Use the ServerAPI.Log in the initialization code prior to the TShock log being initialized, so that everything is at least recorded.

nicatronTg:
@WhiteXZ you can just clone the tshock repo locally and checkout the ```whitexz-general-devel``` branch to make these changes. Let me know if you need help.

nicatronTg:
@NyxStudios/tshock-contributors it's worth noting that anyone in the repo can checkout this branch and work on these requests (including myself) from Olink now, since it's in our repo rather than a side branch

Olink:
Its worth noting that I raised several concerns about this feature internally and do not think this should get merged forward until said concerns have been discussed.

WhiteXZ:
Failed queries are now added to a queue.
ILog no longer has an SQL field.
Log.cs no longer has a Dispose method (and thus doesn't dispose TShock.Log).
Startup errors that occur before the log is created are now logged by the ServerAPI logger

Olink:
I'm still pretty concerned with the SQL failure cases.  Still need to have people discuss instead of just throwing spaghetti at the wall and hoping it sticks.

WhiteXZ:
if (_failures.Count > 0) is now while(_failures.Count > 0 && success)

The reverse for loop is now a foreach loop as it should have been after I switched from RemoveAt to Clear

PR #871, opened 2015-02-28
Update deprecated SendData code, a few log methods, add command specifier to possibly everything

Enerdy:
In the first commit, I replaced a few of the old SendData methods which still used SendData(string, Color.Red) and SendData(string, Color.Yellow) instead of SendErrorMessage and SendInfoMessage. I also replaced some Log commands with TShock.Log to reflect WhiteX's changes.

In the second, massive commit, I did the tiresome job of adding the command specifier to pretty much every tshock command. Might be hard to review as there are in fact a lot of commands, but this had to be done at some point.

Still on the second commit, I added the static strings Commands.Specifier and Commands.SilentSpecifier for easier and safer access to those thanks to a null check.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -978,7 +978,7 @@ namespace TShockAPI
 				}
 				catch (UserManagerException ex)
 				{
-					args.Player.SendMessage(ex.Message, Color.Red);
+					args.Player.SendErrorMessage(ex.Message);
 					TShock.Log.ConsoleError(ex.ToString());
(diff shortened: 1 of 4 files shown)


nicatronTg:
Looks good to me :+1:. Unexpected but highly loved changes, actually!

Olink:
Wtf github app, y u no let me merge :(  Looks fine to me

nicatronTg:
@Olink the mobile website I think works better than the app unfortunately :(

PR #881, opened 2015-03-15
Add /permission & methods for adding user permissions

Enerdy:
Users now have a permissionManager object, along with an SQL column (Permissions) to store their permissions. Should we add the permission fields Groups have to Users? (I.E. Group.Permissions, Group.TotalPermissions).

The /permission command encases /group addperm, delperm and listperm with support for users as well. I haven't removed the latter as it is still debatable whether we should encase everything on a /permission command or just put in the permission management subcommands into the /user command. The most interesting feature on this is probably the different colors between the permissions an object actually owns (yellow) and the permissions an object inherits from its parent/group.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -24,10 +24,11 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using Terraria;
-using TShockAPI.DB;
 using TerrariaApi.Server;
+using TShockAPI.DB;
 using TShockAPI.PermissionSystem;
 
 namespace TShockAPI
@@ -282,6 +283,10 @@ namespace TShockAPI
 			add(new Command(Permissions.mute, Mute, "mute", "unmute")
 			{
 				HelpText = "Prevents a player from talking."
+			});
+			add(new Command(Permissions.managepermissions, Permission, "permission")
+			{
+				HelpText = "Manages group and user permissions."
 			});
 			add(new Command(Permissions.savessc, OverrideSSC, "overridessc", "ossc")
 			{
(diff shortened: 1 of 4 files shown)


TShockAPI/DB/UserManager.cs
```diff
@@ -183,6 +185,40 @@ public void UpdateLogin(User user)
             }
 	    }
 
+		public String AddPermissions(String name, List<String> permissions)
```
nicatronTg: The AddPermissions & DeletePermissions methods seem clunky to me. There's no commenting to signify what they do in relation to the existing group based permission system.

Why do they return strings? Why do they not just throw exceptions, and why are they not just booleans for success/fail?

TShockAPI/DB/UserManager.cs
```diff
@@ -183,6 +185,40 @@ public void UpdateLogin(User user)
...
+
+			IPermissionList oldperms = user.permissionManager.GetPermissions(); // Store old permissions in case of error
+			permissions.ForEach(p => user.AddPermission(p));
+
+			if (database.Query("UPDATE Users SET Permissions = @0 WHERE Username = @1;", user.permissionManager, name) == 1)
+				return "User " + name + " has been modified successfully.";
+
+			// Restore old permissions so DB and internal object are in a consistent state
+			user.SetPermission(oldperms.GetPermissions());
+			return "";
```
nicatronTg:
```csharp
return "Looks like something happened that shouldn't have"; //?????
```

TShockAPI/DB/UserManager.cs
```diff
@@ -183,6 +185,40 @@ public void UpdateLogin(User user)
...
+		}
+
+		public String DeletePermissions(String name, List<String> permissions)
```
nicatronTg: Same comment about returning strings here.

TShockAPI/DB/UserManager.cs
```diff
@@ -314,8 +351,9 @@ public class User
 		public string Registered { get; set; }
         public string LastAccessed { get; set; }
         public string KnownIps { get; set; }
+		public IPermissionManager permissionManager;
 
-		public User(string name, string pass, string uuid, string group, string registered, string last, string known)
+		public User(string name, string pass, string uuid, string group, string registered, string last, string known, string permissions = null)
```
nicatronTg: Is there any good reason in general why this constructor takes now 8 string objects when nearly all of these objects have internal data representations that aren't strings?

TShockAPI/Group.cs
```diff
@@ -140,7 +140,7 @@ public Group(string groupname, Group parentgroup = null, string chatcolor = "255
         /// Checks to see if a group has a specified permission.
         /// </summary>
         /// <param name="permission">The permission to check.</param>
-        /// <returns>Returns true if the user has that permission.</returns>
+        /// <returns>Returns true if the group has that permission.</returns>
 		public virtual bool HasPermission(string permission)
```
nicatronTg: My knowledge of C# might be rusty, but why are these methods virtual if they're never overridden?

> The virtual keyword is used to modify a method, property, indexer, or event declaration and allow for it to be overridden in a derived class. --MSDN

TShockAPI/Group.cs
```diff
@@ -140,7 +140,7 @@ public Group(string groupname, Group parentgroup = null, string chatcolor = "255
...
         {
```
nicatronTg: This file has very mixed tab + spaces combinations. While not necessarily a rule, @Olink brought up the idea in conversation a lot that we should be converting these issues as we notice them to tabs, rather than spaces.

TShockAPI/Commands.cs
```diff
@@ -1609,6 +1614,331 @@ private static void TempGroup(CommandArgs args)
...
+				name = match.Groups[4].Value;
+
+			bool all = name == "*";
+			User user = TShock.Users.GetUserByName(name);
+			Group group = TShock.Groups.GetGroupByName(name);
+
+			List<string> permissions;
+			if (string.IsNullOrWhiteSpace(match.Groups[5].Value))
+			{
+				player.SendErrorMessage("You must provide at least one permission. Separate multiple permissions with whitespaces");
```
nicatronTg: The programmer colloquialism "whitespaces" should really never be shown to users. "Separate with spaces" should suffice. There should be a period at the end of this sentence.

TShockAPI/Commands.cs
```diff
@@ -1609,6 +1614,331 @@ private static void TempGroup(CommandArgs args)
...
+			else
+				permissions = match.Groups[5].Value.Split(' ').ToList();
+
+			if (all)
+			{
+				if (IsWhiteSpace(param))
+				{
+					TShock.Groups.ForEach(g => TShock.Groups.AddPermissions(g.Name, permissions));
+					TShock.Users.GetUsers().ForEach(u => TShock.Users.AddPermissions(u.Name, permissions));
+					player.SendSuccessMessage("Modified all groups and users.");
```
nicatronTg: Because users have no log of the commands the run, "all users and groups" would be better suited to say which permissions and groups. Same with all groups & all users.

TShockAPI/Commands.cs
```diff
@@ -1609,6 +1614,331 @@ private static void TempGroup(CommandArgs args)
...
+					player.SendSuccessMessage("Modified all users.");
+				}
+			}
+			else
+			{
+				if (IsWhiteSpace(param))
+				{
+					if (group != null && user != null)
+						player.SendErrorMessage(
+							"Ambiguity between group '{0}' and user '{0}'. Use -g or -u to select either the group or the user", name);
```
nicatronTg: Sentences after periods. Not trying to nitpick but this is easy and a nice touch for consistency.

TShockAPI/Commands.cs
```diff
@@ -1609,6 +1614,331 @@ private static void TempGroup(CommandArgs args)
...
+				name = match.Groups[4].Value;
+
+			bool all = name == "*";
+			User user = TShock.Users.GetUserByName(name);
+			Group group = TShock.Groups.GetGroupByName(name);
+
+			List<string> permissions;
+			if (string.IsNullOrWhiteSpace(match.Groups[5].Value))
+			{
+				player.SendErrorMessage("You must provide at least one permission. Separate multiple permissions with whitespaces");
```
nicatronTg: Pretend all issues with grammar, specificity, and help are here too. This stuff needs to be a powerful replacement for the existing system, not an equivalent quality replacement.

TShockAPI/Commands.cs
```diff
@@ -1609,6 +1614,331 @@ private static void TempGroup(CommandArgs args)
...
+			if (!string.IsNullOrWhiteSpace(match.Groups[2].Value))
+				param = match.Groups[2].Value[0];
+			string name = match.Groups[3].Value;
+			if (string.IsNullOrWhiteSpace(name))
+				name = match.Groups[4].Value;
+
+			// We definitely don't need to print the entire permission system every time someone uses this command
+			if (name == "*")
+			{
+				player.SendErrorMessage("This sub-command doesn't take * as a parameter.");
```
nicatronTg: This error should really be "Refusing to list all permissions in the permission system for space." or something that clarifies why the command failed, not that it did. The fact that it didn't run implies they can't, but why are we refusing to list it?

TShockAPI/DB/UserManager.cs
```diff
@@ -42,7 +43,8 @@ public UserManager(IDbConnection db)
 			                         new SqlColumn("Usergroup", MySqlDbType.Text),
 									 new SqlColumn("Registered", MySqlDbType.Text),
                                      new SqlColumn("LastAccessed", MySqlDbType.Text),
-                                     new SqlColumn("KnownIPs", MySqlDbType.Text)
+                                     new SqlColumn("KnownIPs", MySqlDbType.Text),
```
nicatronTg: The tab/space conflict here also should be fixed in a patch if you're modifying code that is broken.

nicatronTg:
There's an overall security implementation question here too -- all players are equal in this permission system. SuperAdmin A creates, say admin B and moderator C. All have permission modify commands.

I haven't dug too deep into everything, but from what I can tell, if mod C has permission to modify permissions, mod C can revoke everyone of B's permissions, grant himself any permissions he pleases, etc.

Assuming the superadmin still ignores all permission checks (being a superadmin returns true to all checks and ignores all negations), then the superadmin is still safe at least, but this is little reassurance if mod C can give himself group modify permissions, and then elevates himself to superadmin.

When I implemented a system like this at Mercury, my solution was to only allow adding and removing permissions that a user had at their disposal. This means that mod C could only add and revoke permissions that mod C already had.

This doesn't solve the fact that mod C can still revoke all of mod C's powers from admin B. For that, another layer of complication gets added, because immunity should really allow admin B be immune to modification from mod C. This can be done by having an integer immunity level exposed on a per user and per group basis. Great, we're adding more checks here to make things more complicated, but it does secure the system more as far as access control goes.

In this state, mod C can only affect users equal or lesser than mod C with only mod C's permission set. Mod C cannot revoke admin B's powers, nor can they modify admin B's groups. And, best of all, even if the superadmin stuff is broken right now, they can't be touched if their immunity is the highest immunity in the chain (thus, unmodifiable by any implied lower users).

nicatronTg:
>  I haven't removed the latter as it is still debatable whether we should encase everything on a /permission command or just put in the permission management subcommands into the /user command. 

I would move everything into either a permission command or a user command. I don't really care which gets the modification system, only that there is only one way of doing things. Permissions are complicated enough as it is, and if there are multiple ways documented and they both actually work, that's a great nightmare of explaining the functionality differences to future users of the system.

TShockAPI/Commands.cs
Enerdy: My bad, I'm terrible at understanding which parts of the TShock code are "good" or "bad", and as I've seen multiple error messages without periods at the end the convection wasn't quite clear to me.

TShockAPI/DB/UserManager.cs
Enerdy: For those, I simply copied the implementation found at GroupManager.cs with the necessary changes. I have the terrible habit of using as much already built-in TShock code as possible for some reason (I don't trust my coding skills?). Then again I didn't see any problems with returning a string for an error, makes it easier to handle than throwing exceptions, but we definitely should return only one type of message (probably errors), because right now it returns both types and it gets confusing when a player receives an error as a success message.

TShockAPI/DB/UserManager.cs
Enerdy: I decided not to change the way we store data in the User class for this PR. Then again, the parameter I added is one of the few whose data representations is actually a string. :p

TShockAPI/Group.cs
Enerdy: I am well aware of this - yet again, I ripped this off Group.cs but I had reasons to keep it a virtual, as the general consensus for this PR was that nothing here is final and that's why we're commiting to a test branch, and it could be overriten in a future class. Leave it up to the person who implements that to make it virtual?

TShockAPI/Commands.cs
Enerdy: I had trouble deciding how to present this message, so I went for the currently existing implementation in /group addperm. It'd be easy to print which permissions were modified when speaking of individual groups/users, but what about the latter? We should in fact be telling the user what changed, but I believe "all" should still be used, as if the user changes a permission accross a large array of groups at the same time (50+) I'm afraid it might spam them too much (we could use Log.Info here for if they want the full log of changes, though?)

TShockAPI/DB/UserManager.cs
Enerdy: I only noticed this when seeing the diff in github haha. VS's diff doesn't show the difference.

TShockAPI/Group.cs
Enerdy: See above.

TShockAPI/DB/UserManager.cs
Enerdy: This is bad and I feel bad.

TShockAPI/Commands.cs
Enerdy: Noted.
Regarding this specific error message though - I'm not sure if it's clear enough. Do you have a better way of saying this?

TShockAPI/DB/UserManager.cs
nicatronTg: You can take a permission manager here though.

TShockAPI/Commands.cs
nicatronTg: Those are bad; it's okay for a misunderstanding I just would like new code to not be like the old.

TShockAPI/Commands.cs
nicatronTg: Probably make a determination on the result size and say all (50+) or something with how many were changed rather than just all. A large number gives insight as to what was changed even if you don't list all permissions.

TShockAPI/Commands.cs
nicatronTg: I actually thought there wasn't a better way of expressing that other than saying "Target is ambiguous" but that only makes the message sound more technical, not less.

TShockAPI/DB/UserManager.cs
nicatronTg: I use Sublime so I can catch this easily enough. I wonder if we could get a precommit hook installed that would check this for people. That would be cool.

TShockAPI/DB/UserManager.cs
nicatronTg: We've been removing code that passes strings for errors progressively. Yes, you should be throwing exceptions and not passing strings for errors. That means that using this method will either blindly return the results from this message or will have to do string parsing.

Enerdy: I see. Thanks for clarifying that.

Olink:
All new methods need to have xml documentation end of story.  We dont need 500 more warnings.

TShockAPI/DB/UserManager.cs
Olink: No, because iirc we construct the user from a db representation which is all strings.  I think using Group and PermissionsManager classes instead is better, but only if requires little effort. As long as constructors are documented having 100 strings isnt totally bad.

TShockAPI/Group.cs
Olink: SuperAdmin : Group
Its overriden?

nicatronTg: Ahh, ok @Olink. Nvm then.

Olink: Also that makes me want to close this PR right now.  Never tell us that we can fix your bad code in the future because you didn't know how to implement it right the first time.  The whole point of PR is not to get half assed code into the codebase but to have quality code come in so that the real devs don't have as much work to do.

TShockAPI/Group.cs
Olink: It should be mentioned that I only meant modifying areas of a file you have modified, not entire files IE what @nicatronTg has been doing.

TShockAPI/Commands.cs
Olink: Also why are you whitespace delimiting?  Why not comma separated like most other list commands.

Enerdy:
Closing as this needs more work and it's been idling for a while.

PR #891, opened 2015-03-23
SQL log actually ensures its table exists

WhiteXZ:
Because it didn't before. Whoops.

### DIFF
--- a/TShockAPI/SqlLog.cs
+++ b/TShockAPI/SqlLog.cs
@@ -74,9 +74,9 @@ namespace TShockAPI
 				);
 
 			var creator = new SqlTableCreator(db,
-											  db.GetSqlType() == SqlType.Sqlite
-												? (IQueryBuilder)new SqliteQueryCreator()
-												: new MysqlQueryCreator());
+				db.GetSqlType() == SqlType.Sqlite
+					? (IQueryBuilder) new SqliteQueryCreator()
+					: new MysqlQueryCreator());
 			creator.EnsureTableStructure(table);
 		}
 
--- a/TShockAPI/SqlLog.cs
+++ b/TShockAPI/SqlLog.cs
@@ -22,6 +22,7 @@ using System.Data;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using MySql.Data.MySqlClient;
 using TShockAPI.DB;
 
 namespace TShockAPI
@@ -55,7 +56,7 @@ namespace TShockAPI
 		/// <summary>
 		/// Sets the database connection and the initial log level.
 		/// </summary>
-		/// <param name="db"></param>
+		/// <param name="db">Database connection</param>
 		/// <param name="textlogFilepath">File path to a backup text log in case the SQL log fails</param>
 		/// <param name="clearTextLog"></param>
 		public SqlLog(IDbConnection db, string textlogFilepath, bool clearTextLog)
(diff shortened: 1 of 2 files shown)


TShockAPI/SqlLog.cs
```diff
@@ -63,6 +64,20 @@ public SqlLog(IDbConnection db, string textlogFilepath, bool clearTextLog)
...
+			var table = new SqlTable("Logs",
+				new SqlColumn("ID", MySqlDbType.Int32) {AutoIncrement = true, Primary = true},
+				new SqlColumn("TimeStamp", MySqlDbType.Text),
+				new SqlColumn("LogLevel", MySqlDbType.Int32),
+				new SqlColumn("Caller", MySqlDbType.Text),
+				new SqlColumn("Message", MySqlDbType.Text)
+				);
+
+			var creator = new SqlTableCreator(db,
+											  db.GetSqlType() == SqlType.Sqlite
```
nicatronTg: Bad formatting?

WhiteXZ: Yeah not sure what happened there.

TShockAPI/SqlLog.cs
```diff
@@ -232,9 +247,8 @@ public void Write(string message, TraceLevel level)
 					return;
 				}
 
-				_database.Query("INSERT INTO Logs (LogLevel, TimeStamp, Caller, Message) VALUES (@0, @1, @2, @3)",
-					level, DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
-					caller, message);
+				_database.Query("INSERT INTO Logs (TimeStamp, Caller, LogLevel, Message) VALUES (@0, @1, @2, @3)",
```
nicatronTg: By changing the table structure in this way, don't you break backwards compatibility with the already released MySQL log system?

TShockAPI/SqlLog.cs
```diff
@@ -232,9 +247,8 @@ public void Write(string message, TraceLevel level)
...
+					DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture), caller, (int)level, message);
```
bladecoding: Should use DateTime.UtcNow

TShockAPI/SqlLog.cs
WhiteXZ: There is no structural change, just a change to the INSERT statement?
(And seeing as the table was never created previously, there's no backwards compatibility to break)

TShockAPI/SqlLog.cs
WhiteXZ: The text logger and the old TShock log uses and used DateTime.Now

nicatronTg: Time to get with the times and switch all of the code!

Olink: "I checked BanManager and it doesn't.  Get #REKT nerd." - bcccc aka "I'm high6 and I made SQL"

bladecoding: See: https://github.com/NyxStudios/TShock/blob/general-devel/TShockAPI/DB/BanManager.cs#L157

TShockAPI/SqlLog.cs
nicatronTg: @WhiteXZ your previous statement would work if someone manually created a table (such as the way GrandPa-G did).

TShockAPI/TextLog.cs
```diff
@@ -212,7 +212,7 @@ public void Write(string message, TraceLevel level)
 			}
 
 			var logEntry = string.Format("{0} - {1}: {2}: {3}",
-					DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
+					DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
```
nicatronTg: Text log should probably not be in UTC, because it's going to cause issues when people's logs don't line up anymore.

nicatronTg: My bad telling you to change this specific one.

nicatronTg:
Looks good to me. cc @Olink

PR #896, opened 2015-04-08
DEBUG preprocessor checks for ILog

WhiteXZ:
Because TraceLevel does not have a Debug value

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2469,7 +2469,9 @@ namespace TShockAPI
 			{
 				if (type == 100 || type == 164 || type == 180 || type == 261 || (type > 289 && type < 298) || (type >= 325 && type <= 328) || (type >= 345 && type <= 352))
 				{	
+#if DEBUG
 					TShock.Log.Debug("Certain projectiles have been ignored for cheat detection.");
+#endif
 				}
 				else
 				{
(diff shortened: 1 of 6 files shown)


Olink:
I dont quite understand this commit, why didnt you just put the preprocessor conditional inside Debug() so it changed one file in one line.

nicatronTg:
Yeah, that.

WhiteXZ:
This way TShock.Log.Debug doesn't exist unless you're using a debug build, so people won't attempt to use it

Olink:
But people will want to use it?  They just want the function to NOP if debug isnt specified.

WhiteXZ:
Why would they want to use a function that does nothing when DEBUG is not specified?

Olink:
Because we build in release, so they can never physically compile against a build with the debug log statement, which means even if they build for debug they cant use it.  You might as well as delete the entire function and its uses at that point.

nicatronTg:
Because it allows them to not change their codebase for debug...

You :ninja: Zack. That.

Olink:
I would be inclined to agree that in a Release build we don't even need to compile code for debugging, however, since we provide release binaries to the server owners, there is no way for a developer to run a debug build on an actual production environment to understand why something doesnt work, and traditionally, you don't remove functions entirely from a build (unless they serve 0 purpose and are only for your development of your software for your testing), you NOP them.

Olink:
While not related to this PR, can we take a moment to discuss how bad we are at removing deprecation :kappa:

The entire Log.cs file needs to be removed, as its a shim to TShock's Log member.

nicatronTg:
Looks much better @WhiteXZ.

PR #900, opened 2015-04-13
Move password hashing to BCrypt

nicatronTg:
# Why we have a problem now

1. Passwords are not [salted](https://en.wikipedia.org/wiki/Salt_(cryptography)). That means two users with the same password have the same hash. This is also a large problem for password cracking and hash collisions.

2. Our lowest supported algorithm, MD5, [is not a slow algorithm](http://security.stackexchange.com/questions/19906/is-md5-considered-insecure), and rainbow tables for several character lengths already exist.

3. None of our hashing algorithms were designed for hashing passwords. That's why we're moving to BCrypt.

# BCrypt

This is the best possible implementation of BCrypt using [OWASP's guidelines for password storage](https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet#). We could use PBKDF2 or SCrypt, but both are easier to screw up, and SCrypt in particular adds a large amount of memory overhead, while PBKDF2 requires specifying the number of rounds, rather than a work factor. In other words, we would conceivably specify, say, 60,000 rounds of hashing, rather than a work factor of 10 in the config file.

BCrypt automatically generates a salt, and therefore manual, cryptographically secure salt generation is not necessary. Furthermore, the salt and work factor for each hash are stored inside the hash itself, so that the work factor can be determined and solved for each independent hash.

BCrypt's work factor is what makes it future proof. Two years down the line, we don't need to change algorithms again -- we can just upgrade the default work factor or push people to change. The work factor is used to compute the number of rounds required for the hashing algorithm, and therefore the speed required to verify (and more importantly, brute force hashes). As an example, running TShock with a work factor of 12 versus 10 has a noticeable delay in modern hardware. The default work factor has been set to 7, but this can be changed to suit a server owner.

# Changes required

* We no longer allow for specifying a crypto algorithm, but for good reason: if a server is compromised, so are the passwords. If a server owner users MD5, passwords are as good as gone in the event that an attacker can compromise the server.
* TShock no longer exposes its internal hashing system for BCrypt. Because BCrypt requires a work factor to verify, comparison by simply checking for equivalence between a hash and a stored hash is not the conventional way to verify BCrypt hashes. Instead, the User object contains a VerifyPassword method. Plugins that manually verify hashes need to change to support this. This is not an ideal solution, but is the only way to support this right now. Any plugin that is overriding login code is playing with fire anyway, but I can understand that there are edge cases (such as forum authentication).
* VerifyPassword now has several different jobs. To support old passwords, we support verifying that a user is not using an old algo. Because BCrypt passwords are stored in a way that makes them uniquely identifiable as BCrypt passwords (specifically, $2a$workfactor$hash), we can differentiate them from old hashes. In addition, we re-hash a password if its work factor is less than the desired work factor stored in the config. These are operations that can only happen when a user logs in, and is seamless. No config changes are required to support BCrypt.
* The internal TShock methods for verifying passwords have been moved to live in the User object, so that we can always verify ancient passwords by ourselves, even when other plugins can't, for a long time.
* User.Password was stored in such a way that anyone could write and read anything to it. This resulted in several parts of the code storing either a hashed password or a plain text password, which could have been stolen by a rogue plugin by reading the user object at the wrong time. As a result, the password field is now an internal set. It can still be grabbed, but not set externally.
* A User.CreateBCryptHash(string) method has been added, to generate and store the BCrypt hash in the user object.
* A BCryptWorkFactor item has been added to the config file, which is the work factor used by BCrypt. BCrypt requires it to be above 4, and less than 32. We default it to 7; the standards say 10. Because we have a work factor, this hashing algorithm is future proof; changing algorithms again is unnecessary, as we can just upgrade the default work factor later. This instantaneously adds security.

## Deprecations / pitfalls

* TShock.Utils.HashAlgo, TShock.Utils.HashPassword(bytes), and TShock.Utils.HashPassword(string) have all been deprecated in favor of User.VerifyPassword(string). While this is going to be subject to the standard deprecation policy, as soon as a user logs in, their password will be converted and stored as a BCrypt hash. Their password will not be able to be validated by the HashPassword method after this time.
* The HashAlgorithm config option no longer applies to new hashes, but must be kept to read the old ones, if they aren't sha512, the default.

# It's a death march to change, right?

It's a death march to change, yes. However, for users transitioning to this system who have no plugins that replicate the login code, **this change is seamless, and has sensible fallbacks in the event that something is configured wrong**. There are no other ramifications that can happen from this, except for some broken plugins that replicate the login code. **The only reason for a server owner to use an alternative password algorithm is to reduce security. Therefore, there is no reason to let server owners put their users at risk.**

As a bonus, this will likely work on XP, without requiring extra work to specify an alternative hashing algorithm.

# Has this been tested?

- [x] Work factor upgrades okay.
- [X] Passwords convert okay.
- [X] Users register okay.
- [X] Users are added okay.
- [X] Users can't login with an invalid password.
- [X] Users can't register with blank passwords. This was a vestige of IP address users.
- [X] Users have a minimum password requirement.


Enerdy:
Looking pretty good :+1:.

I see no reason why shouldn't we do this, unless there are significant drawbacks. I never really played around with different hash algorithms, nor have I hosted a server to make me worry about password security in the past, but if the current methods are insecure, we should definitely move onto this.

Olink:
They are not insecure, they just arent over the top, which this addresses.  Thechances of someone maliciously hacking a terraria servers mysql db or sqlite file and stealing accounts from it is never going to happen, and if it does it doesnt allow for them to wreck your server anymore than a grief client and a ddos can.

Adding this causes inconsistancies in the database and maintainer hell in the codebase.  As its a relatively unneeded and never requested feature, it simple causes more pain to us over time.


If we do accept this, even with my :-1: Id be a lot happier if all crypto code added was in a BCrypt class in one place maybe under a Crypto namespace, instead of literred in utils, if i misread and it already does this then great.  Id like at least another 24 hours to look at the design implications and try to future proof ot.

nicatronTg:
@Enerdy there are no real drawbacks, except for plugins that implement custom login code.

The current hashing system is, again, flawed. For example, here are two users who have an identical password, in first in BCrypt and then again in SHA512, for reference.

screen shot 2015-04-13 at 1 14 20 pm

screen shot 2015-04-13 at 1 15 21 pm

In addition to removing code from the Utils class, this also encourages people to move more user specific code into the user class, which increases maintainability and readability. All code submitted in this pull request is documented and is the standard implementation.

nicatronTg:
Because that hash was not I trivially googled and found the result.

http://www.md5lab.com/?md5=e12f39b6f3bc9d40b69e412d5a5da2de

screen shot 2015-04-13 at 1 18 01 pm

nicatronTg:
Waiting for input from @MarioE, @Ijwu, @WhiteXZ

Simon311:
I see no reason to use such a secure system, when we have password transmitted to the server in plain text.
Currently, if anyone was targetting user passwords - chances are they'd target anything but hashes.
Besides - if anyone was to have enough access to steal a database - that would probably be console access.
I can imagine few scenarios in which this might prove useful.

nicatronTg:
To more directly counter the claim that the existing system isn't insecure, I'm going to cite [CWE-759](https://cwe.mitre.org/data/definitions/759.html). Password hashing without salt is insecure.

It also falls under OSWAP's Top 10 A2 for 2013, "[Broken Authentication and Session Management](https://www.owasp.org/index.php/Top_10_2013-A2-Broken_Authentication_and_Session_Management)" as well as [A6](https://www.owasp.org/index.php/Top_10_2013-A6-Sensitive_Data_Exposure).

> User authentication credentials aren’t protected when stored using hashing or encryption. See A6.
> Are any old / weak cryptographic algorithms used? (from A6)

@Simon311 your argument makes no sense; a man in the middle attack over the internet to compromise a TShock password is a far more unlikely scenario than a server being hacked, especially with the prevalence of people hosting on XP and Server 2003, both of which are unsupported by Microsoft and have documented security flaws. Your argument is that we shouldn't make it more secure because, well, apparently if some vector isn't as secure as it could be, nothing should be secure. Or am I missing something?

nicatronTg:
screen shot 2015-04-13 at 2 05 50 pm

The fact that a server admin can say this is even more disheartening.

MarioE:
From what I can tell it looks okay. Should definitely wait the 24 hours for further feedback, though.

nicatronTg:
@MarioE appears okay with this; waiting until tomorrow for any more useful comments & critiques from @Olink.

nicatronTg:
Per @Olink's feedback earlier, this request will also add all of the dlls to a consistent place for referencing, so the deploy script can have everything packed nice and together.

- [ ] Add BCrypt library to deploy zip.
- [ ] Move all resources into a resources folder for deploy script.

Olink:
I don't even know if its worth commenting, given the fact that someone had a power hungry ego trip, but I will raise them and face the comment being deleted and him messaging Github to have me blocked from here as well, because any constructive criticism is harassment:
* As mentioned before, this is a good time to move stuff out of Utils, that don't really need to be in there.  When was the last time you obtained a hash by calling into Lib.Utils.HashSomething().  It should be in its own class, Crypto.BCrypt() or another static member to obtain an instance of the algo.
* The encrpytion/hashing/salting whatever that it does now, on default of workfactor 7 is untimed.  While I agree that 1/100 sec is probably not a problem, based on someones arbitrary timings, there is no guarentee someone wont want to use a higher factor.  Our implementation should be smart and not allow for the main thread to be hung by the calculations.  the one line change (ok it might be more, but its trivial) to make this not main threadbound seems like hardly a reason to not implement it.
* Assuming you obsoleted all the hash functions for legacy code, I disagree.  If we are using it internally, we shouldnt be deprecating it given the discussion weve had where people will have legacy passwords forever.  At worst we should rename them to something better, such as Hash(), or Encrypt(), and that would result in plugins using them to break, and hopefully cause end users to not use it, Deprecating it without removing it is what we said you weren't going to do anymore.

Given the animosity @nicatronTg has for me, I doubt that this even matters, but I would like to have til tomorrow to actually look at the implementation.  This is a relatively large reaching change and I can't just look it over in a few hours.

TShockAPI/Commands.cs
```diff
@@ -942,7 +960,7 @@ private static void RegisterUser(CommandArgs args)
                 if (TShock.Users.GetUserByName(user.Name) == null && user.Name != TSServerPlayer.AccountName) // Cheap way of checking for existance of a user
```
Olink: some indentation issue

TShockAPI/Commands.cs
```diff
@@ -976,7 +994,14 @@ private static void ManageUsers(CommandArgs args)
 				var user = new User();
 
 				user.Name = args.Parameters[1];
-				user.Password = args.Parameters[2];
+				try {
```
Olink: brace is on same line

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+					return true;
+				}
+				return false;
+			}
+			return false;
+		}
+
+		/// <summary>Upgrades a password to BCrypt, from an insecure hashing algorithm.</summary>
+		/// <param name="password">string password - the raw user password (unhashed) to upgrade</param>
+		protected internal void upgradePasswordToBCrypt(string password)
```
Olink: Lower Case function naming is a no-no.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+				this.Password = oldpassword; // Revert changes
+			}
+		}
+
+		/// <summary>Upgrades a password to the highest work factor available in the config.</summary>
+		/// <param name="password">string password - the raw user password (unhashed) to upgrade</param>
+		protected internal void upgradePasswordWorkFactor(string password)
+		{
+			// If the destination work factor is not greater, we won't upgrade it or re-hash it
+			int currentWorkFactor = Convert.ToInt32((this.Password.Split('$')[2]));
```
Olink: This should really use int.TryParse() as this can throw an exception in the event that a database has an invalid salt.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+				return ret.Aggregate("", (s, b) => s + b.ToString("X2"));
+			}
+		}
+
+		/// <summary>
+		/// Returns a hashed string for a given string based on the config file's hash algo
+		/// </summary>
+		/// <param name="password">string to hash</param>
+		/// <returns>string hash</returns>
+		protected internal string hashPassword(string password)
```
Olink: Lower case naming...

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+					{"sha256-xp", () => SHA256.Create()},
+					{"md5-xp", () => MD5.Create()},
+			};
+
+		/// <summary>
+		/// Returns a hashed string for a given string based on the config file's hash algo
+		/// </summary>
+		/// <param name="bytes">bytes to hash</param>
+		/// <returns>string hash</returns>
+		protected internal string hashPassword(byte[] bytes)
```
Olink: Lower case naming...

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+			{
+				throw new ArgumentOutOfRangeException("password", "Password must be > " + TShock.Config.MinimumPasswordLength + " characters.");
+			}
+			this.Password = BCrypt.Net.BCrypt.HashPassword(password.Trim(), workFactor);
+		}
+
+		/// <summary>
+		/// A dictionary of hashing algorithms and an implementation object.
+		/// </summary>
+		protected internal readonly Dictionary<string, Func<HashAlgorithm>> HashTypes = new Dictionary<string, Func<HashAlgorithm>>
```
Olink: Lower case naming...

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+		}
+
+		/// <summary>
+		/// Returns a hashed string for a given string based on the config file's hash algo
+		/// </summary>
+		/// <param name="password">string to hash</param>
+		/// <returns>string hash</returns>
+		protected internal string hashPassword(string password)
+		{
+			if (string.IsNullOrEmpty(password) || password == "non-existant password")
```
Olink: I know you are going to say this is outside the scope of your pull request, but thats not really a reason not to make the change anyways, and making it to the main branch could result in merge land which is a pain, but this magic string we have here should be a const somewhere, probably in UserManager or User.

TShockAPI/Commands.cs
```diff
@@ -919,16 +919,34 @@ private static void RegisterUser(CommandArgs args)
...
 			{
 				var user = new User();
-
+				string echoPassword = "";
 				if (args.Parameters.Count == 1)
 				{
 					user.Name = args.Player.Name;
-					user.Password = args.Parameters[0];
+					echoPassword = args.Parameters[0];
+					try
```
Olink: This try/catch block is repeated multiple times throughout this file, consider making a private helper function that does all this on one place and call it in these places such that we dont have the same code repeated that would require multiple changes across the file if we wanted to change anything.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+		/// <param name="password">string password - The password to check against the user object.</param>
+		/// <returns>bool true, if the password matched, or false, if it didn't.</returns>
+		public bool VerifyPassword(string password)
+		{
+			try
+			{
+				if (BCrypt.Net.BCrypt.Verify(password, this.Password)) 
+				{
+					// If necessary, perform an upgrade to the highest work factor.
+					upgradePasswordWorkFactor(password);
```
Olink: You are recalculating the BCrypt every time a user attempts to log in?  I apologize, the usage of the library is very undocumented.  Why not use the same logic above to ensure that the workfactor is what it needs to be?

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+					TShock.Log.ConsoleError(e.ToString());
+				}
+			}
+		}
+
+		/// <summary>Creates a BCrypt hash for a user and stores it in this object.</summary>
+		/// <param name="password">string password - the plain text password to hash</param>
+		public void CreateBCryptHash(string password)
+		{
+			if (password.Trim().Length < Math.Max(4, TShock.Config.MinimumPasswordLength))
```
Olink: Trimming the password twice instead of a local variable?  Interesting choice.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+			catch (UserManagerException e)
+			{
+				TShock.Log.ConsoleError(e.ToString());
+				this.Password = oldpassword; // Revert changes
+			}
+		}
+
+		/// <summary>Upgrades a password to the highest work factor available in the config.</summary>
+		/// <param name="password">string password - the raw user password (unhashed) to upgrade</param>
+		protected internal void upgradePasswordWorkFactor(string password)
```
Olink: Protected implies a subclass, as we do not have any subclasses of User, this should just be private.

TShockAPI/DB/UserManager.cs
Olink: Protected implies a subclass, as we do not have any subclasses of User, this should just be private.

TShockAPI/DB/UserManager.cs
Olink: Protected implies a subclass, as we do not have any subclasses of User, this should just be private.

TShockAPI/DB/UserManager.cs
Olink: Protected implies a subclass, as we do not have any subclasses of User, this should just be private.

TShockAPI/DB/UserManager.cs
Olink: Protected implies a subclass, as we do not have any subclasses of User, this should just be private.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
...
+		public string Password { get; internal set; }
+		public string UUID { get; set; }
 		public string Group { get; set; }
 		public string Registered { get; set; }
-        public string LastAccessed { get; set; }
-        public string KnownIps { get; set; }
+		public string LastAccessed { get; set; }
+		public string KnownIps { get; set; }
 
 		public User(string name, string pass, string uuid, string group, string registered, string last, string known)
```
Olink: This entire class is missing legacy comments, a suggestion to make the 1k warnings go away is to fix these in areas we are working in.

TShockAPI/DB/UserManager.cs
```diff
@@ -308,34 +311,201 @@ public class User
 	{
 		public int ID { get; set; }
 		public string Name { get; set; }
-		public string Password { get; set; }
-        public string UUID { get; set; }
+		public string Password { get; internal set; }
```
Olink: From bc: "Internal means public to the assembly that it is marked Internal in.  This means that TShock can assign to this, however, plugins can not."  If I understood the commit correctly, your intention was to make this field only assignable from within User, as to control who is setting it (specifically the verify functions).  I apologize if I misunderstood the intention of marking this internal.

Olink:
Id say ive now looked over the commits.  I think my three main points above are still a concern.  Ive also pointed out several inconsistent style issues.
To reiterate my points above:
* Future proofing our encryption algo sounds like a smart move if we decide to switch to a new implementation we will have to duplicate a lot of code.  I think moving algos to its own file and interface allows for easy addition and substitution down the road.  While the interface changes i admit are a bit of a stretch in the immediate area, moving algo specific calls to its own file outside of user is the correct choice.  Single responsibility principle.  http://en.wikipedia.org/wiki/Single_responsibility_principle  Currently user provides a responsibility of accessing the algos, and representing the user object in the software.
* With the potential of having multiple clients hitting at the encryption algos on the main thread multiple times, this could inadvertently result in deadlocking a server by some aggressive users.
* This point is relatively opinionated and honestly I don't see a problem either way.  Obsoleting code we are still using is going to generate warnings.  Obsoleting them breaks our new found policy on promptly removing deprecated functions.  Not obsoleting them implies we intend on keeping them around.  

I still think this feature is clunky, but I can't really disagree that our current implementation isnt as bad.  I still vote :-1: because of the lack of design considerations, and would be willing to flip it to a :+1: assuming we can find a middle ground.

nicatronTg:
screen shot 2015-04-13 at 11 42 43 pm

@Olink "This API supports the .NET Framework infrastructure and is not intended to be used directly from your code."

nicatronTg:
Oops, selected wrong one. Mistake.

Olink:
http://puu.sh/hdmKT/3620bd0688.png

nicatronTg:
Considering the fact that the pull request works, code wise, and the only remaining issues appear to be minor tweaks that can be slipped in the general-devel branch later, I'm limiting this conversation to BCrypt specific implementation issues and the remaining deploy script changes.

WhiteXZ:
fwiw choosing to slip minor tweaks in later sounds like a good way to never do the tweaks.
My opinion is do 'em while the PR is open, get it out of the way. That way we don't have more things that need fixing up later on down the road.

Other than that, if everyone is documented properly then I'm okay with this.

nicatronTg:
@WhiteXZ the "minor tweaks" don't appear to be crypto related, and with 30+ commits in this PR, it's getting a bit sprawling. Nothing here is broken, non-working, et cetera.

I mean, I could go and reformat all of the user file to have the proper indentation, but that would also have the nasty side effect of screwing up the diff. The protection settings were intentional; nothing, outside of the User class, should ever be calling those internal functions. This includes anyone who derives a subclass outside of this assembly or inside it.

nicatronTg:
@WhiteXZ sorry to keep piling on here, but I also want to add that things like future proofing in a crypto class are really irrelevant, because again, the point of BCrypt is to not have to switch (increasing the work factor is the mechanism by which this is possible).

Using a variety of different parse mechanisms is also debatable, but this is the password hash, stored in the database, that's protected and internal set only by this function at any time. Have we ever supported a user messing around inside the database and mucking up random data?

MarioE:
Just FYI, protected internal allows subclasses *OR* anything inside of the assembly to access it.

stevenh:
This will cause major issues for hosters, so while if this was in to start with it would be good it doesn't seem to make sense at this time :-1:

nicatronTg:
@stevenh care to explain?

nicatronTg:
@stevenh to clarify, what are you doing with either the API or the database that will break your implementation that you cannot implement?

* If you're resetting user passwords, your current method will work, because we convert old hashes to the new ones on login.
* If you're verifying user information, BCrypt can be added to any number of frameworks in any number of languages. This release is unlikely to go live for a week or more -- in that case, it should be trivial to update your code until then.

stevenh:
When servers are setup, they will create the default accounts that the admin (customer) can access.

If you change the way this is done then all control panels will need to be updated.

This makes implementation server version dependent, which is unfortunately a hard problem to solve.

I guess you could rely on the converting of old hashed to new ones on login assuming that works every time and is not a one time thing?

nicatronTg:
@stevenh yes, you could rely on using the old hash every time, because we only convert when you join, and subsequently, we have to support this for any age password.

So, your existing implementation, while inelegant, will work.

PR #911, opened 2015-04-29
Adds a User object to TSPlayer

WhiteXZ:
The User object is loaded into the TSPlayer on login and set to ```null``` on logout.
TSPlayer.UserAccountName and TSPlayer.UserID are now deprecated and point to User.Name and User.ID respectively.

Also fixes some tabbing issues encountered along the way


nicatronTg:
cc. @NyxStudios/people @NyxStudios/tshock-contributors

PR #913, opened 2015-04-30
User Permissions

Enerdy:
This PR is not yet finished, but I'm pushing it at @nicatronTg's request. My next todo is to address User IDs, as every user should have an ID at all times. Discuss.


Enerdy:
Would it be wise to close this PR and commit to new-permissions so that others can contribute, too? Just applying the concepts of Github Flow, which seems to be what we'll be using from now on.

nicatronTg:
That's kinda up to you, but for the record, it doesn't currently build.
Because it's a pull request from one branch to another non general-devel
branch, I don't see a reason why not.

Enerdy:
Now that the build has been unbroken, I've merged this Pull Request into NyxStudios:new-permissions so that we may follow github flow guidelines. NyxStudios:new-permissions will eventually be merged with general-devel once it is ready for deployment.

nicatronTg:
Thanks @Enerdy!

Issue #908, opened 2015-04-27
Feature Freeze / 1.3 Code Cleanup / Defeat the Kishin


nicatronTg:
![](http://dl.dropbox.com/u/1253613/KishinMadness2_zps681ca6c5.png)
This started as an idea I presented in Slack. Here's the basic premise of what I want to do before 1.3 hits:

1. Improve the codebase significantly using best practices.
2. Improve the documentation of the codebase, so that we generate zero warnings on compiling TShockAPI.csproj (this is not the API, so it should be possible).
3. Update the wiki with any changes we make in doing so.

In order to accomplish this, we're going to need some effort to list out some code quality targets that we want to meet, aside from the goals above. Some preliminary targets:

* AnythingManager - There are a lot of "Manager" classes that do a variety of database calls and use classes like "Group" and "User" as simple datastores. These classes are basically assorted SQL functions running and executing, returning errors as strings or throwing exceptions depending on the implementation date at will.
* Many of TShock's hooks - There are a lot of times when TShock hooks the API, then does around 20 different, completely isolated things inside the hooks. For example, when do we setup the config file? When do we initialize the log file? Those are all stuck inside similar giant hook objects.

Lastly, we need to put a freeze on the creation of new feature sets. Right now, this is planned to start after ```new-permissions``` is merged in by @Enerdy. In other words: no new functionality that is added on top of existing layers for the sake of adding new functionality. Log file size splitting, for example, would be nice, but should only be added if it comes as a side benefit from switching to Log4net during the code cleanup. New commands should also be avoided, while things like the config file are fine to modify (because simplifying or modifying subsystems might grant the ability to turn them off or modify them). This will be lifted once this issue is closed.

nicatronTg:
Please report in if you want tasks assigned to you for the cleanup. If you present an issue and want to tackle it, you get that task instead of getting one randomly assigned. cc. @NyxStudios/tshock-contributors @NyxStudios/people

WhiteXZ:
Some things to decide upon so we have 1 style in the codebase:
* use var or type. E.g., `var player = args.Player` or `TSPlayer player = args.Player`
* private/public/property naming. Re# suggests `private _name` `public name` `Name {get;set;}`. Re# is not god so we should go with whatever everyone is most comfortable with
* Documentation - We currently have some things documented in a `///Name - Description` format and others in a `///Description` format.
[...]

PR #918, opened 2015-05-01
Add new code style guidelines

nicatronTg:
These are the code style guidelines from the consensus discussion from #908. Please sign off @Ijwu, @Enerdy, @WhiteXZ, @tylerjwatson, or suggest more changes.

Sign-off list:

- [x] @Enerdy 
- [x] @WhiteXZ 
- [ ] @Ijwu
- [ ] @tylerjwatson
- [x] @MarioE

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,9 +20,13 @@ Required:
 - Use UpperCamelCase for public function names.
 - Prior to developing, make sure your clone is up to date with general-devel. This means that we don't get merge commits in your pull request.
 - Document all code with public access at minimum. Use the ```<param name="Name">Description</param>``` style -- do not repeat the param type or name in the xml comment field.
+- Document functions that can throw exceptions with ```<exception cref="NameOfException"/>```.
 - Use explicit type declaration (```int, long, float```) when the type is not easily inferred (such as return types).
 - Use implicit type declaration (```var```) when the type is easily inferred (such as private lists, temporary values, etc).
 - Use Microsoft code conventions on variable naming (including ```_name``` for private members).
+- When using static methods on primitives, use the CLR type. E.g. ```String.Format``` instead of ```string.Format```.
+- Always use properties, not public fields.
+- Document deprecations and fail compilation if they're included with ```[Obsolete("Use blah instead of blahx...", true)]```.
 
 ### Dev Team Guidelines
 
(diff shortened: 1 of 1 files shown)


Enerdy:
I'm still not 100% confident with using `_field` for private fields instead of simply lower-casing them. I thought microsoft used `m_field`? Not that the latter is much better. We should simply use lower-case naming for private fields IMO.

With that being said I like the other changes, and I'm willing to sign this off even if we go with `_field`; I've used this in the past and it won't be hard for me to adapt.

nicatronTg:
I really dislike the ```m_field``` convention @Enerdy, and I'll agree with you in that I think that private fields are fine lowercased. I'm going to try again and cc @NyxStudios/tshock and see if more people want to provide feedback on this.

WhiteXZ:
``` m_field``` is horrible. 
I've been using ```_field``` in Orion

nicatronTg:
I suppose the benefit to using ```_field``` is that it's a lot more noticeable that it's private than a lowercase versus uppercase letter.

nicatronTg:
@MarioE got any feedback on this?

MarioE:
Definitely use ```_field``` for privates. That's all my input on naming conventions. Will give style input shortly.

nicatronTg:
Seems like consensus is ```_field``` which lines up with the rest of this change. I'm going to wait on @MarioE for additional feedback on the rest of the style changes.

MarioE:
I will update this list as needed.

* Always use properties instead of public fields.
* When using static methods on primitives, use the CLR type. Use ```String.Format``` instead of ```string.Format```.
* Use ```[Obsolete(true)]``` to mark something as obsolete **and** give a compiler error when attempting to use the obsolete item.
* If a method or constructor can throw an exception, please document it with ```<exception cref="...Exception"/>```.

nicatronTg:
@MarioE can you clarify something for me? Marking something ```[Obsolete("Please ...", true)]``` will still allow it to work in existing code, but will throw a compiler error for new code, correct?

MarioE:
That's correct.

nicatronTg:
Cool. I agree with those then.

nicatronTg:
@MarioE I've pushed the latest version; are you on board? I'm okay with this version, but I've unchecked @WhiteXZ and @tylerjwatson for the time being to see if they're still okay with this.

Enerdy:
Cool. I'm fine with those now.

nicatronTg:
@MarioE @Enerdy @WhiteXZ feel free to merge whenever.

Issue #923, opened 2015-05-03
Update Binaries


MarioE:
Update all binaries to their latest versions (MySql.Data.dll, Newtonsoft.Json.dll, etc).

Strinkor:
:+1:

nicatronTg:
Done in 786a43b2f989215497f44b7fa1b01c410c86a071.

PR #926, opened 2015-05-04
Add new prebuilts

nicatronTg:
This adds new prebuilts for:

* SQLite
* MySQL
* Newtonsoft's JSON libraries

We build our own HttpServer (in this org) that should probably be updated, but I'm not that familiar with it (Steven Hartland did that).

BCrypt should still be up to date since I only added it fairly recently.

Also, all assemblies are for .NET 4.5. Also, it compiles on Mono.

This should satisfy #923, but I don't know if you want to recompile & pull changes for the HttpServer or not @MarioE?


WhiteXZ:
Does this mean we now need to add .NET 4.5 as a requirement for running TShock (when 5.0 is released)?

nicatronTg:
I was under the impression that it already was, so no?

MarioE:
HttpServer doesn't really need to be updated.

nicatronTg:
@MarioE if it looks good to you, feel free to merge.

PR #937, opened 2015-06-07
Add opt-in config to enable sending stats to stats.tshock.co

finnjohnsen:
"phone home" functionality should be in my opinion be disabled pr default. The property name is trying to encourage the admins to enable this. EnableSendStatsToTShockToHelpTheProject

### DIFF
--- a/TShockAPI/ConfigFile.cs
+++ b/TShockAPI/ConfigFile.cs
@@ -433,6 +433,9 @@ namespace TShockAPI
 		[Description("The minimum password length for new user accounts. Minimum value is 4.")]
 		public int MinimumPasswordLength = 4;
 
+		[Description("Enable to send stats to stats.tshock.co")]
+		public bool EnableSendStatsToTShockToHelpTheProject = false;
+
 		/// <summary>
 		/// Reads a configuration file from a given path
 		/// </summary>
--- a/TShockAPI/StatTracker.cs
+++ b/TShockAPI/StatTracker.cs
@@ -21,11 +21,12 @@ namespace TShockAPI
 
 		public void Initialize()
 		{
-			if (!initialized)
+			if (!initialized && TShock.Config.EnableSendStatsToTShockToHelpTheProject)
 			{
-				initialized = true;
+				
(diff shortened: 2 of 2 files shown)


Ijwu:
Looks good. Only issue I can see is this edge-case scenario:

1. We add `StatTracker` re-initialization to the `/reload` command.
2. Server owner _disables_ the tracker at first.
3. Server owner _re-enables_ the tracker and uses `/reload`.
4. The tracker is still disabled due to the `initialized` bool being set to true so it never actually re-initializes.

Seeing as that's edging on occam's razor level edge-case, I'm alright with this PR. :+1:

finnjohnsen:
When I download and use any software, I find it a violation of my privacy if it phones home uploading my personal usage statistics without me explicitly allowing it.

Ijwu:
The issue is that you're essentially disabling the stat tracker. I don't know how useful @nicatronTg is finding the data from it, it may be his call. 

It's a case of "sorry, can't make everyone happy". Do we cripple our stat tracker to alleviate privacy concerns? Or should we just provide an opt-out ability?

I'd say that decision lies in the hands of @nicatronTg and other contributors with more information behind the stats collected, the usage of the stats, and the tracker itself.

nicatronTg:
I'm against this. If anything, I would prefer it to be opt-out rather than opt-in. Right now, the stats collected are used for just enumerating over the active servers and nothing more. We've traditionally never had a opt-out at all for this, and nobody (except you) has complained.

WhiteXZ:
I also agree an opt-out would be better than an opt-in, however I also think this should be part of the 5.0 release or Orion (if we plan to keep a stat tracker in 5.0/Orion) rather than in the current code base.

Enerdy:
I agree with both @white and @nicatronTg. The StatTracker doesn't send any sensible data anyways, and it would break our statistics, but if it bothers you I suppose it could be opt-out. This is better than opt-in because it means people will have it enabled by default, and will only disable it if they don't want to be sending data over, minimizing the impact (as I don't think anyone would bother with it if it was opt-in).

PR #954, opened 2015-07-06
Adding exception at used REST port problem

Patrikkk:


WhiteXZ:
cc @nicatronTg @Enerdy @MarioE @tylerjwatson

TShockAPI/Rest/Rest.cs
```diff
@@ -79,12 +79,20 @@ public Rest(IPAddress ip, int port)
 
 		public virtual void Start()
 		{
-			if (listener == null)
```
Enerdy: What happened to this? Is it always null or are there other reasons that deprecate this statement?

nicatronTg:
Looks good to me!

Enerdy:
:+1:

PR #997, opened 2015-07-12
Fix HandleSpawnBoss

Ijwu:
Update spawnboss to handle new bosses and legions as well as fix packet structure.


nicatronTg:
Sure thing packet guy.

Ijwu:
sorry i forgot we did stuff with packets, whodathunk?

PR #1016, opened 2015-07-16
Serverpassword syntax error gave wrong command

JordyMoos:
The syntax error message for serverpassword said the command was "/password <new password>" but it should be "/serverpassword ..."


WhiteXZ:
If you could switch to using tabs instead of spaces and fix the indentation, I'll merge this in

PR #1030, opened 2015-07-18
Move most startup arguments to TShock from API

Celant:


TShockAPI/TShock.cs
```diff
@@ -533,47 +533,153 @@ private void HandleCommandLine(string[] parms)
...
-						break;
+						{
+							ConfigFile.DumpDescriptions();
+							Permissions.DumpDescriptions();
+							ServerSideConfig.DumpDescriptions();
+							RestManager.DumpDescriptions();
+							Environment.Exit(1);
+							break;
+						}
+					case "-config":
```
Enerdy: We should probably get rid of this all together in the future, or simply deprecate this in favor of `serverconfig` (remove this and rename -serverconfig to -config?). As long as we have the same options in our config than in vanilla config, there should be no need for two config files.

Enerdy:
Might need someone to rebase the unnecessary 5 commits once this gets in. Otherwise, :+1:.

nicatronTg:
This is being accepted because the other one was accepted.

Issue #1022, opened 2015-07-16
Does not have projectile permission to update projectile. (465)


niek1414:
(After getting hit?) by the cultist all non-admin players get "disabled".
The console message: Player ** has been disabled for Does not have projectile permission to update projectile. (465).
After respawn of the cultist the problem still exists. This happened multiple times (3) in the same session.
It did not matter if the user is logged in as a registered user.

Patrikkk:
Thanks for the report, this will be fixed in the next release :)

PR #1033, opened 2015-07-18
Fix more projectile permission checks.

Enerdy:
I didn't commit this immediately as I wanted to give others the chance to commit additional ProjectileIDs.

As of current, I only added the projectile mentioned in #1022.


Ijwu:
Looks good to me.

PR #1118, opened 2015-09-17
SSC Saving over playes with Ignore SSC perms in group or temp group

tysonstrange:
I've added these 2 changes to my server to avoid users who are in temp groups or been given the ignore ssc perms AFTER login from having their SSC inventory saved over.

Related issue:
https://github.com/NyxStudios/TShock/issues/1112


Ijwu:
Other than that one line I pointed out, I'm ok to +1 this. cc @tylerjwatson for confirmation.

tylerjwatson:
Will change message in future patch

PR #1120, opened 2015-09-23
Added rate limiting to RESTful API

Celant:
Added token bucket and timer to degrade token bucket to prevent flood attacks

### DIFF
--- a/TShockAPI/Rest/Rest.cs
+++ b/TShockAPI/Rest/Rest.cs
@@ -22,6 +22,7 @@ using System.ComponentModel;
 using System.Net;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Reflection;
 using HttpServer;
 using HttpServer.Headers;
@@ -44,21 +45,24 @@ namespace Rests
 		public IParameterCollection Parameters { get; private set; }
 		public IRequest Request { get; private set; }
 		public SecureRest.TokenData TokenData { get; private set; }
+		public IHttpContext Context { get; private set; }
 
-		public RestRequestArgs(RestVerbs verbs, IParameterCollection param, IRequest request)
+		public RestRequestArgs(RestVerbs verbs, IParameterCollection param, IRequest request, IHttpContext context)
 		{
 			Verbs = verbs;
 			Parameters = param;
 			Request = request;
 			TokenData = SecureRest.TokenData.None;
+			Context = context;
 		}
 
-		public RestRequestArgs(RestVerbs verbs, IParameterCollection param, IRequest request, SecureRest.TokenData tokenData)
+		public RestRequestArgs(RestVerbs verbs, IParameterCollection param, IRequest request, SecureRest.TokenData tokenData, IHttpContext context)
 		{
 			Verbs = verbs;
 			Parameters = param;
 			Request = request;
 			TokenData = tokenData;
+			Context = context;
 		}
 	}
 	public class Rest : IDisposable
@@ -66,6 +70,8 @@ namespace Rests
 		private readonly List<RestCommand> commands = new List<RestCommand>();
 		private HttpListener listener;
 		private StringHeader serverHeader;
+		public Dictionary<string, int> tokenBucket = new Dictionary<string, int>();
+		private Timer tokenBucketTimer;
 		public IPAddress Ip { get; set; }
 		public int Port { get; set; }
 
@@ -84,6 +90,11 @@ namespace Rests
 				listener = HttpListener.Create(Ip, Port);
 				listener.RequestReceived += OnRequest;
 				listener.Start(int.MaxValue);
+				tokenBucketTimer = new Timer((e) =>
+				{
+					DegradeBucket();
+				}, null, TimeSpan.Zero, TimeSpan.FromMinutes(Math.Max(TShock.Config.RESTRequestBucketDecreaseIntervalMinutes, 1)));
+
 			}
 			catch (Exception ex)
 			{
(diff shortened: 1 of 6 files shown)


TShockAPI/Rest/Rest.cs
```diff
@@ -194,6 +224,23 @@ protected virtual object ProcessRequest(object sender, RequestEventArgs e)
...
 
+					int tokens = 0;
+					if(tokenBucket.TryGetValue(e.Context.RemoteEndPoint.Address.ToString(), out tokens))
+					{
+						if(tokens >= 5)
+						{
+							tokenBucket[e.Context.RemoteEndPoint.Address.ToString()] += 1; // Tokens over limit, increment by one and reject request
+							return new RestObject("403")
+							{
+								Error = "Rate limited. Too many REST requests made."
```
nicatronTg: This should probably tell server owners that the IP has been rate limited due to invalid login attempts.

tylerjwatson: I would not tell the remote client that they have hit a threshold because they can allude to hitting a limit and then can handle such situations. Return a generic unhelpful error in this case, and log an error locally so that brute attempts are locatable by server owners

nicatronTg: Why is this code here and not where actual processing of usernames and passwords takes place?

nicatronTg:
Limits all endpoints, which would throttle legitimate API use significantly with no user configuration for expiration and limit. No go.

nicatronTg:
I don't want to have to post this twice, so I'll post here rather than slack @Celant.

> I'm going to add configuration to it tomorrow, providing the user can configure the limits is it okay for it to stay as a limit to all end points?

That doesn't solve the problem at all. The problem at hand is bruteforcing the password authentication endpoint. A modern REST API has a very high rate limit. Github has 60 per hour for unauthenticated users, and 5,000 per hour with authenticated requests. It also provides the current rate limit back in every reply, so an application can throttle itself accordingly. https://developer.github.com/v3/#rate-limiting

The solution would be to move this into the login code, remove any error messages pertaining to a rate limit, and remove the silly line that means that if you make more than 11 requests, you get exactly the same treatment as someone who attempts 10 or less (if anything it should be logarithmic in how much time is required to remove a token, with the more tokens added to the pool, the greater the time).

Celant:
Justified feedback. Whilst I agree for general use the limit is currently
way too low, I don't think it should be entirely removed. What do you think
to user configurable limits, and a second limit for logins? Done from the
same place but using a custom check if the endpoint is login?

On Wednesday, 23 September 2015, Lucas Nicodemus <notifications@github.com>
wrote:

> I don't want to have to post this twice, so I'll post here rather than
> slack @Celant <https://github.com/Celant>.
>
> I'm going to add configuration to it tomorrow, providing the user can
> configure the limits is it okay for it to stay as a limit to all end points?
>
> That doesn't solve the problem at all. The problem at hand is bruteforcing
> the password authentication endpoint. A modern REST API has a very high
> rate limit. Github has 60 per hour for unauthenticated users, and 5,000 per
> hour with authenticated requests. It also provides the current rate limit
> back in every reply, so an application can throttle itself accordingly.
> https://developer.github.com/v3/#rate-limiting
>
> The solution would be to move this into the login code, remove any error
> messages pertaining to a rate limit, and remove the silly line that means
> that if you make more than 11 requests, you get exactly the same treatment
> as someone who attempts 10 or less (if anything it should be logarithmic in
> how much time is required to remove a token, with the more tokens added to
> the pool, the greater the time).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NyxStudios/TShock/pull/1120#issuecomment-142467151>.
>


--

nicatronTg:
The only problem with implementing general limiting is that it can't be implemented in this API revision. It would have to be on a new version, which is something that is pegged for an upgrade with the move to Orion.

If we were to implement an API rate limiting system, it should have similar defaults to Github's (and same behavior in sending back the number of requests remaining and the reset time to clients).

Celant:
I see, so as a recap: Change it to limit the login endpoint only, make the
limits configurable, and remove the high end cap on tokens. If I'm missing
anything let me know else this'll be fixed tomorrow morn.

On Wednesday, 23 September 2015, Lucas Nicodemus <notifications@github.com>
wrote:

> The only problem with implementing general limiting is that it can't be
> implemented in this API revision. It would have to be on a new version,
> which is something that is pegged for an upgrade with the move to Orion.
>
> If we were to implement an API rate limiting system, it should have
> similar defaults to Github's (and same behavior in sending back the number
> of requests remaining and the reset time to clients).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NyxStudios/TShock/pull/1120#issuecomment-142468141>.
>


--

nicatronTg:
Needs changelog update @Celant otherwise lgtm; cc @NyxStudios/tshock please code review asap to get this fix live.

nicatronTg:
lgtm

Ijwu:
Can we get these errors changed to me more ambiguous or secure. Changing all of the account related errors to something akin to "Username or password may be incorrect or this account may not have sufficient privileges."

https://github.com/Celant/TShock/blob/66e55b3a5cb1a9b56ec49427772399446b23b16b/TShockAPI/Rest/SecureRest.cs#L173

Also, slightly unrelated, is it just me or does this seem like TMI: https://github.com/Celant/TShock/blob/66e55b3a5cb1a9b56ec49427772399446b23b16b/TShockAPI/Rest/Rest.cs#L236

As an API consumer, you don't need a stack trace.

nicatronTg:
I agree with @ijwu's issues, actually. Holding off on my okay until those are resolved.

stevenh:
If I'm reading this right it limits requests per interval for **all** requests.

If we're trying prevent brute force attacks, this should only be done only requests which fail auth.

Taking this approach would prevent locking out legitimate connections as 5 per min is quite low..

nicatronTg:
Do merge at your will @WhiteXZ @Ijwu

Celant:
Rebased at @nicatronTg's request. Merge when ready.

PR #1121, opened 2015-09-26
Replace interpolated strings

hastinbe:
String interpolation added in 049abd91460e7921da588f44a1a247764675746c requires C# 6.0 or Mono >= 4.0 which is not available in Debian, Gentoo, openSUSE, and Ubuntu stable repositories. Requiring  experimental/unstable packages while breaking backwards compatibility is not a good trade-off.

Signed-off-by: Beau Hastings <beausy@gmail.com>

### DIFF
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -972,8 +972,8 @@ namespace TShockAPI
 						}
 						else if (Itembans.ItemIsBanned(player.TPlayer.inventory[player.TPlayer.selectedItem].name, player))
 						{
-							player.Disable($"holding banned item: {player.TPlayer.inventory[player.TPlayer.selectedItem].name}", flags);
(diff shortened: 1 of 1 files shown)


tylerjwatson:
Mono >= 4.0 is listed as a requirement for tshock. Mono 4.0 is not unstable, or experimental, and using those signed PPAs is required for tshock. 

Mono-3.2 as marked as stable in these repositories actually has stability issues in the simple generational GC. Just because it is marked as stable by a package maintainer doesn't make it so.

Enerdy:
Changes introduced in C# 6.0 require no additional package, as the compiler replaces them at compile time; they're simple shortcuts. This is not an issue to the end user as we provide pre-compiled binaries, unless you intend to build the project yourself.

tylerjwatson:
Will reopen at request of other contributors.

PR #1123, opened 2015-09-27
Add support for composite primary keys

hastinbe:


tylerjwatson:
Hi, 

Thanks for the PR.  Adding constraints to a database with data already in the table can be a flakey operation.  Has this been tested with an already existing tshock database?  Is this going to break tshock for people who already run a server?

hastinbe:
@tylerjwatson will work with new and existing db, and will not break existing tshock servers. What I did here was change the syntax of the query builder so it declares the primary keys as a collection, rather than on a per column basis when generating the CREATE TABLE query. The former will generate invalid SQL when more than one SqlColumn.Primary is true.

However I did fix an issue with SQLite as noted in the last commit.

hastinbe:
Closing this as I will follow up with a more comprehensive PR.

PR #1126, opened 2015-10-09
Fix #1082

### LINKED ISSUE #1082: Gender potion does not save gender after leaving/restart.
As stated in the title upon using a Gender potion and leaving/rejoining the server you will become a male again (If you started out as a male) and likewise when you start as a female you'll end up like that.

hastinbe:
Stores character skin variations in tsCharacter table


WhiteXZ:
Tested and works fine. Good stuff

PR #1154, opened 2016-01-11
Add a PlayerPermission hook

Enerdy:
Adds a PlayerPermission event to `TShockAPI.Hooks.PlayerHooks` that can be hooked to. This should ease the implementation of alternative permission engines.

This adds a new method to the `TSPlayer` class: `TSPlayer.HasPermission()`. This functions exactly like `Group.HasPermission()` except that first it checks to see if the PlayerPermission event has been handled, and if it has, returns true without checking group permissions. If it hasn't been handled, it will first check if the player has any temporary groups associated with them, and if it has, return the value of said group's `HasPermission()` method; otherwise, it will return the value of the original `Group.HasPermission()`.

With that being said, the only place where `Group.HasPermission()` is still used as a stand-alone is on the `ban` command for offline users (offline users don't have `TSPlayer` objects) and REST routes (only the user's name and usergroupname are used).

In addition, this commit adds extra documentation to places and fixes indentation in other places.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -175,7 +175,7 @@ namespace TShockAPI
 				return true;
 			foreach (var Permission in Permissions)
 			{
-				if (ply.Group.HasPermission(Permission))
+				if (ply.HasPermission(Permission))
 					return true;
 			}
 			return false;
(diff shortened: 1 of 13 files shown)


Strinkor:
:+1:

Ijwu:
lgtm

nicatronTg:
I'll say that @WhiteXZ should merge this if he's satisfied with 81fff9b.

PR #1166, opened 2016-02-28
Concurrent Dictionary Access Fix

CoderCow:
A server has been suffering from several performance issues and I tracked the problem down to threads getting stuck in 

> System.Collections.Generic.Dictionary`2[[...]].FindEntry()

After I synchronized the two dictionaries of TSPlayer with this code, the issues were resolved.

There's also [this arcticle](https://blogs.msdn.microsoft.com/tess/2009/12/21/high-cpu-in-net-app-using-a-static-generic-dictionary/) which explains this problem of unsynchronized dictionaries in a bit more detail.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2137,8 +2137,9 @@ namespace TShockAPI
 				{
 					args.Player.TilePlaceThreshold++;
 					var coords = new Vector2(tileX, tileY);
-					if (!args.Player.TilesCreated.ContainsKey(coords))
-						args.Player.TilesCreated.Add(coords, Main.tile[tileX, tileY]);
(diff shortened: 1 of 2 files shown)


tylerjwatson:
Hi @CoderCow 

Good find, and welcome back to the community. :) The patch looks good, so +1 from me.  

Have you had this running for a while?  It doesn't seem like a deadlock is likely, but you do use more than one mutex variable here.

CoderCow:
Heya, thanks.

Yes, this fix has been live on the related server since more than 12 days now.

WhiteXZ:
If it runs fine then +1 from me.
Welcome back

tylerjwatson:
Merge'd.

nicatronTg:
Thanks @CoderCow :)

PR #1187, opened 2016-05-25
Updated project to use NuGet where possible

Celant:


Enerdy:
Waiting on travis, but should be good.

PS: We're ignoring the two-man rule for these Pull Requests speed up the release of TShock for Terraria 1.3.1.

Enerdy:
No longer an issue with references, we'll fix the remaining.

PR #1193, opened 2016-05-29
Add an option to enable Announcement Boxes, now disabled by default

Enerdy:
This was requested by a few users. If it doesn't make it in, users can always make a plugin that does the same, or even add a limit of messages per second to those. I thought it'd be something we should include considering it's as easy as toggling a variable in `Main.cs`.


WhiteXZ:
Sounds like a plugin feature, not a TShock feature

Strinkor:
:+1: to @Enerdy for a Announcements Boxes messages limit.

Enerdy:
Sounds like a plugin every server would need to prevent trolls IMO.
Another possible solution would be to add a permission, though it might be a bit harder to implement.

nicatronTg:
@Enerdy did this ever become a plugin?

Enerdy:
Nope.

nicatronTg:
@enerdy oh :(

nicatronTg:
I've reopened this PR. It's a two line change, it never became a plugin, etc. If we really want to not include it feel free to deny this PR and we can delete the branch.

PR #1216, opened 2016-06-15
Properly follow tileData for region checks. Fix #1159

### LINKED ISSUE #1159: [Regions] Statues & Mannequins calculated wrong
TShock Bamboo Build: #585
Issue: When attempting to place Mannequins/Statues on a area that is not protected, but are not atleast 1 blocks above the protected area will fail.

RED AREA: Protected Region
GREEN AREA: Unprotected Region Z'd over Protected Region
YELLOW AREA: Placing statue or Mannequin on this platform shows error message of protected region, but allows you to place it anyway. SIDE NOTE: Mannequin becomes indestructible after placement, and must remove the block standing on it to remove it!
BLUE AREA: This height is safe for placing statues and Mannequins without error message.

FINAL NOTE: Attempting to place either of these directly on the asphalt platform which is protected, results in automatic destruction of the placed statue or mannequin.

Picture:
image

cracker64:
Pretty simple, the manual checks are just completely wrong and missing many many types.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2231,12 +2231,8 @@ namespace TShockAPI
 			if (tileData == null)
 				return true;
 
-			if (tileData.Width == 3 && tileData.Height == 2) { x -= 1; y -= 1; }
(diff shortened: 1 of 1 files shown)


WhiteXZ:
LGTM. Gotta ~~love~~ hate legacy code

PR #1219, opened 2016-06-19
Make GeoIP Great Again!

Enerdy:
GeoIP.dat is also loaded from root instead of `/tshock` now since that allows us to more easily bring it along the release zip without having to create a `/tshock` folder.


Enerdy:
Context: `EnableGeoIP` has been sitting on the config file for some time now while providing no functionality at all. At some point in time, we stopped shipping `GeoIP.dat` with tshock releases, which is necessary to run the GeoIP service.

That doesn't necessarily mean we need to provide it ourselves, granted we provide documentation in how to grab the legacy GeoIP database yourself, but seeing as it was designed as a config file "opt-in", using it should be as easy as enabling said option.

Patrikkk:
:+1: :D  (Why don't you :lgtm: your self Enerdy o.o )

Issue #1220, opened 2016-06-20
Separate permissions for /give and /item.


tlworks:
Players will be able to give someone else items through the /give command while the /item command only gives the player the desired item, due to this I think it would be better to add a separate permission for /item and /give to prevent players on non-SSC servers to spam someone else's inventory full with tons of items and prevent group abuse on SSC servers by using /give to give players items which they are not allowed to do so.

mistzzt:
#1221

mistzzt:
#1222

PR #1221, opened 2016-06-20
Separate item give perm from item spawn

Patrikkk:
As requested in #1220 .
Feel free to deny or accept. Just thought I would make a PR about it.
Seems like a reasonable feature request tho.


Simon311:
I was pretty sure this was already implemented, but somehow it isn't

mistzzt:
You should modify the default permission for admin group. E.g. Give admins the permission to give item.

Issue #1198, opened 2016-05-31
Gem locks not protected in Regions.


tlworks:
Basically there's nothing to add here, Gem Locks are not protected in Regions.

Using latest TShock build 4.3.15.0

Enerdy:
Those are used for games, right? Why would you want a flag (gem) that cannot be taken?

This is more like a feature request than a bug, considering Gem Locks aren't actual tiles, and region protection is only meant to protect tiles from being altered (and to a further extent, signs and chests, if the option is enabled in the config). Could take a config option, perhaps?

ProfessorXZ:
This issue was fixed and merged.

PR #1223, opened 2016-06-21
Add protection for gem locks in region

mistzzt:
Gem locks will be protected in region.
You can also turn off this function in **config.json**.
For #1198


mistzzt:
It seems like that the Travis build failed due to this reason:
**The command "sudo apt-get update -qq" failed and exited with 100 during .**

Not my fault?

Patrikkk:
Yep, it's all good :) :+1:

WhiteXZ:
Re-ran the test. Works now.

PR #1224, opened 2016-06-23
Show proper error message at db connect fail

Patrikkk:
If server cannot connect to the database (only experienced it at MySQL) the console will simply close without any error message or log, leaving the user confused.


WhiteXZ:
Will this cause issues for hosts such as Multiplay who use command-line args so as to not have to provide direct input to the console?

Patrikkk:
That is correct. Thanks for pointing that out. :)
I amended the PR.

PR #1228, opened 2016-06-25
Fixing SSC for real

Patrikkk:
This whole things is a mess. (Yes, I know TShock 5.0 soon :tm: )
Why was HasPermission method called before player's group was actually set?
Why was the bypass ssc function tied to a completely irrelevant boolean? 
We might never know :smile: 
Anyway, jokes aside. Tested and works. :)

(Without this fix players will never get their ssc character)


WhiteXZ:
LGTM

Issue #1140, opened 2015-12-03
Request: /setdungeon command to change cultists/old man's spawning area


Kojirremer:
Just what the tin says. I want to be able to move the dungeon map point to somewhere a bit more difficult to reach, since just anybody can go and cause the lunar events nonstop.

I had in mind something like the already-existing /setspawn command, but that deals with the dungeon location instead (just the location point, not a whole freshly-generated dungeon).

TEdit has this feature, but it's a bit hard to use on already-live maps, and bad things happen if you try to use TEdit on InfiniteChests-powered maps.

loohney:
Would love to see this too. 
We had players triggering the lunar event over and over again which is actually pretty annoying.

I've prevented this by selecting the region where the cultist spawns with worldedit and filled the selection with a solid block. He won't spawn anymore, also you can make this block of maybe dirt look naturally and creating a new entrance into the dungeon manually.
It's not a fix but it does the job.

Marcus101RR:
Moving the spawnpoint of dungeon won't do much good, as you would need to build a custom dungeon for cultists to spawn properly. That in fact, means unmodified dungeon walls...

Kojirremer:
And what would be the problem? WorldEdit can set unsafe dungeon walls. I've already tested moving the spawn point with TEdit and it works, but as I said before, it's not easy to do when the world is already live (and "tainted" by InfiniteChests).

Marcus101RR:
My world Regeneration moves dungeon spawn, and loads the dungeon at the new location. Cultists work, so yes, if you move the dungeon you need to move the physical dungeon too.

PR #1234, opened 2016-07-01
Added the SetDungeon command.

webmilio:
Added feature requested in #1140.
The dungeon still has to be physically moved but this can be accomplished using WorldEdit, allowing for live "dungeon moving". You also need to kill the already existant Cultists or reload the map.

capture 2016-07-01 18_29_12
capture 2016-07-01 18_27_40

For some reason, github shows my code as being wrongly indented when its not the case.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -512,7 +512,12 @@ namespace TShockAPI
 				AllowServer = false,
 				HelpText = "Sets the world's spawn point to your location."
 			});
-			add(new Command(Permissions.worldsettle, Settle, "settle")
+            add(new Command(Permissions.worldspawn, SetDungeon, "setdungeon")
+            {
+                AllowServer = false,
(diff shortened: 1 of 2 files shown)


Enerdy:
We've kinda declared feature freeze (no new features added, maintain code up to date with the Terraria version) in preparation for TShock5. That was over a year ago, though.

GitHub is not lying to you, your indentation is, in fact, a bit off. Our entire codebase (barring a few exceptions that were merged without previous examination) uses hard tabs instead of spaces, which is Visual Studio's default setting. This is not usually visible unless you enable the "show spaces" option. To fix this, change your Visual Studio settings at `Tools -> Options -> Text Editor -> C# -> Tabs` from `Insert Spaces` to `Keep Tabs`, then select the code portions you've edited and format them (shortcut: CTRL+E, then F).

Other than that, I approve 👍

PR #1240, opened 2016-07-15
Fix bug: players cannot place sensors

mistzzt:
Players without permission `Permissions.allowclientsideworldedit` can't place sensors.
When one places a sensor the first time, the sensor will disappear, but it will appear as the player try again.


WhiteXZ:
:+1:

PR #1242, opened 2016-07-18
Fixes #1241

### LINKED ISSUE #1241: Fix actuator hacking check
The [Presserator](http://terraria.gamepedia.com/Presserator) is an item that enables placing actuators without having them selected.

ProfessorXZ:
While testing it turned out the "Presserator" doesn't work at all, so I included a fix for this too.
Everything was tested and the Presserator seems to work fine now, and people aren't able to place objects if they have the item in their inventory.


WhiteXZ:
:+1:

PR #1257, opened 2016-07-27
Add toggleparty permission | Fix #1256

### LINKED ISSUE #1256: Add permission for Party Center use
Since it triggers an event, I think it should have it's permission.

Patrikkk:


Enerdy:
:+1:

Issue #1251, opened 2016-07-27
Validation on PlaceTileEntity


MarioE:
Placing tile entities ignores everything.

Issue #1261, opened 2016-07-28
Validation on UpdateItemDrop


MarioE:
Presumably this can be used to crash clients by sending invalid item IDs.

ProfessorXZ:
That is correct, but it is already handled by our GetDataHandler :+1:

MarioE:
That seems to only handle packet 21 and not packet 90. I might be mistaken, though.

ProfessorXZ:
Ah, my mistake. Forgot packet 90 is UpdateItemDrop. Fixing it now...

Issue #1254, opened 2016-07-27
Validation on KillPortal


MarioE:
You are able to kill other people's portals.

PR #1262, opened 2016-07-28
Fixes #1248, #1254 and #1261

### LINKED ISSUE #1248: Validation on CatchNPC
~~Currently it is possible to "catch" any NPC in the world and delete it. Let's fix that.~~

~~E.g., you can "catch" all town NPCs and remove them.~~

Edit: You can't actually catch any NPC, seems like I missed that code. Reverting the catch to the client is still a good idea, though.

ProfessorXZ:
**Changes**

- Players can no longer "Catch" NPCs (tested).

- Players can no longer use KillPortal packets to remove other players' portals (tested).

- Fixes possible client crashes caused by invalid item netIDs.


ProfessorXZ:
**Update:**

- Fixed issue #1251. This only applies to Logic Sensors, other entities are handled as intended.

Ijwu:
lgtm

PR #1264, opened 2016-07-29
Use packet 107 for sending chat messages

ProfessorXZ:


Enerdy:
`Length` is the pixel width a message can have until it splits lines. If set to -1, it should break them based on the client's screen width, which should be the default behaviour. Might wanna use that instead of 800.

Enerdy:
:+1:

Issue #1160, opened 2016-02-07
Item Frame Duplication and Placement of Banned Items


Angelyss:
Heya tShock team! ^^;

I've recently encountered an item duplication bug involving Item Frames. The Item Frame does not have to be in a protected region. To reproduce this exploit, simply take a stackable item with a few items in the stack, then right click the item frame repeatedly quickly.

![2016-02-07_00026](https://cloud.githubusercontent.com/assets/14846414/12873436/94cdf564-cdf7-11e5-983a-738808921e7d.jpg)
![2016-02-07_00027](https://cloud.githubusercontent.com/assets/14846414/12873437/973a5f72-cdf7-11e5-9b9b-a0d36325a2c1.jpg)
![2016-02-07_00028](https://cloud.githubusercontent.com/assets/14846414/12873438/9ae18470-cdf7-11e5-84d7-6df297b57be3.jpg)

Of course, the solution then would be to ban Item Frames. We did that as soon as they were released (same with weapon racks) to prevent problems with people losing items to protected Item Frames. However, it's still possible to place Item Frames and other banned items due to the slight delay between switching to the item and the stone debuff kicking in.

![2016-02-07_00031](https://cloud.githubusercontent.com/assets/14846414/12873490/e556976a-cdf8-11e5-983b-6affeb3af0cf.jpg)
![2016-02-07_00032](https://cloud.githubusercontent.com/assets/14846414/12873491/eac7523e-cdf8-11e5-82da-58b41d56c7e5.jpg)

In that small time frame you will be able to place Item Frames (or other banned items), though they won't immediately appear on-screen sometimes. Simply leaving and rejoining the server will fix that and the item will be properly placed.

![2016-02-07_00033](https://cloud.githubusercontent.com/assets/14846414/12873495/0ae36eb8-cdf9-11e5-92c1-fd9259951fa3.jpg)

Thank you for reviewing this issue!

WhiteXZ:
@Enerdy

tylerjwatson:
@collabs, can someone explain to me what's going on here on a technical level?

WhiteXZ:
Certain tiles have special cases for placement (such as item frames, mannequins, minecart tracks). If we don't handle these cases properly they stop working as intended.

The issue with placing banned items is separate - the time it takes for Terraria to apply the de-buff we tell it to is so long that it allows people to use the banned item before getting de-buffed. Fixing this may be difficult

Gummin:
[...]

PR #1265, opened 2016-07-31
Fixes #1252 and #1160

### LINKED ISSUE #1252: Validation on PlaceItemFrame
Players can use this to modify any item frame or just force items to drop on the server.

ProfessorXZ:
**Changes:**

- The  `messageLength` parameter was removed from TSPlayer.SendMessage because it has no use (see Slack for further details).

- Players can no longer modify protected Item Frames.

- Players can no longer duplicate items using Item Frames.


Enerdy:
Sugoi desu
:+1:

PR #1267, opened 2016-08-01
Fixes #1197

### LINKED ISSUE #1197: Unicorn Mount Dismounting Teleports you to previous location
This is a strange bug, but I think it has to do with TShock and how it handles your movement via Mount and back to walking. If you are on the ground while using the Unicorn Mount, simply go full speed and then Dismount with "R". You will get the Rod of Discord Teleport noise on you and animation.

Perhaps its something to do with a permission or maybe just threshold? Can't seem to find a fix, but Doesn't happen on superadmins account last I checked. Will confirm later.

ProfessorXZ:


Enerdy:
:+1:
Interesting way of fixing it.

PR #1268, opened 2016-08-02
Fixes #1250

### LINKED ISSUE #1250: Fix quick stack ignoring chest protection
We have chest protection in regions, but this is inconsistent and does not apply to quick stack.

ProfessorXZ:


TShockAPI/TShock.cs
```diff
@@ -471,6 +473,45 @@ private void NetHooks_NameCollision(NameCollisionEventArgs args)
...
+			if (tsplr == null)
+			{
+				args.Handled = true;
+				return;
+			}
+
+			if (args.Chest != null)
+			{
+				if (Config.RegionProtectChests && !tsplr.HasPermission(Permissions.editregion) 
+					&& !Regions.CanBuild((int)args.Position.X, (int)args.Position.Y, tsplr) && Regions.InArea((int)args.Position.X, (int)args.Position.Y))
```
Enerdy: Why use `AND` instead of `OR` here? The first check always needs to be true, so `&&` is fine, but I don't think the player needs to be both unable to edit regions and not have the permission to edit the specific tile for us to handle the packet. `Regions.InArea` should also come before any region check, naturally.
The following would be a better fit for this check (with proper syntax, I wrote this in the GitHub editor so it could be incorrect):
```if (Config.RegionProtectChests && Regions.InArea((int)args.Position.X, (int)args.Position.Y) && (!tsplr.HasPermission(Permissions.editregion) || !Regions.CanBuild((int)args.Position.X, (int)args.Position.Y, tsplr)))```
Verbose:
IF `chests are protected inside regions` AND `there is a region in the area` AND EITHER `player doesn't have permission to edit regions` OR `player cannot build in the specific region`, handle the event.

ProfessorXZ: True, this was a derp on my end.

ProfessorXZ:
Fixed :+1: RegionManager.CanBuild was modified to remove unneeded checks.

WhiteXZ:
Is this tested and working?

ProfessorXZ:
It was, yes.

WhiteXZ:
:+1: then :)

Issue #1260, opened 2016-07-28
Validation on LoadNetModule and NetLiquidModule


MarioE:
It appears that liquids can be placed anywhere using this packet and module.

PR #1270, opened 2016-08-10
SyncExtraValue validation. Fixes #1024

### LINKED ISSUE #1024: Expert mode coin dupe
In expert mode; creatures pick up dropped coins and re-drop them on death.
If the coins are half a screen away or so(at 1920x1080) creatures dupe the coins as in the don't actually pick the coins up but they still drop them on death.

Was able to accumulate 300+ platinum coins in about 5 minutes at a mob farm.
Tested  4.3.5 although I have noticed the behavior since 4.3.0 without realizing what I was seeing.
Also tested vanilla and it doesn't happen.

ProfessorXZ:
**Changes:**

- Fixed Expert mode coin duplications.

- Added more comments to TSPlayer.

It is possible that I'll have to add more checks to SyncExtraValue to prevent players from giving out coins to mobs, but I will do that in a later commit as I'm still exploring potential ways of exploiting this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Players can no longer quick stack items into region protected chests (@ProfessorXZ)
 * Rope placement is no longer blocked by range checks (@ProfessorXZ)
 * The Drill Containment Unit breaks blocks properly now (@ProfessorXZ)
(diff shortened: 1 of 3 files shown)


WhiteXZ:
:+1:
Will the other exploit fixes come in this PR, or another PR?

ProfessorXZ:
They will be included in another PR.

ProfessorXZ:
**Update:**

- Fixed #1260

PR #1271, opened 2016-08-14
Fixed item duplication issues

ProfessorXZ:
**Changes:**

- Fixed item duplication caused by range checks and invalid netIDs

- Fixed an issue where clients could send corrupt packets by sending invalid tile IDs in packet 17


WhiteXZ:
:+1: apart from merge conflicts. Will deal with them if someone else +1s

PR #1282, opened 2016-09-10
Fixed IceRod not placing ice blocks

ProfessorXZ:


WhiteXZ:
:+1:

Ijwu:
The two man rule does not include the submitter. Next time let at least two others shoot you a thumbs up before merging. Somebody else may merge it if they see one other has given a thumbs up. This is implied thumbs up.

PR #1283, opened 2016-09-10
Fixed UI issues caused by setting wind speed to high values

ProfessorXZ:


WhiteXZ:
:+1:

Enerdy:
The thing I don't get about this PR is: even if Terraria multiplies speed by 100 to perform calculations, why would you have to do the same if the only purpose of your check is to find out if `speed` is negative? A negative number will still be negative after being multiplied by 100.

WhiteXZ:
The issue is with integer overflows, @Enerdy: excessively high speeds will be multiplied by 100 and overflow, resulting in -Int32.MaxValue speeds

Enerdy:
Interesting. I thought that would throw some kind of exception, but apparently it simply turns into negative.

PR #1284, opened 2016-09-11
Update the version codename

nicatronTg:
This updates the version codename to be something more timely and topical.


nicatronTg:
Can we see some feedback on this @Enerdy @hayley

tlworks:
very-nice-borat-plm

PR #1288, opened 2016-09-25
Fix #1287 | Quickstack to nearby chests issue

### LINKED ISSUE #1287: Quickstack to nearby chests
Hey everyone,

since this new update (and also in the earlier unstable builds of this update) it is not possible to quick stack from my inventory to nearby chests. If i open a chest and then click quick stack everything is finde.
Maybe this is because i'm running it on gentoo linux or does any windows user has the same problem? I tried to register in your forum but the captcha didn't work so i had to post this here.

Patrikkk:
CheckRangePermission accepts coordinate which is divided by 16.
The Position vector which is passed in ForceItemIntoChestEventArgs is
multiplied by 16.
(To be more exact. `new Vector2((float)(Main.chest[i].x * 16 + 16),
(float)(Main.chest[i].y * 16 + 16))`)
The + 16 is there to get the middle of the player so it can make an
accurate check when checking distance between chest and player.
So I'm not really sure if the Position arg should be even passed in the
eventarg.

Anyway, CheckRangePermission always returned true because of the above,
and handled the event, so players could never quickstack into nearby,
unless they had RangeChecks off.
"True if the player should not be able to place the tile." (In our case,
stack items into chest)

Why doesn't CheckRangePermission contain a check on an actual range
permission of the player's group? So groups with the permission could
bypass the check.
Just wondering if there is any legit reason for that.

### DIFF
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -503,7 +503,7 @@ namespace TShockAPI
 					return;
 				}
 
-				if (CheckRangePermission(tsplr, (int)args.Position.X, (int)args.Position.Y))
+				if (CheckRangePermission(tsplr, args.Chest.x, args.Chest.y))
 				{
 					args.Handled = true;
 					return;


ProfessorXZ:
I think it would be better to fix the actual hook in TSAPI though, as people who might use it will probably be confused if something doesn't work. This is just my opinion though (unless changing the code in TSAPI would break something).

Patrikkk:
I disagree. 
The `Position` variable which is passed in `ForceItemIntoChestEventArgs` is a world coordinate (tile).
The hook is doesn't really need "fixing". Tho description could be changed. And not sure if there is a point to send the middle of the chest ( + 16) instead of just sending the normal tile coord of it as we usually do. But yeah, that's up to White if he want's to change it.

So, the fix is this. To use Chest.x and y on `CheckRangePermission` method.

PR #1294, opened 2016-10-11
Remove Firework command (fixes #1083)

### LINKED ISSUE #1083: /firework often does not seem to work.
Below is a print screen which shows that the command /firework has failed. Upon leveling up people receive a firework above their head but most of the time it fails to do so.

https://gyazo.com/7ba9ddc3714f29b1ef390b64c386b5e0

nicatronTg:
Seems like the firework command is broken and in #1083 seems like the solution was to remove it. Thoughts?


nicatronTg:
'Fix' no longer needed, command works as intended.

PR #1302, opened 2016-10-14
Auth rewrite. Closes #1064.

### LINKED ISSUE #1064: Disable the auth system if a superadmin exists
Currently, the auth system's functionality is determined by the presence of auth.lck inside the TShock config directory. This has both benefits and drawbacks. If you lose your superadmin password, you can delete auth.lck and make a new account very easily, without needing to use SQL. However, if you never use the auth system and run auth-verify, your server is open to attack via bruteforce.

I think that we should disable the auth system if a superadmin exists in the database. If a superadmin exists, just disable the entire system.

An alternative solution is to disable the auth system automatically after 10 minutes of non-use, and to send a very big scary warning to the console if this happens (warning, your server is open to attack, etc).

Enerdy:
A possible solution to the problem. Here are the changes included:
* The auth system will now be disabled on start-up if a superadmin user exists. An `auth.lck` file is still created to avoid having to fetch all users in the future, which in big, old servers could potentially slow down the start-up by a lot.
* The `auth-verify` command has been removed. `auth` now checks if the player is a **permanent** superadmin on use and disables the system if so.
* The `auth` command is now more informative in its error messages, and should not fail due to thrown exceptions anymore.
This closes #1064.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -208,14 +208,10 @@ namespace TShockAPI
 				ChatCommands.Add(cmd);
 			};
 
-			add(new Command(AuthToken, "auth")
+			add(new Command(AuthToken, "auth", "auth-verify")
 			{
 				AllowServer = false,
 				HelpText = "Used to authenticate as superadmin when first setting up TShock."
-			});
-			add(new Command(Permissions.authverify, AuthVerify, "auth-verify")
-			{
-				HelpText = "Used to verify that you have correctly set up TShock."
 			});
 			add(new Command(Permissions.user, ManageUsers, "user")
 			{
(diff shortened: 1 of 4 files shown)


TShockAPI/TShock.cs
```diff
@@ -789,36 +789,43 @@ public static void HandleCommandLinePostConfigLoad(string[] parms)
...
+			else if (!File.Exists(Path.Combine(SavePath, "authcode.txt")))
+			{
+				var r = new Random((int)DateTime.Now.ToBinary());
+				AuthToken = r.Next(100000, 10000000);
 				Console.ForegroundColor = ConsoleColor.Yellow;
-				Console.WriteLine(
-					"TShock Notice: authcode.txt is still present, and the AuthToken located in that file will be used.");
-				Console.WriteLine("To become superadmin, join the game and type {0}auth {1}", Commands.Specifier, AuthToken);
+				Console.WriteLine("TShock Notice: To become SuperAdmin, join the game and type {0}auth {1}", Commands.Specifier, AuthToken);
 				Console.WriteLine("This token will display until disabled by verification. ({0}auth-verify)", Commands.Specifier);
```
nicatronTg: "This token will display until disabled by verification. (/auth-verify)" shouldn't be here.

nicatronTg: screen shot 2016-10-14 at 5 05 53 pm

TShockAPI/Commands.cs
```diff
@@ -208,15 +208,11 @@ public static void InitCommands()
 				ChatCommands.Add(cmd);
 			};
 
-			add(new Command(AuthToken, "auth")
+			add(new Command(AuthToken, "auth", "auth-verify")
```
nicatronTg: I would just remove `/auth-verify` if `/auth` serves the same purpose. It's simpler and more convenient. I thought you did, actually, based on your description.

screen shot 2016-10-14 at 5 12 24 pm

nicatronTg:
Also _the changelog wasn't updated_.

nicatronTg:
To clarify I interpreted your original description as a full removal of the `/auth-verify` command, so I reviewed thinking it should be gone. I realize in hindsight that you aliased it, which is fine, but it's still more elegant to just have one command instead. That's my 2c, but other than kicking valid logged in super admins, this all seems legit to me.

TShockAPI/Commands.cs
Enerdy: I did remove the original `auth-verify` command, I added its command alias to `auth` for "backward compatibility". Will help old users that may be accustomed with running `auth-verify` first thing when running a clean install.

Actually, `auth` has `AllowServer=false` so the above doesn't even work from the console right now (though it could simply be changed), but you shouldn't have to verify if you're using the console as you can just add a superadmin user.

nicatronTg: @Enerdy I wasn't attempting to verify via console, but I was kicked and thought the inclusion of the command was erroneous in and of itself. (See latest comment).

Issue #1206, opened 2016-06-03
Is this supposed to happen upon Restart?


tlworks:
Assuming the cause is that the user shown in the image 'PGC' leaves after the server is already shut down?

https://i.gyazo.com/358e495037555904ee9fe46d8cd272a7.png

Don't really know what the cause is of this.

Enerdy:
That is, in fact, the probable cause of that exception. Still, there should be a way of avoiding this.

tlworks:
It's simple, disallow people/players to join after the command /restart (or well, /exit) is executed as back in 1.2.4.1 the servers saved the world rather quick where people didn't even had the time to join back before it actually restarted. Ever since 1.3 the world saving takes quite a while so people can simply join back in while the restart (shut down) is executed.

Patrikkk:
Couldn't reproduce :/
I started up server (0 players)
I execute restart and quickly join the server. It joins server and then I get kicked once world finished saving.

tlworks:
Would still suggest to prevent joining after the command /exit is issued, as joining after using /exit would be irrelevant to begin with or KICK the player with the message that the server has been shut down or is shutting down, it also seems that is has to be 'perfectly timed' in order to reproduce properly.

Depending on the amount of players online, CPU usage & the world size, the saving after /exit could take longer, making it able for people to join back on.

Issues mostly happens when the user surpasses the part where it should KICK the user upon finished world saving.

tlworks:
Also error log:

[07/05/16 11:09:48] [TShock] Error: 2016-07-05 11:09:48 - Utils: INFO: Broadcast: Gbvkhfkauj has left.
[07/05/16 11:09:48] [TShock] Error: 2016-07-05 11:09:48 - TShock: INFO: Gbvkhfkauj disconnected.

nicatronTg:
If the server is going down, I think this is a non-issue/"exception normal." It's not something I expect _anyone to pick up_ but I'm going to flag it as PR wanted, because why not. Don't expect this to be fixed any time soon.

PR #1303, opened 2016-10-16
Fix /"

### LINKED ISSUE #1301: Typing /" bypasses the chat handler
This was mentioned by OFF/tlworks before but was never reported here. Players are able to bypass mutes etc by typing /" in chat. This breaks the whole handler and chat is no longer handled by our hook. If I remember correctly the issue lies in the `ParseParameters(string)` method. I believe the nature of that code is what causes this (keep in mind that I am not 100% sure about this).

Simon311:
Fixes #1301.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -628,12 +628,29 @@ namespace TShockAPI
 			if (cmdPrefix == SilentSpecifier)
 				silent = true;
 
-			var args = ParseParameters(cmdText);
-			if (args.Count < 1)
-				return false;
+			int index = -1;
+			for (int i = 0; i < cmdText.Length; i++)
+				if (IsWhiteSpace(cmdText[i]))
+				{
+					index = i;
+					break;
+				}
+			string cmdName;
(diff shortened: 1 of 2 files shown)


TShockAPI/Commands.cs
```diff
@@ -628,12 +628,29 @@ public static bool HandleCommand(TSPlayer player, string text)
 			if (cmdPrefix == SilentSpecifier)
 				silent = true;
 
-			var args = ParseParameters(cmdText);
-			if (args.Count < 1)
-				return false;
+			int index = -1;
+			for (int i = 0; i < cmdText.Length; i++)
+				if (IsWhiteSpace(cmdText[i]))
```
nicatronTg: Whether or not this is syntactically correct or not, stylistically there should be some brackets around this for loop.

nicatronTg:
Let me clarify that last comment: I tested and made sure that command parsing wasn't changed too.

Simon311:
Let me fix something quickly

Simon311:
Once this is accepted, I will push commits and make another PR fixing #1206.

PR #1306, opened 2016-10-16
Update all the GPL headers #1305

### LINKED ISSUE #1305: Update all the GPL headers
All of the GPL headers need to say the current year (2016) and files that don't have GPL headers need to have GPL headers.

adilsoncarvalho:
Update all years on the GPL headers to 2016 and add those headers when missing.

Resolves #1305

### DIFF
--- a/TShockAPI/BackupManager.cs
+++ b/TShockAPI/BackupManager.cs
@@ -1,6 +1,6 @@
 ﻿/*
 TShock, a server mod for Terraria
-Copyright (C) 2011-2015 Nyx Studios (fka. The TShock Team)
+Copyright (C) 2011-2016 Nyx Studios (fka. The TShock Team)
 
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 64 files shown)


nicatronTg:
@adilsoncarvalho just FYI you don't need to worry about adding the GPL header to anything other than code files c:

nicatronTg:
Also just drop a note when you're done (you've labeled this as WIP so I assume it is) c:

adilsoncarvalho:
I'm adding it only to all `.cs` files @nicatronTg

adilsoncarvalho:
@nicatronTg there are lots of files without a header on a submodule. Should I add to them and commit? I saw no similar issue on its repository.

If not, the work here is done. I skipped also the generated files. No point in adding a header on them.

nicatronTg:
@adilsoncarvalho no, but thank you for the offer! The submodule is proprietary code generated by a decompiler and monkey patched to work correctly. The license on code in that repository starts with "it's complicated," as we aren't really cross licensing it or anything.

Thanks!

adilsoncarvalho:
LOL. This sould be a license! "It's complicated" 😛

nicatronTg:
Definitely should :p

PR #1310, opened 2016-10-17
Fix for HealOtherPlayer exploit, also fix #1309

### LINKED ISSUE #1309: "Invalid command entered" when no command entered
Maybe this was caused by fixing security issues(?), error message is also displayed when I just pressed 'Enter' key (or whitespace only). It's a bit annoying.

Is there any way to avoid this, something like

if (cmdText.Trim() == "")
{
    return true;
}

at HandleCommand() in TShockAPI/Commands.cs?

Simon311:
Fix for HealOtherPlayer exploit, also fix #1309.
The code should self-explanatory.

### DIFF
--- a/TShockAPI/ConfigFile.cs
+++ b/TShockAPI/ConfigFile.cs
@@ -295,6 +295,10 @@ namespace TShockAPI
 		[Description("Disable a player if this number of projectiles is created within 1 second.")]
 		public int ProjectileThreshold = 50;
 
+		/// <summary>ProjectileThreshold - Disables a player if this number of projectiles is created within 1 second.</summary>
+		[Description("Disable a player if this number of projectiles is created within 1 second.")]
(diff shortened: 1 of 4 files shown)


Simon311:
Needs some fixing, one more commit incoming

Simon311:
This is tested.

TShockAPI/ConfigFile.cs
```diff
@@ -295,6 +295,10 @@ public class ConfigFile
 		[Description("Disable a player if this number of projectiles is created within 1 second.")]
 		public int ProjectileThreshold = 50;
 
+		/// <summary>ProjectileThreshold - Disables a player if this number of projectiles is created within 1 second.</summary>
+		[Description("Disable a player if this number of projectiles is created within 1 second.")]
```
WhiteXZ: Description is incorrect

Issue #901, opened 2015-04-15
Rename /userinfo, add an actual /userinfo command


Enerdy:
Currently, the `userinfo` command displays an online player's IP Address, the user they're logged in as, and that user's group. The command alias can be quite missleading akin to the command's current functionality IMO.

What I suggest is that we rename/move the current `userinfo` functionality somewhere else (perhaps name it `playerinfo` instead?), and since we're at it, include a new `userinfo` command that actually retrieves information regarding an user, and therefore requires the full username to be given and doesn't care if the user is online or not. This should probably either receive the userinfo alias, or we could look into adding it to the `user` command.

Right now, the only way to check an user's group is in the database (`user group <username>` does not retrieve the current group...), and this new command could include additional information such as registration date and the WIP user-based permissions. Basically, we'll be killing two birds with one stone.

nicatronTg:
To be honest, the information should just be added to userinfo rather than making a new command or anything else. Again, however, it really makes you wonder how many people use userinfo at all.

For the sake of complexity, the whole idea of a "user" versus a "player" is annoying. I think the verbiage should change somewhere, but I'm not sure a better term to use.

Enerdy:
The main concern here is that `userinfo` will only search for online players, so there is currently no way of quick checking an user's group outside of a database query or REST.

nicatronTg:
Hmm, good point. I agree with the rename, then, and adding a new command.

nicatronTg:
From slack: "A good compromise would be to get the player info too if they’re logged in, or else just find the offline user’s info and only return that."

WhiteXZ:
PlayerInfo is misleading as the command is always working with the user object. Functionality for off line users while keeping the current name would be best imo.

Ijwu:
:+1: to the compromise.

bippity:
:+1:

Simon311:
Currently, this functionality is available in my UserExists plugin, as well as the functionality to view info regarding bans.
Feel free to put that in TShock.

nicatronTg:
@Simon311 feel free to send pr

nicatronTg:
(Was closed by #1312). Thanks for pointing this out @doggofalchion.

PR #1312, opened 2016-10-18
Add /accountinfo, fix offline bans exploit, fix #1309 more

### LINKED ISSUE #1309: "Invalid command entered" when no command entered

Simon311:
I've implemented the /accountinfo command, which allows viewing _any_ user's information, as opposed to /userinfo. However, /userinfo remained untouched, for convenience of server admins.

/accountinfo has two permissions: tshock.accountinfo.check and tshock.accountinfo.details.
Having only the first permission will allow you to execute the command, and discover whetever a certain username is registered and if so, when its last login took place.
The second permission will allow you to see very detailed information about a user, including its last login time, group, last known IP and exact registration time.
The need for two permissions is dictated by the two cases this command can be used: for everyday use, and for admin purposes.
Everyday usage can be described as assisting new players in selecting a username for themselves. At the same time last login time may be used by regular players like a "last seen" command, allowing to see when their friends have last logged into the server.
Admin usage includes any other uses, such as IP-banning a hacker, finding out a certain user's group, finding out when a certain user account was registered. Exposing this information to regular users might be considered a security risk.

I've also discovered and fixed some counter-intuitive behavior in KnownIps object handling, which allowed a hacker to successfully evade offline bans on certain setups. This behavior may be considered an exploit.

More fixing has been applied for #1309, which became an issue after one of my previous pull requests. This should require no subsequent fixing.

Overall issues addressed by this: #901, #1309.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -244,6 +244,10 @@ namespace TShockAPI
 				DoLog = false,
 				HelpText = "Registers you an account."
 			});
+			add(new Command(Permissions.checkaccountinfo, ViewAccountInfo, "accountinfo", "ai")
+			{
+				HelpText = "Shows information about a user."
+			});
 			#endregion
 			#region Admin Commands
 			add(new Command(Permissions.ban, Ban, "ban")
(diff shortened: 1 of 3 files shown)


nicatronTg:
@Simon311 you made mention (in Slack, maybe?) that this issue may not have entirely fixed offline bans. Is this still a thing that's valid?

PR #1313, opened 2016-10-20
Add Hardened Stone to the list of tiles that can be altered by players.

### LINKED ISSUE #1138: Hardened Sand (Possible permission issue)
Corruption / Crimson / Hallowed counterpart of Hardened Sand isn't changing back to normal when contaminator / purification powder is used upon it. ( It will change back during world reset ) [ MIGHT only occur only in our server ].

However, when a "superadmin" used it it changed back, when a normal user used it, it would not change. As if the Hardened Sand block requires special permission. Be advised, other blocks around do change biomes by the users. Leading to believe it may be a permission issue, or another issue in tshock definition regarding a user changing block types.

image

DogooFalchion:
I added the hardened stone set of tiles to the list of exceptions in GetDataHandlers as well as refactored the existing ancient implementation into one that takes advantage of the TileID.Sets.Conversion sets for tile conversions.

This should fix #1138.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -244,6 +244,10 @@ namespace TShockAPI
 				DoLog = false,
 				HelpText = "Registers you an account."
 			});
+			add(new Command(Permissions.checkaccountinfo, ViewAccountInfo, "accountinfo", "ai")
+			{
+				HelpText = "Shows information about a user."
+			});
 			#endregion
 			#region Admin Commands
 			add(new Command(Permissions.ban, Ban, "ban")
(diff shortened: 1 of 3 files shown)


DogooFalchion:
I accidentally forgot how this forking thing worked and made changes to general-devel.  I will reopen this PR from an feature-branch.

nicatronTg:
@DogooFalchion this would've worked either way just because the destination is all that really matters, and you forked from gen-dev.

DogooFalchion:
But I wouldn't have been able to make a second PR from Gen-dev on my repo given its "dirty" nature until this is approved.  Unless github is a lot smarter than I give it credit :)

nicatronTg:
Yep, no, you're right if that was your use case c:

PR #1316, opened 2016-10-22
Add Hardened Stone to the list of tiles that can be altered by players.

### LINKED ISSUE #1138: Hardened Sand (Possible permission issue)

DogooFalchion:
I added the hardened stone set of tiles to the list of exceptions in GetDataHandlers as well as refactored the existing ancient implementation into one that takes advantage of the TileID.Sets.Conversion sets for tile conversions.

This should fix #1138.

https://github.com/NyxStudios/TShock/pull/1313

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -1823,7 +1823,7 @@ namespace TShockAPI
 						{
 							// Grass <-> Grass
 							if ((TileID.Sets.Conversion.Grass[tile.type] && TileID.Sets.Conversion.Grass[newtile.Type]) ||
-                                // Dirt <-> Dirt
+								// Dirt <-> Dirt
 								((tile.type == 0 || tile.type == 59) &&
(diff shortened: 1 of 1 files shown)


nicatronTg:
cc @NyxStudios/tshock

WhiteXZ:
@DogooFalchion did you check that the sets used contain all the relevant IDs?

DogooFalchion:
I did, thats why one of this looks murdered.

Ijwu:
LGTM. Merging.

Issue #1314, opened 2016-10-20
ShowFiletoUser MOTD color parsing failure


DogooFalchion:
While the spec for Terraria is one color per line, the design of the MOTD color parsing fails if multiple colors are present in one line.  The reason is because as it removes the first color string %###,###,###%, it does not keep track of that.  The string is now shorter than it was and the indices in the next match no longer correspond with the correct characters in the string.  

A solution could be if multiple matches are made, the MOTD should show some intelligible error to the player and not show the MOTD as well as a console message for the server owner explaining the one color per line terraria limit.

Another solution is to properly handle multiple colors by only using the first or last color and to not using String.Remove().

See https://tshock.co/xf/index.php?threads/something-went-wrong.4851/ for an example.

Simon311:
Doesn't the new chat packet support multiple colors and item display in the string?
Should probably just remove the colors altogether and tell people to format it the Terraria-way

ProfessorXZ:
TShock parses files on it's own. It will iterate over the file and look for the rgb format, thus it's not the packet's fault really.

Simon311:
@ProfessorXZ Did you even read what I wrote?

ProfessorXZ:
Yes I did, and I told you it is not the packet's fault. 
>Doesn't the new chat packet support multiple colors and item display in the string?

Simon311:
@ProfessorXZ Well then, let me reiterate.
I am fully aware of how TShock handles motd and I did not blame the packet at all.
What I said is, that there is now a packet and a whole system of tags to support colors natively in Terraria, and therefore we should not apply any new additions to our system. In fact, we should probably remove our own system for parsing colors and tell people to rely on Terraria's chat tags.
Next time, please re-read my full message for as long as it takes to understand.

Simon311:
The current system also does not require any additions on our end. If we want to keep the backwards-compatibility - then no changes are required at all, aside from changes to our guidelines on formatting the MOTD file.
![image](https://cloud.githubusercontent.com/assets/1058851/19550575/c1da8bec-96b2-11e6-9457-f827d23b5dc6.png)

Simon311:
The issue though is mostly about our method not correctly handling an error, this needs some work indeed.

WhiteXZ:
For the record, the alternative to simply removing our parsing was to translate as such:
`%255,255,255%This is some text! %255,0,0%This text changes colours on the same line!`
->
[...]

PR #1317, opened 2016-10-22
Update MotD file parsing to convert color formats into smart text.

DogooFalchion:
I have updated the Regex and logic to find and replace all text on a line with the smart text chat format.  This will replace all text that follows one of our old color formats until the next color format with smart text blocks.  We are still parsing a single line at a time, so the color does not carry over to the next line.  If we want to include that as a feature we certainly could make an improvement to keep track of the last color code used and apply that to all text until the next color code.

%255,000,000% This is text. -> [c/ff0000:This is text.]

This should address #1314.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -4806,7 +4806,7 @@ namespace TShockAPI
 
 		private static void Motd(CommandArgs args)
 		{
-			TShock.Utils.ShowFileToUser(args.Player, "motd.txt");
+			TShock.Utils.ShowFileToUser(args.Player, FileTools.MotdPath);
 		}
 
 		private static void Rules(CommandArgs args)
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -4811,7 +4811,7 @@ namespace TShockAPI
 
 		private static void Rules(CommandArgs args)
 		{
-			TShock.Utils.ShowFileToUser(args.Player, "rules.txt");
+			TShock.Utils.ShowFileToUser(args.Player, FileTools.RulesPath);
 		}
 
 		private static void Whisper(CommandArgs args)
--- a/TShockAPI/FileTools.cs
+++ b/TShockAPI/FileTools.cs
@@ -40,7 +40,7 @@ namespace TShockAPI
 		/// </summary>
 		internal static string MotdPath
 		{
-			get { return Path.Combine(TShock.SavePath, "motd.txt"); }
+			get { return Path.Combine(TShock.SavePath, FileTools.MotdPath); }
 		}
 
 		/// <summary>
--- a/TShockAPI/FileTools.cs
+++ b/TShockAPI/FileTools.cs
@@ -40,7 +40,7 @@ namespace TShockAPI
 		/// </summary>
 		internal static string MotdPath
 		{
-			get { return Path.Combine(TShock.SavePath, FileTools.MotdPath); }
+			get { return Path.Combine(TShock.SavePath, "motd.txt"); }
 		}
 
 		/// <summary>
--- a/TShockAPI/Rest/RestManager.cs
+++ b/TShockAPI/Rest/RestManager.cs
@@ -280,7 +280,7 @@ namespace TShockAPI
 		[Token]
 		private object ServerMotd(RestRequestArgs args)
 		{
-			string motdFilePath = Path.Combine(TShock.SavePath, FileTools.MotdPath);
+			string motdFilePath = FileTools.MotdPath;
 			if (!File.Exists(motdFilePath))
 				return this.RestError("The motd.txt was not found.", "500");
 
--- a/TShockAPI/Rest/RestManager.cs
+++ b/TShockAPI/Rest/RestManager.cs
@@ -280,7 +280,7 @@ namespace TShockAPI
 		[Token]
 		private object ServerMotd(RestRequestArgs args)
 		{
-			string motdFilePath = Path.Combine(TShock.SavePath, "motd.txt");
+			string motdFilePath = Path.Combine(TShock.SavePath, FileTools.MotdPath);
 			if (!File.Exists(motdFilePath))
 				return this.RestError("The motd.txt was not found.", "500");
 
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -1173,7 +1173,7 @@ namespace TShockAPI
 		{
 			if (args.Handled)
 				return;
-			
+
 			if (!Config.AllowCrimsonCreep && (args.Type == TileID.Dirt || args.Type == TileID.FleshWeeds
 				|| TileID.Sets.Crimson[args.Type]))
 			{
@@ -1577,7 +1577,7 @@ namespace TShockAPI
 			if (Config.DisplayIPToAdmins)
 				Utils.SendLogs(string.Format("{0} has joined. IP: {1}", player.Name, player.IP), Color.Blue);
 
-			Utils.ShowFileToUser(player, "motd.txt");
+			Utils.ShowFileToUser(player, FileTools.MotdPath);
 
 			string pvpMode = Config.PvPMode.ToLowerInvariant();
 			if (pvpMode == "always")
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -837,6 +837,8 @@ namespace TShockAPI
 			ComputeMaxStyles();
 			FixChestStacks();
 
+			Utils.UpgradeMotD();
+
 			if (Config.UseServerName)
 			{
 				Main.worldName = Config.ServerName;
--- a/TShockAPI/Utils.cs
+++ b/TShockAPI/Utils.cs
@@ -50,7 +50,8 @@ namespace TShockAPI
 		/// <summary>instance - an instance of the utils class</summary>
 		private static readonly Utils instance = new Utils();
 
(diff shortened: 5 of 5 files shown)


nicatronTg:
Copy pasting my thoughts from Slack:

> as a former server owner, I can say that the _most frustrating thing_ is when I drop in new executables because I’m updating over the weekend
> and then magically my setup breaks because someone asshole didn’t add a converter for a config file or upgrade their database format for me
> The reality is that most server owners won’t read the changelog — they’ll just grab the latest version of the software and install it
> and when something changes or _breaks their setup_ they get to complain and wonder why something isn’t working
> you don’t update windows and find out that all your preferences got deleted, in all but the most stupid cases, they migrate every preference to an equivalent in the new system or provide an explanation why it’s no longer needed

I kinda want to hear from @Simon311 on this. @Simon311 what, exactly, is covered by the TShock color system that isn't by this system? Setting up a base color and mixing colors? I'd love some input as to what that is and what's going on.

Simon311:
@nicatronTg Examine this sample:

> This server is running [i/s1:29]TShock for Terraria[i/s1:29].
> %000,255,000%Type [c/00FF00:/help] for a list of [c/FFFF00:commands].
> %255,000,000%Current map: [c/00FF00:%map%]
> Current players: %players%

This worked perfectly. If we merge this PR, this will quite likely break. If it does - I will be forced to change this motd file either to:

> This server is running [i/s1:29]TShock for Terraria[i/s1:29].
> %000,255,000%Type %255,000,000%/help%000,255,000% for a list of %255,255,000%:commands %000,255,000%.
> %255,000,000%Current map: %000,255,000%%map%
> Current players: %players%

or

> This server is running [i/s1:29]TShock for Terraria[i/s1:29].
> [c/00FF00:Type ][c/FF0000:/help][c/00FF00: for a list of ][c/FFFF00:commands].
> [c/FF0000:Current map: ][c/00FF00:%map%]
> Current players: %players%

both of which are **uglier and close to unreadable**. Most importantly, both of these options introduce **junk** into actual text.
Not to mention this PR **does break existing behaviour** (although I doubt this is used by a lot of people).

nicatronTg:
> Not to mention this PR does break existing behaviour (although I doubt this is used by a lot of people).

Not sure how this breaks existing behavior. It provides a conversion path up to the new format. I guess it breaks default base coloring but you can achieve the same thing with your second example. Arguably you're right in that it does cause the MOTD to kinda look bad, but at the same time, it does migrate existing MOTDs to a new format.

As it stands @WhiteXZ's commit currently breaks everything out of the box and this simply provides a way to up convert to that.

_An alternative suggestion/solution would be to have someone write their own MOTD plugin that supports the old format or provides a fallback._

_An alternative PR could introduce the old format_ but there is some credence to the fact that the old format was designed in a remarkably hacky way, so I wouldn't exactly call it a robust or resilient format.

Simon311:
> Not sure how this breaks existing behavior.

If I use my first example, the real Terraria tags will NOT work because they will be nested inside TShock-generated Terraria tags, which is not supported by Terraria.

nicatronTg:
> If I use my first example, the real Terraria tags will NOT work because they will be nested inside TShock-generated Terraria tags, which is not supported by Terraria.

I'm super tired, and I'm not sure I'm reading this right. What's a "TShock-generated Terraria" tag and what's a real Terraria tag? This PR looks like it would convert your TShock tags to Terraria tags and work as intended, no?

nicatronTg:
(That is, aside from having color underlays where the old color would resume post-Terraria tag)

Simon311:
@nicatronTg 
This:

> %000,255,000%Type [c/00FF00:/help] for a list of [c/FFFF00:commands].

Will not work. Text-wise, it will output:

> Type [c/00FF00:/help] for a list of [c/FFFF00:commands].

Instead of

> Type /help for a list of commands.

More over, I've spotted yet another issue with this.
The item tags, as well glyph tags, etc will not work in any colored string if TShock color tags are used. There is no workaround this behavior other than using Terraria-tags. Example:

>%255,000,000%This server is running [i/s1:29]TShock for Terraria[i/s1:29].

will not work, instead of heart items it will display the text. Since TShock tags are not closed, there's no workaround.

Essentially, this PR is saying "use only one type of tags or fuck off", and I am not ok with this.

Simon311:
This whole issue started because one dude could not follow the gudelines properly, and now we're stripping shit from TShock. How great, love you all.

nicatronTg:
@Simon311 I've updated my code review to reflect the issues described. However, while @WhiteXZ's contribution to TShock by way of commit 09bbf8abe8d6156f64f619f7cf3442dc42da4f26 was technically against the guidelines, the guidelines also say that no reverts should happen without sign-off from another contributor. Furthermore, while I understand and appreciate your frustration with this issue, it's only a temporary setback that isn't reflective of our entire process. This is a small, one time issue, and can easily be repaired through collaboration. Profane language doesn't help this discussion one bit.

Since this PR has requested changes, it's worth discussing several possibilities:

1. We can keep White's commit and add this PR (which fixes at least some of the original functionality while nudging people in the right direction).
2. We can create another PR to revert White's commit, but that creates the bug which started the issue in the first place.
3. We can not accept this PR, keep White's commit, and encourage third party plugin development (a less than optimal option).

I would argue that 1 is the best option, given that changes are requested and can happen to ensure optimal functionality. I might add that re-writing a config file for someone would probably be the best idea here: convert it fully to the new format and reload rather than applying on the fly. That seems the _best in my eyes_ but is up for discussion.

nicatronTg:
_Humans are fallible and make mistakes. Mistakes can always be fixed. TShock code isn't worth chopping people's heads off over any day._

Simon311:
@nicatronTg Well, the buggy behavior is a regex fault, which I did not even want to check to begin with. Making this work with regex will probably require a super-freaky regex.

Yet another reason to return to the old system is the fact that the new chat packet still has a color field, 3 bytes long. Allowing those 3 bytes to be wasted while occupying significantly larger amount of bytes to format the message Terraria-way is a huge waste. Even Re-Logic seem to understand that, since they've kept that field.

I think a good option is to revert White's commit then fix the bug without changing the whole system.

nicatronTg:
> Allowing those 3 bytes to be wasted while occupying significantly larger amount of bytes to format the message Terraria-way is a huge waste. Even Re-Logic seem to understand that, since they've kept that field.

While this is technically true, I have my doubts about ReLogic's intentions with leaving or adding things to the protocol. While we know that Terraria's network code is extremely inefficient, I don't think it's worth splitting hairs over 3 bytes, but that's just me.

> I think a good option is to revert White's commit then fix the bug without changing the whole system.

The way I see it, if @DogooFalchion updates this code to be a better backwards compatibility layer I'm happy with it. I want to err on the side of not wasting as much work as possible, but that's ust me. If you want to open a PR to revert @WhiteXZ's commit that's fine, but I, personally, am not going to be doing that.

Simon311:
@nicatronTg The old system would only use 3 bytes. The new one will use 3 bytes (not used, wasted) + 11 bytes for the [c/000000:] tag, which is 14 bytes vs 3 bytes.

nicatronTg:
@Simon311 right, but this is a one-time batch message sent on login, which is negligible over the long run. We're talking a minor increase of bytes one time -- it's not like we're changing every message we ever send to use these color tags resulting in a huge amount of extra overhead on each chat packet sent from the server, for example. Bytes are cheap when they're only sent once.

Simon311:
@nicatronTg You're trying superhard to ignore the fact that this solution is crap in every way and I fail to see why.

nicatronTg:
@Simon311 well _this solution_, the one that I've flagged as needing changes, does not work. A future solution in this PR can solve almost all of the issues side some negligible bytes.

There is a school of thought that you [don't bite newcomers](https://en.wikipedia.org/wiki/Wikipedia:Please_do_not_bite_the_newcomers), and I think that same school of thought should apply to all pull requests. I'm not saying anything about newness here, but I am saying that someone has put _real time and effort_ into trying to get this solution to work, and throwing it out is mean on our parts. We've found some flaws, want them changed, etc., but overall moving to the Terraria format does, in the long run, improve the codebase. In a 4.4 or 5.0 release we can just cull the old code and have everyone on the new tag system which simplifies our maintenance in the long run. I digress, however. On this particular pull request, I think that if the author wants to make changes that conform to our specs (working in the described configs above) then I'm happy. **I don't want to waste someone's hard work, and I don't want to lose a potential contributor.**

Simon311:
@nicatronTg I don't have any problems with @DogooFalchion, I can see that he wants to help and all, he's cool. I am sorry about times I've said bad things about him. Do you have any problems with him?
This will require more work anyway, the question is will @DogooFalchion continue implementing a bad solution and get it merged because you "don't bite" resulting in a 10 times better MOTD plugin being created by someone else, or will @DogooFalchion decide that fixing the original bug is the real concern and PR a good solution.

nicatronTg:
@Simon311 as far as I'm concerned with the fixes I described, I'm happy with providing an upgrade path to the new format across the board, as the byte load is not meaningful to me, and the old format was kinda bad in the first place.

As I said, if you want to submit a PR reverting white's change, you can, but I'm not going to revert nor approve that.

To summarize key goals in this PR:

1. Ingest old TShock format verbatim.
2. Interpret TShock format the way it currently works, by gap filling and re-writing old config to new style (even if that means chaining several color codes together).
3. Save to motd.txt to completely remove old format from existence. Make conversion as easy as possible.

Then, we can remove the converter in 4.4 or 5.0 once everyone is up to snuff.

DogooFalchion:
The motd has the test I originally used as well as your sample @Simon311   Do you have any other suggestions?

http://puu.sh/rSZCZ/d79daa00eb.png

nicatronTg:
@DogooFalchion While you said it's a PITA I still think upgrading everything to the new format is the way to go, but this works for me.

Issue #1100, opened 2015-08-26
[SSC] Passive SSC


Ijwu:
```
darkunderdog [12:56 PM] 
Is there a plugin for SSC lets people import their Character data from their file into the server?

darkunderdog [1:00 PM]
I like SSC for plugins that use SSC like Jumppads and Forcefields - But also want to offer the flexability for people to keep their character inventories.

darkunderdog [1:00 PM]
Is this possible / does a plugin exist to do this?

ijwu [1:03 PM] 
It is possible, but not in a plugin

ijwu [1:03 PM]
The issue is that they have to somehow get you their player file

ijwu [1:03 PM]
Which is simply not possible in a plugin or within Terraria

ijwu [1:03 PM]
It is possible to have your own service which allows players to upload their file (to a website for example) and then their data is inserted into your DB

darkunderdog [1:07 PM] 
@ijwu: So I know there are plugins like /loadout that @popstarfreas uses that takes a users character data and saves it to a database so they can switch out different loads. couldn't something be implemented prejoin to auto save there character data into the database then use that for SSC?

ijwu [1:07 PM] 
Hmm, didn't think about that. Yeah, possibly.

ijwu [1:07 PM]
I'd have to take a look at SSC more in depth

ijwu [1:08 PM]
But it might be possible to have SSC not clear a person's inventory on join

ijwu [1:08 PM]
and instead just track what they joined with
```

tysonstrange:
Why don't you just turn off SSC or use the Ignore SSC perm?

darkunderdog:
darkunderdog [1:00 PM]
I like SSC for plugins that use SSC like Jumppads and Forcefields - But also want to offer the flexability for people to keep their character inventories.

also Ignore SSC perm while works to allow people to use their Character Inventory it does not allow people to make any changes to their Character Inventory - For example if in game you get a few items then exit - when you come back those items are gone.

bippity:
The awkward moment when you thought this feature already existed ._.

nicatronTg:
Part of #929 maybe? Either way, flagging this as PR wanted too.

nicatronTg:
Unfortunately, this issue title didn't properly help convey what really needs to happen for this fix, which is a semi-passive SSC system with bi-directional data flow. I detailed this more in #929 but I'm going to say that #1321 is still an adequate fix for most server owners, as it enables characters to have their SSC data imported when they join, which lets them start "where they were prior to joining" while still keeping the benefits of SSC.

[...]

PR #1321, opened 2016-10-23
Enable user to upload their data from when they join as SSC

DogooFalchion:
This will allow a user to run /uploadssc that will upload their characters data when they joined as their actual SSC data.  This is to address #1100.  This is controlled by a permission, tshock.ssc.upload.  It also adds the functionality to upload arbitrary playerdatas as their own, how much this was needed is unclear.  We could always remove that function and just use the existing method someway.  Let me know your suggestions~

### DIFF
--- a/TShockAPI/DB/CharacterManager.cs
+++ b/TShockAPI/DB/CharacterManager.cs
@@ -131,15 +131,15 @@ namespace TShockAPI.DB
 			string initialItems = String.Join("~", items.Take(NetItem.MaxInventory));
 			try
 			{
-				database.Query("INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, spawnX, spawnY, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8);", 
+				database.Query("INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, spawnX, spawnY, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8);",
 							   user.ID,
 							   TShock.ServerSideCharacterConfig.StartingHealth,
 							   TShock.ServerSideCharacterConfig.StartingHealth,
 							   TShock.ServerSideCharacterConfig.StartingMana,
-							   TShock.ServerSideCharacterConfig.StartingMana, 
-							   initialItems, 
-							   -1, 
-							   -1, 
+							   TShock.ServerSideCharacterConfig.StartingMana,
+							   initialItems,
+							   -1,
+							   -1,
 							   0);
 				return true;
 			}
@@ -159,7 +159,7 @@ namespace TShockAPI.DB
 		public bool InsertPlayerData(TSPlayer player)
 		{
 			PlayerData playerData = player.PlayerData;
-			
+
 			if (!player.IsLoggedIn)
 				return false;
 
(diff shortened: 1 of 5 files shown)


TShockAPI/DB/CharacterManager.cs
```diff
@@ -131,15 +131,15 @@ public bool SeedInitialData(User user)
...
-				database.Query("INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, spawnX, spawnY, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8);", 
+				database.Query("INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, spawnX, spawnY, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8);",
 							   user.ID,
 							   TShock.ServerSideCharacterConfig.StartingHealth,
 							   TShock.ServerSideCharacterConfig.StartingHealth,
 							   TShock.ServerSideCharacterConfig.StartingMana,
-							   TShock.ServerSideCharacterConfig.StartingMana, 
-							   initialItems, 
-							   -1, 
-							   -1,
```
Patrikkk: Before pushing, if possible, try to review your code completely you are about to push so things like this don't get into the PR. It's only to have the PR clean and the changes easy to see. :) (Since no actual change happened here, but it shows up only because an extra space got in at the end of the lines)

Patrikkk:
I *might* have noticed a bug in the feature. 
Example:
Player joins the server, they log in to their account. They execute ´/uploadssc´ which overwrites their already existing SSC data in the DB with their join-data.  They walk around a bit. Autosave can happen, or they leave and their SSC gets saved. So basically it is like they have never executed ´/uploadssc´ because their current character data will get written to the DB.
I'm thinking this command should be allowed only before logging in. Or when they execute the command their character should be updated.

DogooFalchion:
PlayerData.RestoreData(TSplayer) will overide their default SSC with what is stored in the DB, correct?  In which case after running this command their original vanilla inventory and their actual SSC inventory will match their join data one.

Patrikkk:
You are correct, Sir :) I've overlooked.
To be more exact though. When you use PlayerData.RestoreCharacter() basically their `TPlayer` variables get updated by the data inside PlayerData. 
So yeah, in the current context `args.Player.DataWhenJoined.RestoreCharacter(args.Player);` is functional and correct.

nicatronTg:
@Patrikkk I dunno why you didn't test this in-game first, but it works fine with leaving and joining. You should have tested it first then commented, imho.

The one thing to consider (which I don't consider blocking on this) is that someone can repeatedly run the upload command to continually update their data. Obviously, this would be up to server owners to setup in such a way that they either move people out of the group that uploads ssc data or implement some other form of protection against repeated uploads. Either way, this doesn't bother me much -- I don't know a good way to implement a "one time use" command in TShock without retrofitting the database or something else.

Patrikkk:
Sorry @nicatronTg. I'm unable to download and test the code at the moment :(   (It is not a requirement that reviewing code can only be done by actually running it, no?)

And yeah I agree. Tbh this is the only TShock command I met with so far which is "one time use". 
Guess we could implement that in TShock 5 though so plugins can also easily have one-time-use commands.

nicatronTg:
> Guess we could implement that in TShock 5 though so plugins can also easily have one-time-use commands.

This kinda sounds ghastly.

> (It is not a requirement that reviewing code can only be done by actually running it, no?)

I dunno that we've ever set guidelines other than it's your neck on the line if you merge something that's broken.

DogooFalchion:
Ok, change to a third person supportable command in place.  This way they can issue the command to an admin and an admin can ~~overwrite their SSC data~~ upload the data for other players.

nicatronTg:
cc @NyxStudios/tshock

nicatronTg:
Despite #1100's changing scope, I still 100% support merging this. Waiting on final code check from someone else prior to merge (cc @NyxStudios/tshock).

Patrikkk:
Reviewed new PR. Lgtm

Ijwu:
LGTM. Seeing as @nicatronTg is on the go-ahead for this I'll merge this in.

PR #1363, opened 2016-12-26
Players shouldn't be able to change their groups through /user commands

ProfessorXZ:
Even though the `tshock.superadmin.user` permission is meant for superadmins only there are still people who do not follow that "rule". This lets mods give themselves superadmin powers, which is "BAD BAD". The check should be there nevertheless as I see no reason why you would let anyone change their own permissions.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1013,6 +1013,12 @@ namespace TShockAPI
 				return;
 			}
 
+			if (!args.Player.IsLoggedIn && args.Player.RealPlayer)
+			{
+				args.Player.SendErrorMessage("You are not logged in.");
+				return;
+			}
+
 			string subcmd = args.Parameters[0];
 
 			// Add requires a username, password, and a group specified.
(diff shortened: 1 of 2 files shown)


TShockAPI/Commands.cs
```diff
@@ -1109,6 +1109,12 @@ private static void ManageUsers(CommandArgs args)
 
 				try
 				{
+					if (user.Name == args.Player.User?.Name)
```
WhiteXZ: Why doesn't this check `if (user == args.Player.User)`?

ProfessorXZ: Because the command itself never checks the `TSPlayer.IsLoggedIn`, thus I assumed there might be people who'd like to give /user access to guests. This would be _really_ stupid yes, but I see it as "prevention is better than cure"

nicatronTg:
While I definitely understand that this is a problem, I've actually used systems in the past that use this as a "gate" that prevents accidental command use except without a hurdle to jump over.

PR #1367, opened 2016-12-29
Document and fix the fishing quest count glitch in SSC.

DogooFalchion:
Problem:
We were sending the correct packet in a state where the client was half connected.  The terraria handshake was not yet complete, and the netmessage protocol says to drop a packet to players not fully connected if the remoteclient is -1. 

Remoteclient == -1 means send this packet to everyone.  Thus, when we sent the fishing quest packet to all players, the actual player it was for never got sent it.  By changing the remoteclient to the specific player, the packet will always go out.  I do not believe we need to send this packet to everyone, but since we already were I left it in.

Solution:
Change the remoteclient in RestoreCharacter in PlayerData.cs to the player in question as we did with every other packet.

### DIFF
--- a/TShockAPI/PlayerData.cs
+++ b/TShockAPI/PlayerData.cs
@@ -438,8 +438,17 @@ namespace TShockAPI
 			{
 				player.TPlayer.buffType[k] = 0;
 			}
+
+			/*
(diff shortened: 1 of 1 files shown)


DogooFalchion:
This fixes https://github.com/NyxStudios/TShock/issues/1345

PR #1368, opened 2016-12-30
Fix for trap doors.  

### LINKED ISSUE #1364: Two bugs from the second prerelase.
So I tried the pre relase 2 of the 4.3.21 and all the trapdoors stopped working. We can't close them and open them makes them dissapear. I taked out the mod and they worked again and the dissapeared trapdoors appeared again.
Also, we used the Super Absorbant Sponge, but the mod detected it as a cheat.

DogooFalchion:
Also add in protection in the tile conversion logic to not needless trigger sendtilesquare.  Remove existing trapdoor bypass.

This should fix #1364.

I added a check for any tiles changing from closed to open or from nothing to closed/open or closed/open to nothing.  This protects us as best we can, afaik.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -1719,15 +1719,7 @@ namespace TShockAPI
 			var tileX = args.Data.ReadInt16();
 			var tileY = args.Data.ReadInt16();
 
-			bool isTrapdoor = false;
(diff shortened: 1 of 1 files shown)


WhiteXZ:
Tagging @NyxStudios/tshock for any second opinions.
I'm content to go with this

Issue #1350, opened 2016-12-16
Spawnmob certain mobs do not spawn.


Marcus101RR:
Source: Pre-Release 2 of TShock Mintaka 4.3.21 for Terraria 1.3.4.4 (APIv2.0, OTAPI v2.0.0.15)

Certain mobs such as (Eater of Worlds, Worms Types) do not seem to spawn. I will be testing more to find any other mob types.

[Image]
![image](https://cloud.githubusercontent.com/assets/8344293/21275866/b9d07846-c39c-11e6-8f30-1feaf0c5ce17.png)

ProfessorXZ:
Just to keep track of the issue: 

```

enerdy [11:46 PM]  
So not exactly our problem, most likely could be added as a feature

[11:46]  
Can add the [PR Wanted] tag

marcus101rr [11:56 PM]  
It worked before

[11:56]  
why did it stop now?

[11:57]  
Also, it doesn't work on SB

professorx [11:57 PM]  
@enerdy Maybe, but don't jump to conclusions

[11:57]  
Worm AI types caused issues in the earlier releases

enerdy [11:57 PM]  
I really, really doubt its a tshock issue

[11:57]  
Or rather in this case its more of a debate about the scope of the command

professorx [11:57 PM]  
The last time I checked the code TSAPI did weird stuff

marcus101rr [11:57 PM]  
It worked on old tshock all the time.

professorx [11:57 PM]  
Like, if there are no players in the corruption zone it just despawned

[11:58]  
But apparently vanilla doesn't do that
````

Kojirremer:
Skeletron does not spawn either. Prime does without issue.

tlworks:
EoW, Moon Lord, Cultist, Big Mimics, Dune Splicers, Wyvern, Phantasm Dragon, Some new monsters get transformed into older ones.

gl

ProfessorXZ:
Was resolved in [9050543](https://github.com/Pryaxis/TShock/commit/90505435bafefac46b50dee49513fc48d96530de). Closing

PR #1381, opened 2017-01-19
Fix /spawnboss to allow summoning all the bosses.

DogooFalchion:
I took a look at the decompiled SpawnOnPlayer which is what is used for natural spawn bosses.  I backported those changes into the TServerPlayer SpawnNPC function, wrapped the 200 npcid limit, and only do special mob name look ups if the name for that type is different from what the mobs display name is i.e. Eater of Worlds vs. Eater of Worlds Head. 

As far as I can tell this fixes most of the npc spawn issues, however, someone who knows exactly what non-boss mobs failed to spawn before should test this fix.  I was able to run /sb * which resulted in all sorts of bosses spawning including the moon lord, EoW, and bee queen.

This fixes #1381 and #1350

### DIFF
--- a/TShockAPI/TSServerPlayer.cs
+++ b/TShockAPI/TSServerPlayer.cs
@@ -142,18 +142,26 @@ namespace TShockAPI
 			TSPlayer.All.SendData(PacketTypes.TimeSet, "", dayTime ? 1 : 0, (int)time, Main.sunModY, Main.moonModY);
 		}
 
-		public void SpawnNPC(int type, string name, int amount, int startTileX, int startTileY, int tileXRange = 100,
+		public void SpawnNPC(int type, string name, int amount, TSPlayer player, int tileXRange = 100,
 			int tileYRange = 50)
 		{
 			for (int i = 0; i < amount; i++)
 			{
 				int spawnTileX;
 				int spawnTileY;
-				TShock.Utils.GetRandomClearTileWithInRange(startTileX, startTileY, tileXRange, tileYRange, out spawnTileX,
+				TShock.Utils.GetRandomClearTileWithInRange(player.TileX, player.TileY, tileXRange, tileYRange, out spawnTileX,
 															 out spawnTileY);
 				int npcid = NPC.NewNPC(spawnTileX * 16, spawnTileY * 16, type, 0);
-				// This is for special slimes
-				Main.npc[npcid].SetDefaults(name);
+
+				if (npcid < 200)
+				{
+					// This is for special slimes
+					if (Main.npcName[type] != Main.npc[npcid].displayName)
+						Main.npc[npcid].SetDefaults(name);
+					Main.npc[npcid].target = player.Index;
+					Main.npc[npcid].timeLeft *= 20;
+					NetMessage.SendData(23, -1, -1, "", npcid, 0f, 0f, 0f, 0, 0, 0);
+				}
 			}
 		}
 
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -2156,83 +2156,83 @@ namespace TShockAPI
 					foreach (int i in npcIds)
 					{
 						npc.SetDefaults(i);
-						TSPlayer.Server.SpawnNPC(npc.type, npc.name, amount, args.Player.TileX, args.Player.TileY);
+						TSPlayer.Server.SpawnNPC(npc.type, npc.name, amount, args.Player);
 					}
 					TSPlayer.All.SendSuccessMessage("{0} has spawned all bosses {1} time(s).", args.Player.Name, amount);
 					return;
 				case "brain":
 				case "brain of cthulhu":
 					npc.SetDefaults(266);
-					TSPlayer.Server.SpawnNPC(npc.type, npc.name, amount, args.Player.TileX, args.Player.TileY);
+					TSPlayer.Server.SpawnNPC(npc.type, npc.name, amount, args.Player);
 					TSPlayer.All.SendSuccessMessage("{0} has spawned the Brain of Cthulhu {1} time(s).", args.Player.Name, amount);
 					return;
 				case "destroyer":
 					npc.SetDefaults(134);
 					TSPlayer.Server.SetTime(false, 0.0);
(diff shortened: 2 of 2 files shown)


nicatronTg:
This guy is a pretty good developer. I highly suggest we consider merging
this PR today.

OnsenManju:
Just curious,
`/spawnmob wyvern` won't summon ordinary _Wyvern_ but _Etherian Wyvern_ because [`TShock.Utils.GetNPCByName("wyvern")`](https://github.com/NyxStudios/TShock/blob/b4f37a60c90e114781ce5e983ffced4b30b50363/TShockAPI/Utils.cs#L423-L426) returns type 558?

```
Mobs contained "wyvern" in its name:
type Main.npcName      NPC.name        NPC.displayName
---- ----------------- --------------- -----------------
 87  "Wyvern"          "Wyvern Head"   "Wyvern"
 88  "Wyvern"          "Wyvern Legs"   "Wyvern"
 89  "Wyvern"          "Wyvern Body"   "Wyvern"
 90  "Wyvern"          "Wyvern Body 2" "Wyvern"
 91  "Wyvern"          "Wyvern Body 3" "Wyvern"
 92  "Wyvern"          "Wyvern Tail"   "Wyvern"
558  "Etherian Wyvern" "Wyvern"        "Etherian Wyvern"
559  "Etherian Wyvern" "Wyvern"        "Etherian Wyvern"
560  "Etherian Wyvern" "Wyvern"        "Etherian Wyvern"
```

I think `GetNPCByName()` should use `npc.displayName` (not `npc.name`) for searching..

DogooFalchion:
That has nothing to do with this pull request.  Thats a fundamental change to the TShock API.

OnsenManju:
Ah sorry, I've mentioned the last part of changes in TSServerPlayer.cs.
Okay, I'll do so :)

WhiteXZ:
Awaiting testing before approving

DogooFalchion:
Honestly, since the function is internal to TServerPlayer, and only used with player params in tshock, and require a player to run a command to execute, to me it makes sense to reduce the number of arguments.  Personally, anything with more than 3 arguments is already poorly designed, or a very rare situation that requires it.

That being said, I think the solution, assuming we want to keep that original behavior in tact (which may result in the bug coming back for that type of spawning), is to add a second signature that only does the netdefaults fix.

My opinion is that mob spawning commands should be split out of tshock into a gizmo, much like was suggested back in the old days of TShock, but everyone whined /shrug.

Since it is silly to keep this bug around while we discuss the best solution, I will go ahead and add another commit to add the old signature back.  Perhaps we can open another issue/discuss it among the community whether to moving monster utilities to a mob gizmo.  Gizmos will be shipped with the tshock release, after all...

Patrikkk:
I was only referring to the method in my review, not the sm/sb commands it self. I agree that having more command parameters should be avoided.
>That being said, I think the solution, assuming we want to keep that original behavior in tact (which may result in the bug coming back for that type of spawning), is to add a second signature that only does the netdefaults fix.
You don't necessarily have to create a second signature and go back to the old code. You could just make the method accept an extra parameter, `int target`, use that for passing player index and keep the X and Y ints. This way no feature is killed, and your fix should work as intended. 
About the gizmo topic, Yeah sure, I'm not against it. Could open a discussion about it. (Unless there is one somewhere already)

DogooFalchion:
Why would you recommend having one function handle spawning on players and at positions, resolved by adding another extra parameter to a function we've already agreed has too many parameters.....

nicatronTg:
@Patrikkk seems to me like the obvious solution is to ask for an _overloaded_ method that takes a player and  _deprecation notice_ on the old function.

> You could just make the method accept an extra parameter, int target, use that for passing player index and keep the X and Y ints. This way no feature is killed, and your fix should work as intended.

This is the _worst of both worlds_.

TShockAPI/TSServerPlayer.cs
```diff
@@ -142,18 +142,26 @@ public void SetTime(bool dayTime, double time)
...
-				TShock.Utils.GetRandomClearTileWithInRange(startTileX, startTileY, tileXRange, tileYRange, out spawnTileX,
+				TShock.Utils.GetRandomClearTileWithInRange(player.TileX, player.TileY, tileXRange, tileYRange, out spawnTileX,
 															 out spawnTileY);
 				int npcid = NPC.NewNPC(spawnTileX * 16, spawnTileY * 16, type, 0);
-				// This is for special slimes
-				Main.npc[npcid].SetDefaults(name);
+
+				if (npcid < 200)
+				{
+					// This is for special slimes
```
nicatronTg: I would, in general, prefer a little more clarification as to what and why this happens, for future confused souls who look at the code and ask why this is here.

WhiteXZ:
Until we pull commands like this out of TShock, it feels silly to break existing functionality that is used by plugins without offering a way in which they can still function, especially without deprecating anything in advance.
I.e., plugins that spawn mobs at a specific (x,y) coord using the current signature can't function if the (x,y) is taken from a player object, and have no way around this other than creating fake players with fake positions.

Ugly though it is, an extended signature taking an x, a y, and a target, is the easiest way to provide a solution to this issue. A method that takes a player can easily wrap the aforementioned one.

nicatronTg:
I mean, I don't know if plugins are _really_ using this function, but one can never be too sure in this world.

ProfessorXZ:
https://github.com/Patrikkk/SpawnMobXY @nicatronTg. This is a useful peice of work. So honestly there should be a way to do this without breaking possible plugins like this one.

nicatronTg:
@ProfessorXZ that's a good example, so we should definitely maintain compatibility.

It's also out of date for the current API version.

hakusaro:
Update on this?

DogooFalchion:
A great example of an out of date plugin???
Making the API codebase better should trump maintaining bad code for plugins that may or may not be up to date.  Anyone is welcome to take the changes and put them into the existing codebase and pollute it more.

ProfessorXZ:
@hakusaro no idea what you want to do with this. Might be related to #1383

hakusaro:
Closing / stale / obsolete.

PR #1396, opened 2017-02-15
Add Evil Type option during world creation

### LINKED ISSUE #1386: Ability to choose between Corruption and Crimson when creating new worlds via TShock
Hi, like the title says, this used to be a feature, but it was lost in recent updates, and I was told to make an issue on here about it so that it might possibly become a feature again, so here I am.

My servers boot up fine on already-created worlds, but if I wanna make a new world and not randomize Crimson/Corruption, I have to do so through the game itself.

mistzzt:
This should fix #1386

Spaces were changed to tabs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 This is the rolling changelog for TShock for Terraria. Use past tense when adding new entries; sign your name off when you add or change something. This should primarily be things like user changes, not necessarily codebase changes unless it's really relevant or large.
 
 ## Upcoming Changes
+* Added evil type option during world creation (@mistzzt)
 
 ## TShock 4.3.22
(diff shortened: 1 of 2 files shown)


mistzzt:
`Main.SettingsUnlock_WorldEvil` was set during `Main.InitializeDirect` in OTAPI.

However, initialization of plugins is called in `PreInitialize` hook.

```
OTAPI.Hooks.Game.PreInitialize = () =>
			{
				HookManager.InitialiseAPI();
				//InitialiseAPI();
				InvokeGameInitialize();
			};
```
So I think it should be called in `PostInitialize` hook.

mistzzt:
ok. I finally realized that i have ignored that
```
if (Main.dedServ)
{
	return;
}
```
and `OpenSettings()` will never be called.

PR #1397, opened 2017-02-16
Correct document and add TSPlayer property for TeleportEventArgs

mistzzt:


TShockAPI/GetDataHandlers.cs
```diff
@@ -1159,12 +1159,16 @@ private static bool OnPlayerBuffUpdate(byte id)
...
 		/// </summary>
 		public class TeleportEventArgs : HandledEventArgs
 		{
 			/// <summary>
-			/// ???
+			/// The player who teleported
+			/// </summary>
+			public TSPlayer Player { get; set; }
+			/// <summary>
+			/// ID of the target player
```
WhiteXZ: Is this the ID of the TSPlayer object?

mistzzt: ID is the target player's id.
Player is the one who teleported using rod of discord or wormhole potion.

WhiteXZ: So when using a wormhole potion, ID is the player being teleported *to*,
and when using a RoD, ID should be the same as Player.ID?

mistzzt: After reading terraria's source code, I found that the id is always the player who teleported.
And server use the distance between x,y and players to find the player being teleported to.

So, i made a mistake

mistzzt:
Closed as the Player property is actually not needed.

PR #1414, opened 2017-03-21
Add banning user and ban date fields to ban info and list endpoints

deadsurgeon42:
Adds: `Ban.BanningUser` (string) and the date when the ban occurred `Ban.BanDateTime` (string)

Maybe return null instead of empty strings?


WhiteXZ:
`"unknown"` would probably be better than `null` or `""`

Issue #1371, opened 2017-01-04
/logout Command Permissions Not Granted by Default


aarongoulet:
Users are unable to use the /logout command by default, which could be problematic in some cases (such as superadmins who /auth after having already registered and logged in using a different account, which renders them unable to switch to newly created superadmin account).

ProfessorXZ:
I don't see this as an issue that needs fixing honestly. There probably is a solid explanation why this wasn't added. Personally groups shouldn't have any permissions to begin with. Another thing to note is that if you have access to /auth you most certainly have access to the console, which lets you add the permission to the desired group to make them log out.

kinvaris:
@ProfessorXZ it is an issue that needs fixing but maybe not right now. As @aarongoulet mentioned a user should be able to switch between accounts. In my opinion, the `default` user group should always have some rights e.g. logout, ...

nicatronTg:
I believe this was intentionally not added to the default permission set, because it allows servers who have login before join to ensure that all players connected are logged in.

I'm opposed to changing the default.

DogooFalchion:
1. LoginBeforeJoin does not enforce RegisterBeforeJoin, therefore a new player will still join as a guest, and /register.
2. If a server owner does not want players on the server to be able to logout, then they should change the default group for newly registered players or remove it from default.  (iirc TShock will not force overwrite existing groups...)
3. The obvious to me as an end user is that if I have /login then I should have /logout.  Before this wasn't an issue because /logout wasnt a command.  Now that it is, it kinda makes sense that the two would be tied together somewhat.

2cents is that it takes 20 seconds to make the change.  I suggest the team either close this ticket or allow the PR making the change.  Letting the issue idle for 2 weeks is lol.

OnsenManju:
I feel it might not a technical matter or reasonableness but a something like design concept of TShock though.
> The obvious to me as an end user is that if I have /login then I should have /logout. Before this wasn't an issue because /logout wasnt a command. Now that it is, it kinda makes sense that the two would be tied together somewhat.

If TShock will let users to able to logout by default in future, I also hope current 2 permissions (_tshock.account.login_ and _tshock.account.logout_) will be unified into one.

nicatronTg:
> The obvious to me as an end user is that if I have /login then I should have /logout. Before this wasn't an issue because /logout wasnt a command. Now that it is, it kinda makes sense that the two would be tied together somewhat.

That's logical. I agree with this sound reasoning.

[...]

PR #1417, opened 2017-04-02
Issue #1371: Grant logout permissions by default

aarongoulet:


DogooFalchion:
I suppose this works because default inherits from guest, but since a guest by definition is someone who isnt logged in I am opposed to it being a part of the guest group which by definition is someone who is not logged in (a guest can run logout which seems silly since they arent logged in :) ).

aarongoulet:
@DogooFalchion That makes sense; I probably should have looked a little more closely at the definitions before submitting my request. ;) I'll make the necessary corrections shortly!

aarongoulet:
@DogooFalchion After going back and looking at the source code, I remember the reasoning behind that design decision: the logout command has a built-in handler that helpfully tells you, "You are not logged in," when attempting to log out as guest, which, to a general user, is more helpful (less ambiguous) than an insufficient permissions message.  That being said, it IS a design decision, and I'll defer to your judgement; let me know if you'd still like it moved to the 'default' category.

WhiteXZ:
I have to agree that logout should probably not be in guest by default.
The user-friendly message is good to have, but not providing guests access to logout is a better design.
Depending on how hacky things would have to get, maybe a work-around could be found to keep the 'You are not logged in' message

aarongoulet:
@WhiteXZ From what I can tell it might need to get pretty hacky to change the message due to the fact that permissions are abstracted from the the actual command execution code.  That being said, I'll be happy to update my pull request and move the logout command from the 'guest' group as soon as I get a chance today (I've got to head to work in a few).  Thank you for the feedback!

hakusaro:
Can't tell if trolling

aarongoulet:
Not trolling, just a mistake on my part. I was trying to get a clean, fresh fork without losing this specific pull request, since my changes are so minor. I'm not 100% sure the best way to do this with GitLab.

aarongoulet:
By the way, changes made. Atom removed some trailing whitespace on some other lines, which was another reason I was trying for a clean branch.  Sorry for the confusion!

hakusaro:
Umm, yeah. You don't want to delete the branch associated with a PR. You can update by:

```
git remote add upstream git@github.com:NyxStudios/TShock.git
git fetch upstream
...[git will fetch the upstream remote and create new branches locally to match]...
git merge [whatever git created as the general-devel branch from upstream]
git push origin general-devel
[or to update this pr substitute the last command with:]
git push origin logout-permissions-change
```

aarongoulet:
Thanks! This is actually kinda embarrassing; I work with Git on a daily basis at work but we don't use a lot of the features that make it such an amazing tool.  I appreciate you taking the time to help a new contributor, @hakusaro!

hakusaro:
> I work with Git on a daily basis at work

Hilariously, this has been my usual experience. I know how to do some but not all features. Rebasing is still a nightmare to me and I can never get it to start and run correctly. Once I learned about ```git reflog``` my life became a lot better but I still only know tangentially when to use it. Cheers c:

aarongoulet:
Thanks. :) So, from what I am seeing it looks like the GitHub desktop client already updated my branch from the master  I reviewed my commits/files changed and it looks about right. Not sure why automatic merging is blocked though.

aarongoulet:
Verification:
https://github.com/aarongoulet/TShock/commits/logout-permissions-change

hakusaro:
> Not sure why automatic merging is blocked though.

Pending code review.

screen shot 2017-04-13 at 10 15 43 pm

Issue #1420, opened 2017-04-12
People should be able to rename groups


hakusaro:
Why isn't there already a /group subcommand to rename groups? This seems like a no-brainer.

Marcus101RR:
By renaming the groups, you must remember that anyone in the current group must be also renamed to that group in the User database.

WhiteXZ:
@Marcus101RR See #1422

PR #1422, opened 2017-04-19
Add support for renaming groups

ColinBohn:
New GroupManager DB interface and user command for renaming groups.
Addresses issue #1420.

### DIFF
--- a/TShockAPI/DB/GroupManager.cs
+++ b/TShockAPI/DB/GroupManager.cs
@@ -199,6 +199,24 @@ namespace TShockAPI.DB
 			group.Prefix = prefix;
 			group.Suffix = suffix;
 		}
+		public String RenameGroup(String name, String newname)
+		{
+			if (!GroupExists(name))
+				throw new GroupNotExistException(name);
+
+			if (GroupExists(newname))
+				throw new GroupExistsException(newname);
+
+			if (database.Query("UPDATE GroupList SET GroupName=@0 WHERE GroupName=@1", newname, name) == 1)
+			{
+				GetGroupByName(name).Name = newname;
+				return string.Format("Group \"{0}\" has been renamed to \"{1}\".", name, newname);
+			}
+			else
+			{
+				throw new GroupManagerException(string.Format("Failed to rename group \"{0}\".", name));
+			}
+		}
 
 		public String DeleteGroup(String name, bool exceptions = false)
 		{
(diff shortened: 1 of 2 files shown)


medik:
@hakusaro I feel confused after reading your answer to him. How do you want to alter the existing group membership? Is it to replace the group name key for each user with a group id?

hakusaro:
You would have to change it so that groups either aren't primary on key using something like a UUID in place or you'd have to update all groups at once to make sure that the primary key and all others change at once.

rautamiekka:
Just use integers as primary key which maps to the group name, the standard way everywhere.

hakusaro:
Paging @NyxStudios/tshock to approve and merge this.

QuiCM:
Has this been tested at scale?

hakusaro:
@QuiCM not by me

TShockAPI/DB/GroupManager.cs
```diff
@@ -199,6 +198,39 @@ public void UpdateGroup(string name, string parentname, string permissions, stri
...
+				Group oldgroup = GetGroupByName(name);
+				Group newgroup = new Group(newname, oldgroup.Parent, oldgroup.ChatColor, oldgroup.Permissions);
+				groups.Remove(oldgroup);
+				groups.Add(newgroup);
+
+				// Update all Users in DB to reflect group name change.
+				database.Query("UPDATE Users SET Usergroup=@0 WHERE Usergroup = @1", newname, name);
+
+				// Update all players on server to reflect group name change.
+				foreach (var player in TShock.Players.Where(p => p != null && p.User != null && p.Group == oldgroup))
```
ProfessorXZ: I don't think that this will work entirely. Unregistered users do not have a `User` object bound to them, so the changes will not reflect the default (guest) group. Was this intentional?

ProfessorXZ: So I compiled the code and gave it a test. Permissions and colors are kept in sequence, though prefixes and suffixes seem to be reset (they're empty). I've also confirmed what I said above, the changes do not apply to the guest group. One more thing; I think it should be taken into consideration that the default group's name might change and thus `Group.DefaultGroup` should be updated accordingly. Otherwise, good job 👍 

Here's a screenshot of the test: http://i.imgur.com/pW2Eyrn.png. Might be a good idea if someone else tested it too.

ProfessorXZ:
@hakusaro changes requested 2 weeks ago. This is waiting on Colin.

hakusaro:
Please tag someone if this pull request gets worked in the future. Closing due to staleness.

ProfessorXZ:
I will take this issue upon myself and make a PR with the requested changes within the next 2 days.

PR #1426, opened 2017-04-21
Add Localization support

mistzzt:
* Add EnglishLanguage type to store English texts
* Itemban database now store English item name
* Command `/i` `/give` `/sm` `/tpnpc` can use both English and current language input
---
Currently some plugins need English item name, npc name or prefix etc.
Also, Itemban uses item name to identify banned items. If server owners use latest TShock version with different language from English, all itemban data will be invalid because they were store in English.

This commit will solve the problem by adding a type called `EnglishLanguage` which has methods providing English texts.
Now, Itemban uses English text, and `GetItemByName`, `GetNPCByName`, and`GetBuffByName` can handle user input in both English and other language.

### DIFF
--- a/TShockAPI/Utils.cs
+++ b/TShockAPI/Utils.cs
@@ -32,6 +32,7 @@ using Terraria.Utilities;
 using TShockAPI.DB;
 using BCrypt.Net;
 using Microsoft.Xna.Framework;
+using TShockAPI.Localization;
 
 namespace TShockAPI
 {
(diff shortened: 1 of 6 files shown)


TShockAPI/Utils.cs
```diff
@@ -344,16 +345,27 @@ public List<Item> GetItemByName(string name)
...
+					if (item.Name.ToLowerInvariant().StartsWith(nameLower))
+						found.Add(item.Clone());
+				}
+
+				string englishName = EnglishLanguage.GetItemNameById(i).ToLowerInvariant();
+				if (!String.IsNullOrWhiteSpace(englishName))
+				{
+					if (englishName == nameLower)
+						return new List<Item> { item };
+						return new List<Item> { item };
```
Ijwu: Duplicate line of code here.

mistzzt:
@Ijwu I dont know why we should not return string.Empty for not-found words.
In `Terraria.Lang`, method like `GetItemName` returns `LocalizedText.Empty` for same condition.

Ijwu:
The concern about empty strings was mostly a precaution. I think it may not be too serious and so with things addressed I'm up thumbing this. (+1)

PR #1429, opened 2017-04-25
Fix duplicate items shown in Utils.GetItemByName and a client crash

mistzzt:
discord

As discussed in discord, when using `/i irons` command, you'll get two equal match.

This will fix this problem by passing English name check if current culture is English.


mistzzt:
In build 1550, `/sm s` will send player a very very long text which would make client crash.
By using `PaginationTools.BuildLinesFromTerms` for `Utils.SendMultipleMatchError`, client will receive normal texts.

fix

PR #1431, opened 2017-04-26
Fix an itemban issue due to localization / Remove utf8 modification build process

mistzzt:
```csharp
database.Query("INSERT INTO ItemBans (ItemName, AllowedGroups) VALUES (@0, @1);",
 				               TShock.Utils.GetItemByName(itemname)[0].Name, "");
```
This will give the localized name of an item, but we need English name of the item.


mistzzt:
This build process does not have to be removed.

PR #1432, opened 2017-04-26
Fix an itemban issue due to localization

mistzzt:
```csharp
database.Query("INSERT INTO ItemBans (ItemName, AllowedGroups) VALUES (@0, @1);",
 				               TShock.Utils.GetItemByName(itemname)[0].Name, "");
```
This will give the localized name of an item, but we need English name of the item.

### DIFF
--- a/TShockAPI/DB/ItemManager.cs
+++ b/TShockAPI/DB/ItemManager.cs
@@ -65,7 +65,7 @@ namespace TShockAPI.DB
 			try
 			{
 				database.Query("INSERT INTO ItemBans (ItemName, AllowedGroups) VALUES (@0, @1);",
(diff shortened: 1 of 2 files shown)


TShockAPI/DB/ItemManager.cs
```diff
@@ -81,7 +81,7 @@ public void RemoveBan(string itemname)
 				return;
 			try
 			{
-				database.Query("DELETE FROM ItemBans WHERE ItemName=@0;", TShock.Utils.GetItemByName(itemname)[0].Name);
```
Ijwu: image

Ijwu: :+1:

PR #1438, opened 2017-05-02
Resolve duplicate negative ids items in Utils.GetItemByName

mistzzt:
This was discussed in Discord #support

Note: princess dress has two ids, so we cannot handle this currently.
```csharp
public const short PrincessDress = 865;
public const short PrincessDressNew = 1773;
```
Maybe we can add something like
```
More than one match found:
Princess Dress(865), Princess Dress(1773)
```

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -2293,7 +2293,7 @@ namespace TShockAPI
 			}
 			else if (npcs.Count > 1)
 			{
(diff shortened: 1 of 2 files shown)


Ijwu:
I like the proposed solution of leaving the error response in but displaying the item ID. Any thoughts @NyxStudios/tshock?

hakusaro:
lgtm_6

Issue #1447, opened 2017-05-11
/butcher [npcID] doesn't work in some cases


madman2021:
On a fresh server /butcher id works fine
Once the map progresses past a certain point /butcher id no longer kill the NPC
![capture](https://cloud.githubusercontent.com/assets/2904531/25974239/dcf344b8-369f-11e7-9a19-4630c1c1a9fa.gif)

PR #1448, opened 2017-05-11
Increased /butcher damage to defence fixes issue #1447

madman2021:
Please mind my ignorance with this fix, I don't understand how the number 0.5 was derived but changing it to 0.6 allows NPCs at later stages to be killed using the /butcher command.
I'm guessing after a certain stage the NPCs receive a defense buff ?

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5207,7 +5207,7 @@ namespace TShockAPI
 			{
 				if (Main.npc[i].active && ((npcId == 0 && !Main.npc[i].townNPC && Main.npc[i].netID != NPCID.TargetDummy) || Main.npc[i].netID == npcId))
 				{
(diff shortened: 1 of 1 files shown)


madman2021:
@Ijwu I can't work out a reason either, would it be better if I changed the commit and test with a flat damage of 500 lets say or just roll with this for now ?

Ijwu:
Lets just roll with this. At least this way an NPC is (mostly) guaranteed to die.

WhiteXZ:
I was the one who made the initial change - I don't recall exactly why, but I believe it was because when things like the Moon Lord came out, our old 99999 damage spam wasn't sufficient and I felt that adding more 9s was just lazy. Unfortunately it looks like the way defence works or something has changed, so this is no longer viable :(

PR #1457, opened 2017-06-08
Hooks for Item/Projectile/Tileban permission checks

deadsurgeon42:
Currently, PlayerPermission hook isn't used for those three ban systems, and if a plugin wants to handle those, it has to load before tshock and sidestep the current system.

This pr adds respective hooks in style of `PlayerHooks.PlayerPermission` (albeit with longer names).

### DIFF
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1254,19 +1254,33 @@ namespace TShockAPI
 				return Group.HasPermission(permission);
 		}
 
-		/// <inheritdoc cref="HasPermission(string)"/>
+		/// <summary>
+		/// Checks to see if a player has permission to use the specific banned item.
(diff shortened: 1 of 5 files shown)


TShockAPI/TSPlayer.cs
```diff
@@ -1253,6 +1253,24 @@ public bool HasPermission(string permission)
 			else
 				return Group.HasPermission(permission);
 		}
+
+		/// <inheritdoc cref="HasPermission(string)"/>
```
QuiCM: Sounds nitpicky, but this shouldn't really inherit documentation for `HasPermission(string)`, as the type required is not a string

deadsurgeon42: good now?

QuiCM:
Looks good to me

Patrikkk:
Lgtm 👌

PR #1459, opened 2017-06-19
Cleanup SaveManager

MarioE:
We don't really need to have a separate thread that constantly polls for saves. While this new Task-based method does not completely replicate the old approach, it shouldn't really matter as the saves executing "out of order" does not matter.

### DIFF
--- a/TShockAPI/SaveManager.cs
+++ b/TShockAPI/SaveManager.cs
@@ -18,6 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 using Microsoft.Xna.Framework;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Terraria;
@@ -28,9 +29,6 @@ namespace TShockAPI
 {
 	internal class SaveManager
 	{
-		private readonly object _lock = new object();
-		private readonly List<Task> _saveTasks = new List<Task>();
-
 		private SaveManager() 
 		{
 		}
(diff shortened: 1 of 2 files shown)


tylerjwatson:
Be wary that the waiting needs to happen in *some* thread, and in my opinion it should be on a separate Thread as was originally designed.  The issue I find with this Task`-based solution is using ThreadPool for background-based spinning, which is not a good use of the limited pool of threads.  The main use of the threadpool is for little tasks that do not justify creation of new threads as thread creation is expensive.  Since SaveManager runs for the lifetime of the application, it should rightfully be in its own thread, in my opinion.

MarioE:
Do we even need to wait, though? What's the point of having a separate thread if we just wait all the time? We might as well synchronously run it on the main thread. It's not like saving takes that long, tbh.

tylerjwatson:
Becuase the main thread has a game timer that needs to tick at 60 fps, or 16.667ms. Unless the save fits comfortably in that time (with leaving room for all the update hooks to run) then I'm not okay with it.  

I mean, I don't control the project, but merging it would be against my advice.

MarioE:
The thing is, though, it already blocks for most occurrences of saves.

tylerjwatson:
It does? Then what's the purpose of the thread?

MarioE:
Beats me; I left in the `wait` parameter to preserve compatibility. Otherwise I would've switched to a fire-and-forget sort of thing entirely

hakusaro:
Hi @MarioE @tylerjwatson what's up with this PR?

Issue #1470, opened 2017-07-09
Can't negate permissions with PlayerHooks.PlayerPermission


deadsurgeon42:
```cs
    public static void PlayerPermission(PlayerPermissionEventArgs e)
    {
      if (e.Handled) return;

      // result:
      // true = player does have permission, don't let tshock handle it
      // null = idk, let tshock handle it
      // false = player has the permission negated, let tshock know player doesn't have permission (but how)
      bool? result = UserInfo.HasPermission(e.Player, e.Permission);

      if (result == null)
        return;

      e.Handled = true;
      // e.Negated = (bool) result; (maybe?)
    }
```

Right now `PlayerHooks.PlayerPermission` is limited as there's only two ways of telling TShock what to do:

`e.Handled = true` 🡲 The player has permission, don't check or prevent the player
`e.Handled = false` 🡲 I don't know if the player has permission or not, check
`???` 🡲 *The player doesn't have permission, don't check but prevent the player*

This is good enough for adding permissions, but when you want to negate permissions from the player it falls short.

A simple solution could be changing `PlayerHooks.OnPlayerPermission` return type to `bool?` and adding a `bool Negated` property into `PlayerPermissionEventArgs` (and their cousins `Player(Item/Proj/Tile)banPermissionEventArgs`) which is checked if `Handled` is `false`:

```cs
var args = new PlayerPermissionEventArgs(player, permission);
PlayerPermission(args);
if (!args.Handled)
  return !args.Negated;
else
  return null;
```

Which is then checked in `TSPlayer.HasPermission`:

```cs
public bool HasPermission(string permission)
{
  bool? hookResult = PlayerHooks.OnPlayerPermission(this, permission);
  if (hookResult != null)
   return (bool) hookResult;

   if (tempGroup != null)
     return tempGroup.HasPermission(permission);
   else
     return Group.HasPermission(permission);
}
```

Tho this requires some refactoring (and an api tick?).

bartico6:
You overcomplicated it.
```var args = new PlayerPermissionEventArgs(player, permission);
PlayerPermission(args);
return args.Handled```
is enough. It'll either give back a null when a hook returns null, or the outcome.
Then you just check the hook outcome if it's null (and execute logic) or deny/allow based on the boolean non-null value.

deadsurgeon42:
But `HandledEventArgs.Handled` is `bool` not `bool?` ?

bartico6:
Why not make it one?

[...]

PR #1476, opened 2017-07-11
Turn the player hooks into a tristate #1470

deadsurgeon42:
https://github.com/Pryaxis/TShock/issues/1470
This adds a  new `PermissionResult` enum with 3 values that specify allowed, denied and unknown, which "unknown" means tshock should do the check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming Changes
 
+* API: Changed `PlayerHooks` permission hook mechanisms to allow negation from hooks (@deadsurgeon42)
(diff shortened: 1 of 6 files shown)


deadsurgeon42:
@QuiCM okay, 
@ProfessorXZ they do work the same, like
`ask hook -> allow/block item use -> if inconclusive check group/tempgroup`

ProfessorXZ:
Ok, I failed to see the database managers. The PR is still missing a changelog update nonetheless.

deadsurgeon42:
@Enerdy maybe rename the enum to `PermissionHookResult` and `Inconclusive` to `Unhandled`?

Enerdy:
`PermissionHookResult` feels slightly superfluous there, although it does clarify what the enum is for. I'll leave that one up to you.

And yes, lets not leave this issue in an inconclusive state -- no one would want that.

PR #1477, opened 2017-07-13
Implement IEquatable<User> for User class

deadsurgeon42:
Because right now equality checks use memberwise equality, and this causes inconsistent behavior for some reason.

image

**Expected behavior:** `s.User == user` returns `true` as they're the same user
**What really happens:** `s.User == user` returns `false` even tho they have equal ID and name

Since users are considered unique by `ID` and `Name` properties, this pull request implements equality using only those two members.

_and `ToString()` mapped to `Name` cos @Celant asked so_


Ijwu:
Why?

deadsurgeon42:
Oh right forgot to tell in the above screenshot is that those two user objects, while they're supposed to be same return `false` from `s.User == user`.

TShockAPI/DB/UserManager.cs
```diff
@@ -581,6 +581,67 @@ protected string HashPassword(string password)
...
+			return Equals((User)obj);
+		}
+
+		/// <summary>Serves as the hash function. </summary>
+		/// <returns>A hash code for the current <see cref="User"/>.</returns>
+		public override int GetHashCode()
+		{
+			unchecked
+			{
+				return (ID * 397) ^ (Name != null ? Name.GetHashCode() : 0);
```
MarioE: `Name` has a setter. It's not a good idea to allow this, especially if `User` is keyed in a dictionary.

TShockAPI/DB/UserManager.cs
```diff
@@ -581,6 +581,67 @@ protected string HashPassword(string password)
...
+		public bool Equals(User other)
+		{
+			if (ReferenceEquals(null, other)) return false;
+			if (ReferenceEquals(this, other)) return true;
+			return ID == other.ID && string.Equals(Name, other.Name);
+		}
+
+		/// <summary>Indicates whether the current <see cref="User"/> is equal to another object.</summary>
+		/// <returns>true if the <see cref="User"/> is equal to the <paramref name="obj" /> parameter; otherwise, false.</returns>
+		/// <param name="obj">An <see cref="User"/> to compare with this <see cref="User"/>.</param>
```
MarioE: Incorrect XML doc

Ijwu:
Really, I meant why is this necessary? Why isn't `User.Id == User2.Id` enough?

deadsurgeon42:
Because it misleads people to think two essentially same `user == user` would be true.

PR #1484, opened 2017-08-04
Change field visibility

### LINKED ISSUE #1466: tS' LinuxTcpSocket not consistent with OTAPI/TSAPI-modified TcpSocket
The custom class in TShock called "LinuxTcpSocket" is not consistent with the TcpSocket built into the modified TerrariaServer executable that you get by compiling TShock and its deps.

The inconsistencies are mainly field accessibility, with fields in TShock's Linux version being private, and the exact same fields being public in TcpSocket

Proof

Suggestion would be to either make LinuxTcpSocket fields public like they are in TcpSocket, or make TcpSocket fields private like they are in LinuxTcpSocket.

This is completely invalid for stock Terraria. The fields in TcpSocket are public only in TShock.

bartico6:
As requested. Fixes #1466

I'm late I'm sorry </3

### DIFF
--- a/TShockAPI/Sockets/LinuxTcpSocket.cs
+++ b/TShockAPI/Sockets/LinuxTcpSocket.cs
@@ -12,23 +12,23 @@ namespace TShockAPI.Sockets
 {
 	public class LinuxTcpSocket : ISocket
 	{
-		private byte[] _packetBuffer = new byte[1024];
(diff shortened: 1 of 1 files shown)


Celant:
👎  Took too long to submit

Celant:
Jk, LGTM!

👍

PR #1486, opened 2017-08-14
Add "less than 2gb ram" warning

### LINKED ISSUE #1485: TShock should display a warning if started with < 2GB of free system memory
Per #969 and related discussion, we should display a very prominent warning message if people are starting TShock with under the suggested amount of system ram. We already display this warning on the readme, but not many people read it. This would help resolve that.

bartico6:
Adds the warning requested in #1485.

Fixes #1485.


bartico6:
Failed due to a timeout fetching a dependency.... 🤔

TShockAPI/TShock.cs
```diff
@@ -329,6 +329,10 @@ public override void Initialize()
 
 				Log.ConsoleInfo("TShock {0} ({1}) now running.", Version, VersionCodename);
 
+				if (StatTracker.GetTotalSystemRam(ServerApi.RunningMono) < 2)
+				{
+					Log.ConsoleInfo("This machine has less than 2 gigabytes of RAM installed. Be advised that it might not be enough to run TShock.");
```
Ijwu: `ConsoleInfo` may not be prominent enough. Let's make this stand out.

Ijwu:
Also, I restarted the Travis build and it passes.

bartico6:
Apparently it has to track <FREE> memory, not total. Do you own a linux box with <2048mb free ram that you could run a version of tshock for me on?
I'd send you the binary in question and if it works I'll commit here so it can be merged

Ijwu:
No, but you can spend $3 and buy a VM on Amazon or Azure for a few minutes.

Celant:
@Ijwu And you'll refund the expense, right? :^)

Ijwu:
Yeah, sure. /shrug

hakusaro:
Just export an Amazon billing receipt and expense the Open Collective account for the brief period of time you had the server running to compensate yourself @bartico6 / @ijwu

ProfessorXZ:
@bartico6 what is the status on this?

hakusaro:
@bartico6 ping

bartico6:
ping pong the Available MBytes performance counter I wanted to use doesn't do jack shit for linux, on windows this works fine, however on mono it'll always return -1.

hakusaro:
```/proc/meminfo```?

**Comment by bartico6** [comment]
Isn't that still a pretty ugly/hacky solution that's not guaranteed to work (for instance) on Windows mono?

**Comment by hakusaro** [comment]
> Isn't that still a pretty ugly/hacky solution that's not guaranteed to work (for instance) on Windows mono?

I mean it's just a file right. You can see if /proc/meminfo exists, and if it doesn't, you can just fall back to checking the other method, and if that doesn't work, then you don't give a warning.

**Comment by hakusaro** [comment]
To be clear, it's easier and safer to see if /proc/meminfo exists with a trivial file api versus if you need to run a command, which may fail in unexpected ways and require more parsing, etc.

**Comment by hakusaro** [comment]
Also, windows on mono is...not really supported, imho.

**Comment by bartico6** [comment]
If you honestly think this is a good idea then sure, I can add that...
Please keep in mind it is temporarily on a backburner due to other projects I'm working on (while I'd hate to say it like that, TShock is lower priority than those, hope ya understand)

**Comment by bartico6** [comment]
Works on Linux, using the /proc/meminfo trick. Also changed the message to stand out.

File: TShockAPI/StatTracker.cs [hunk 2]
```diff
@@ -176,8 +177,39 @@ private PluginItem[] GetLoadedPlugins()
 			}
 			return plugins;
 		}
-
-		private long GetTotalSystemRam(bool isMono)
+		public long GetFreeSystemRam(bool mono)
```
hakusaro: Really prefer gaps between function end braces and start of new functions.

And XML documentation for new public members.

TShockAPI/StatTracker.cs
```diff
@@ -176,8 +177,39 @@ private PluginItem[] GetLoadedPlugins()
...
+				}
+				return -1;
+			}
+			else
+			{
+				var pc = new PerformanceCounter("Memory", "Available MBytes");
+				return pc.RawValue;
+			}
+		}
+		public long GetTotalSystemRam(bool isMono)
```
hakusaro: Really prefer gaps between function end braces and start of new functions.

And XML documentation for new modified members.

bartico6:
✅

bartico6:
@Ijwu input, pretty please? :hearts:

ProfessorXZ:
Has this been tested?

ProfessorXZ:
On a Linux based system*

hakusaro:
That screenshot was PuTTY, so yes.

bartico6:
Correct, tested on Ubuntu 17.04.

ProfessorXZ:
One thing that I would like to point out here are the commit messages; messages should be informative, this means that when someone else looks at your work in a later future they have a sense of what you did and what it was related to. It all makes perfect sense while you're working on the PR, but you might even find asking yourself what "Fixed ToString() override" meant 2 months later. Keep in mind that this was not meant to be hostile or anything, it was just a friendly advice or a request as you are definitely not the only person that did this, we all did.

Other than that, seeing all the request changes are implemented I'll merge this now.

bartico6:
Thanks for the advice fam 💓

PR #1493, opened 2017-09-19
Fix #1492

### LINKED ISSUE #1492: Time sometimes displays incorrectly at the turn of the hour
image
image
image

As shown above,using **/time** at the turn of the hour sometimes shows the minute counter as 60 instead of 00 as it should. Pretty straightforward, but thanks for reading!

koneko-nyan:
Fixing time display.
(time % 1.0) * 60.0 can be equal to 59.8, which gets rounded to undesirable 60 displayed in the time format


hakusaro:
@koneko-nyan please update changelog.md with your changes & name, otherwise good to go

PR #1496, opened 2017-09-21
Fixes #1420

### LINKED ISSUE #1420: People should be able to rename groups

ProfessorXZ:
This PR is simply an addon to #1422 as it only implements the changes I've requested a few months back. With that being said most credit goes to Colin.

## Changes

- Added support for renaming groups (/group rename)

- Added missing documentation for `GroupManager` methods

- Updated CHANGELOG.md

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -3077,25 +3077,25 @@ namespace TShockAPI
 					return;
 				case "rename":
 					#region Rename group
-				{
-					if (args.Parameters.Count != 3)
 					{
-						args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}group rename <group> <new name>", Specifier);
-						return;
-					}
+						if (args.Parameters.Count != 3)
+						{
+							args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}group rename <group> <new name>", Specifier);
+							return;
+						}
 
-					string group = args.Parameters[1];
-					string newName = args.Parameters[2];
-					try
-					{
-						string response = TShock.Groups.RenameGroup(group, newName);
-						args.Player.SendSuccessMessage(response);
+						string group = args.Parameters[1];
+						string newName = args.Parameters[2];
+						try
+						{
+							string response = TShock.Groups.RenameGroup(group, newName);
+							args.Player.SendSuccessMessage(response);
+						}
+						catch (GroupManagerException ex)
+						{
+							args.Player.SendErrorMessage(ex.Message);
+						}
 					}
-					catch (GroupManagerException ex)
-					{
-						args.Player.SendErrorMessage(ex.Message);
-					}
-				}
 					#endregion
 					return;
 				case "del":
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -2865,6 +2865,7 @@ namespace TShockAPI
 							"add <name> <permissions...> - Adds a new group.",
 							"addperm <group> <permissions...> - Adds permissions to a group.",
 							"color <group> <rrr,ggg,bbb> - Changes a group's chat color.",
+							"rename <group> <new name> - Changes a group's name.",
 							"del <group> - Deletes a group.",
 							"delperm <group> <permissions...> - Removes permissions from a group.",
 							"list [page] - Lists groups.",
@@ -3072,6 +3073,29 @@ namespace TShockAPI
 							args.Player.SendSuccessMessage("Color of \"{0}\" is \"{1}\".", group.Name, group.ChatColor);
 						}
 					}
+					#endregion
+					return;
+				case "rename":
+					#region Rename group
+				{
+					if (args.Parameters.Count != 3)
+					{
+						args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}group rename <group> <new name>", Specifier);
+						return;
+					}
+
+					string group = args.Parameters[1];
+					string newName = args.Parameters[2];
+					try
+					{
+						string response = TShock.Groups.RenameGroup(group, newName);
+						args.Player.SendSuccessMessage(response);
+					}
+					catch (GroupManagerException ex)
+					{
+						args.Player.SendErrorMessage(ex.Message);
+					}
+				}
 					#endregion
 					return;
 				case "del":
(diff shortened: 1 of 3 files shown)


ProfessorXZ:
This has been put to test and it seemed to work fine, but keep in mind that I do not run an actual server and the test was very primitive. https://i.imgur.com/0IwgboX.png

QuiCM:
Requesting testing from @deadsurgeon42 @Patrikkk @bartico6 and any others who can provide some large-scale testing.
Thank you for anyone who can do so

bartico6:
By large-scale testing you mean running it live on a server, or just using our database to check if it works properly? @QuiCM

QuiCM:
Using your db in a test environment should be sufficient

bartico6:
I'll get back to you tomorrow on that then, it's like 12:10AM here.
Anything specific you want me to do or "just try to break it"?

Ijwu:
@bartico6 Good practice for testing things should include a process that is something like this:
* Test the happy path. Does it work as intended if I do everything correctly?
* Test incorrect user input. Does it work as intended if I write in something incorrectly?
* Test malicious input. Does it work as intended if I'm a meanie and try to overload by putting in huge input? Maybe I put in SQL strings, does it cause an injection?
* Does it work in a normal environment? Pretend you're playing on a real server and use the command like you would a normal player. 

That's about the most I expect from a developer making their changes. Test a few cases of those items to be sure it works properly. We don't expect everyone to be a professional software quality assurance analyst, but I think this is a bit of good practice for testing your own stuff. 

Thanks for helping us out. :)

TShockAPI/Commands.cs
```diff
@@ -3074,6 +3075,29 @@ private static void Group(CommandArgs args)
...
+					try
+					{
+						string response = TShock.Groups.RenameGroup(group, newName);
+						args.Player.SendSuccessMessage(response);
+					}
+					catch (GroupManagerException ex)
+					{
+						args.Player.SendErrorMessage(ex.Message);
+					}
+				}
```
hakusaro: Indentations are inconsistent with rest of file in these braces and the ones below.

TShockAPI/DB/GroupManager.cs
```diff
@@ -200,6 +216,70 @@ public void UpdateGroup(string name, string parentname, string permissions, stri
...
+			{
+				throw new GroupNotExistException(name);
+			}
+
+			if (GroupExists(newName))
+			{
+				throw new GroupExistsException(newName);
+			}
+
+			if (database.Query("UPDATE GroupList SET GroupName = @0 WHERE GroupName = @1", newName, name) == 1)
```
hakusaro: All of these database calls need to be done in a transaction. There's way too much damage that could be done if the something in this chain breaks due to an unforeseen database problem that leaves everything in a super inconsistent state. Transactions give you the ability to abort if something goes wrong and roll back, and they should be 100% supported in SQLite and MySQL.

If this doesn't work in SQLite and MySQL, there needs to be a hard backup done and automatically restored if this operation runs.

TShockAPI/DB/GroupManager.cs
```diff
@@ -214,12 +294,18 @@ public String DeleteGroup(String name, bool exceptions = false)
...
+				throw new GroupManagerException("Failed to delete group '" + name + ".'");
+			return "Failed to delete group '" + name + ".'";
 		}
 
+		/// <summary>
+		/// Enumerates the given permission list and adds permissions for the specified group accordingly.
+		/// </summary>
+		/// <param name="name">The group name.</param>
+		/// <param name="permissions">The permission list.</param>
+		/// <returns></returns>
```
hakusaro: Any idea what this returns?

TShockAPI/DB/GroupManager.cs
```diff
@@ -200,6 +216,70 @@ public void UpdateGroup(string name, string parentname, string permissions, stri
...
+
+			throw new GroupManagerException($"Failed to rename group \"{name}\".");
+		}
+
+		/// <summary>
+		/// Deletes the specified group.
+		/// </summary>
+		/// <param name="name">The group's name.</param>
+		/// <param name="exceptions">Whether exceptions will be thrown in case something goes wrong.</param>
+		/// <returns></returns>
```
hakusaro: What does this return?

hakusaro:
(The first status check on this branch is broken because of a race condition or something in Travis. It picked up my invalid push and now thinks that's related to this branch, but it isn't).

bartico6:
@hakusaro I live in Valve Time, okay?

If you make any changes to the config.json file and then rename a group, the changes you had *will be lost*, as the config that's in memory will be written with a few changes. I'd consider reloading the config and then writing changes, unless that's intended.
Other than that, I tried breaking it in numerous ways and it seems to work:
> The usual use case (renaming a group that's already in use) worked A-OK even with a sqlite database that has 229 thousand registered users (it took about 2-3 seconds on a home computer to rename the default user group with such a big base, I'd consider that a success). ✅ 
Too many args? Reports syntax errors. ✅ 
Too few? Same. ✅ 
Exactly correct? Works as intended. ✅ 
Renaming nested groups that both parent something and are parented by something? Works ✅ 


Unless there's something you guys can come up with to try and break it, I think it's alright. <3

hakusaro:
> If you make any changes to the config.json file and then rename one of the basic groups, the changes you had will be lost, as the config that's in memory will be written with a few changes. I'd consider reloading the config and then writing changes, unless that's intended.

Literally any changes?

bartico6:
Yes, any changes will be lost as the config will be overwritten with the version that was loaded at startup. If you don't /reload after making the changes, then they will be lost when a group is renamed.

hakusaro:
I'm not sure how common it is to modify the config file during runtime without reloading and then doing things like this. You're definitely right, in that it probably is gonna be weird, but the idea of just making changes and not running /reload is also kinda weird imho.

hakusaro:
Proposed solutions to the config rewrite:

- Just reload the config before re-writing it.
- Detect if the saved version is different than the current version and load it.
- Just display a warning and require confirmation if changing the default group.

This makes me wonder if we write the config file anywhere else in TShock and lose changes. Probably worth investigating that too?

ProfessorXZ:
I don't think this is an issue with the rest of the code, but I'll check it out anyway. As for the solution I'll just read the config before re-writing as proposed.

hakusaro:
@Pryaxis/tshock someone...anyone?

PR #1509, opened 2017-11-20
Fix item frame not working properly

### LINKED ISSUE #1507: ITEM FRAME ERROR
## Any stack traces that may have happened when the issue occurred

I am playing with t.shock.
I am going to use the "Item Frame" to display items.
As soon as I hold the items and right click, the weapon is on display.
But as soon as I click on the displayed items, the items disappears!
"Item Frame" is different when placed on the floor and when hang on the wall.
When hang on the wall, it works normally.

There is no problem using the default server driver, not t.shock.

mistzzt:
Fixes #1507

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2109,7 +2109,9 @@ namespace TShockAPI
 						return true;
 					}
 					// If the tile is a pickaxe tile and they aren't selecting a pickaxe, they're hacking.
-					else if ((!Main.tileAxe[tile.type] && !Main.tileHammer[tile.type]) && tile.wall == 0 && ((args.TPlayer.mount.Type != 8 && selectedItem.pick == 0) && !ItemID.Sets.Explosives[selectedItem.netID] && args.Player.RecentFuse == 0))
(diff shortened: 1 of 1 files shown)


Kojirremer:
A snip from Discord:
-----
After trying hard for a while, here are my results. Bear in mind that I run my server with barely enough RAM, so results may not be the same for other testers.
### On latest unstable
Couldn't duplicate, but after several tries the item frame would eventually "eat" the item being displayed.
### On test build
I didn't manage to duplicate here either, and the item wasn't eaten. All appears to be well.
When trying to break a protected item frame without having region-editing permission, the item frame will remain unchanged from server's POV, but from client's POV it will break, then reappear instantly without its contents.
When trying to take what's within by right-clicking (while empty-handed) you will get the area protected message and the client may see the frame as empty, but it will still exist as usual server-side.
When doing the same as above, but while carrying something at the same time, the same as above will happen, except whatever you were carrying will be eaten by the frame.
Due to the fact that it doesn't appear to randomly eat the item under stress anymore I'd say the fix is good.
By the way, I only tested item frames. I think weapon racks were working normally.
On normal builds at least.

ivanbiljan:
Lgtm :+1:

Issue #1510, opened 2017-11-30
Disputed code from Commaster


hakusaro:
As recently mentioned in #1460, Commaster has made the accusation that we stole specific code introduced in one of his pull requests without crediting him. I'm going to quote everything he said here:

> Oh really, we are talking about giving credit? Well, that wasn't the case with #781 which were blatantly stolen in 16cf364 and 038f606.
Shall we start fresh? You remove https://github.com/Pryaxis/TShock/blob/general-devel/TShockAPI/Commands.cs#L1288-L1334 and https://github.com/Pryaxis/TShock/blob/general-devel/TShockAPI/Commands.cs#L1389-L1423 permanently from the codebase and then we can talk about my contributions.
But until then, keep stealing my work

After reviewing the diffs, I was able to see several similarities in how our code functions versus his. A good example of this is the eerie implementation of ```bool force = !args.Player.RealPlayer;``` which names the variable force and assigns it in a way that is atypical from the rest of the codebase. That being said, the ban code is somewhat of a fixed place system and it could be possible we've just accidentally written similar code to him.

Again, look at https://github.com/Pryaxis/TShock/pull/781/files#diff-72108b2e5b27cef0ef5199d364dafc54R1202 and https://github.com/Pryaxis/TShock/commit/038f6069397bcf73f23e8b1228de68f4ddd96542#diff-72108b2e5b27cef0ef5199d364dafc54R1163. Eerie similarity? Or just a coincidence?

Open source somewhat dictates that we should at least investigate this accusation at minimum, and perhaps look into rectifying it by either crediting Commaster or actually rebuilding this block of code/system.

Because this process is somewhat time consuming, I would like to open it up to the community to help identify similarities as well. If code looks eerily familiar across the PR and our codebase, it should be noted here and discussed.

If the issue devolves into a flame war it's going to get locked, though. Locking the issue would slow down the diff process and stifle debate, so I don't really want to do that.

bartico6:
You would be correct, the way `bool force` is defined is in fact really odd.
![xx](http://i.quake.tf/1512056230-1827219487.png)
In fact, so odd that it shows up in only two places of the entire codebase - the two disputed snippets.  
*With that said...* there's only so many ways you can implement banning an offline user (you *have* to fetch them from db, you *have* to check their known IPs and you *have* to check if they can ban said user).  
[...]

Issue #1511, opened 2017-12-02
Rewriting the ban system


QuiCM:
The TShock ban system needs improvements.

The aim of this issue is to discuss what these improvements will encompass.
Feel free to chime in with drawbacks to the current system, improvements you'd like to see, what's good, what's bad, etc.

Requesting input from @Zaicon @bartico6 @hakusaro @Ijwu @ivanbiljan @Pryaxis/tshock

QuiCM:
My quick 2c on what a new system needs:

* Relational persistence model - the current persistence model is, like all the TShock db tables, a mess. However TShock should manage the bare minimum required to function in order to achieve the below.
* Extensibility - The TShock side should just fire hooks and deal with storing basic information. Plugins should be able to customise the rest

hakusaro:
There are a thousand crappy ways to ban someone, and none of them are good. Here are the problems with the current ban system:

* Right now, we have a table that requires you to have IP as a primary key. This shouldn't be a limit.
* You can't "ban accounts" but you can ban character names. But if a user is offline, then the system bans their account name is if it were a character name. But you can log into a user account from a different character name on some servers, so this makes banning that user ineffective. An aside: it would also be a good idea to just be able to deactivate accounts (like disable them in the db).
* Players can join with their name as an IP address, but then banning them becomes ambiguous because it could be an IP or an IP name. Note: this can be solved at the command level but I didn't want to get into the woods of what would or wouldn't happen with it.
* There's ambiguity in basically everything. Are you referring to an online player? An offline player? An offline player while a different online player has that name?
* Utils has a ban method and so does the BanManager and they do different things and one calls the other.
* Every stupid field in the database is a string. Well, shoutouts to the IP field because it's an actual MySQL string that's char capped at 16 characters. So if Terraria ever goes to IPv6 we're hard coded to not support banning v6 addresses.
* Right now, if you ban a thing, the ban system bans anything it can find on that thing. We don't have the ability to just say "ban only this character name" or "ban only this IP" without also getting UUID or other fields filled in for us.
* You can't ban on IP ranges or DNS resolutions or hostnames.
* You can't ban on details about a character, which may uniquely identify them (e.g., costume, inventory item position and stack counts, etc. Basically all of the SSC data we know).

There are two primary approaches to bans, and I think we should stick to one or the other, not co-mingle both:

[...]

PR #1512, opened 2017-12-02
Rewrote the /ban add sub command

hakusaro:
This is a replacement of the /ban add command to accommodate #1510 and the general poor quality of the ban add command.

There are a couple catches (because I'm replicating the existing functionality and not necessarily fixing the problems):

* If you ban an offline player, you ban *anyone with their character name*. On servers where you can login as a different player, you *don't actually ban that player's character name, just their username*. This is the exact same way the system behaves today. A better solution is a database field for user names and character names and let you ban on both.
* If a player joins the server with an IP address name and that IP address gets banned, they will get banned too. Conversely, if a player with an IP address name joins and gets banned but they just registered, they won't have any known IPs so the ban will "fail" and also "succeed" at the same time. The ban system is great.
----
/ban add now replaces add, addtemp, and addip.
New syntax: /ban add <target> [time] [reason]

Examples:
- /ban add Shank 10d Rewrote the ban system.
- /ban add Ash
- /ban add 127.0.0.1 5d Go work on homework.

Note that if you want to specify a reason and a permaban, you need
to use 0 (zero) as the duration.

Examples:
- /ban add Ash 0 Love ya.
- /ban add Shank 0 Hacking.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1367,7 +1367,7 @@ namespace TShockAPI
 							}
 
 							targetGeneralizedName = target.Name;
-							success = TShock.Bans.AddBan(target.IP, target.Name, target.UUID, banReason, false, args.Player.User.Name,
+							success = TShock.Bans.AddBan2(target.IP, target.Name, target.UUID, target.User.Name, banReason, false, args.Player.User.Name,
 								banLengthInSeconds == 0 ? "" : banLengthInSeconds.ToString());
 
 							// Since this is an online ban, we need to dc the player and tell them now.
@@ -1397,7 +1397,7 @@ namespace TShockAPI
 							Regex r = new Regex(pattern, RegexOptions.IgnoreCase);
 							if (r.IsMatch(args.Parameters[1])) {
 								targetGeneralizedName = "IP: " + args.Parameters[1];
-								success = TShock.Bans.AddBan(args.Parameters[1], "", "", banReason,
+								success = TShock.Bans.AddBan2(args.Parameters[1], "", "", "", banReason,
 									false, args.Player.User.Name, banLengthInSeconds == 0 ? "" : banLengthInSeconds.ToString());
 								if (success && offlineUser != null)
 								{
@@ -1445,8 +1445,8 @@ namespace TShockAPI
 							string lastIP = JsonConvert.DeserializeObject<List<string>>(offlineUser.KnownIps).Last();
 
 							success = 
-								TShock.Bans.AddBan(lastIP,
-									offlineUser.Name, offlineUser.UUID, banReason, false, args.Player.User.Name,
+								TShock.Bans.AddBan2(lastIP,
+									"", offlineUser.UUID, offlineUser.Name, banReason, false, args.Player.User.Name,
 									banLengthInSeconds == 0 ? "" : banLengthInSeconds.ToString());
 						}
 
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1305,15 +1305,13 @@ namespace TShockAPI
 						List<TSPlayer> players = TShock.Utils.FindPlayer(args.Parameters[1]);
 						User offlineUser = TShock.Users.GetUserByName(args.Parameters[1]);
 
-						// Determines if the caller is effective root for overriding permissions
-						bool callerIsEffectiveRoot = false;
-
-						// A ban source is effective root if they are not not real and not rest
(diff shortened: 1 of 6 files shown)


mistzzt:
* A default ban reason may be considered as configurable.
* Can IP ban with no player related moved to a new command like _ipban_?
It would be a good idea to implement IP bans on IP range or CIDR.

hakusaro:
I would rather move those two to #1511 given that this system is on its way out anyway. But that's just me.

hakusaro:
(The point here is more to: Rewrite the add command so that it for sure doesn't match the issues in #1510. I just decided to try to roll account bans into this package because we already have bugs in the current system with them.)

TShockAPI/Commands.cs
```diff
@@ -1285,200 +1286,199 @@ private static void Ban(CommandArgs args)
...
-							else
-								args.Player.SendErrorMessage("Invalid player or account!");
-						}
-						else if (players.Count > 1)
-							TShock.Utils.SendMultipleMatchError(args.Player, players.Select(p => p.Name));
-						else
+						// A ban source is effective root if they are not not real and not rest
+						// Super admins ingame who attempt to run this will have to remove
+						// the immune to ban permission if they want to ban an immune player
+						if (!(args.Player is TSRestPlayer) && !args.Player.RealPlayer)
```
deadsurgeon42: Isn't it clearer to simply say `if (args.Player is TSServerPlayer)` as `TSServerPlayer` is exclusive to console?

hakusaro:
I'd like to add that while this is somewhat started from #1510, I would like to point out that we do not find similarities enough between Commaster's code and our code to warrant a speedier response. In particular, after analysis, we determined that it was unlikely that the code was copied.

TShockAPI/Commands.cs
```diff
@@ -1285,200 +1286,197 @@ private static void Ban(CommandArgs args)
...
+						// Case: Players & user are invalid, could be IP?
+						// Note: Order matters. If this method is above the online player check,
+						// This enables you to ban an IP even if the player exists in the database as a player.
+						// You'll get two bans for the price of one, in theory, because both IP and user named IP will be banned.
+						// ??? edge cases are weird, but this is going to happen
+						// The only way around this is to either segregate off the IP code or do something else.
+						if (players.Count == 0)
+						{
+							// If the target is a valid IP...
+							string pattern = @"^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$";
```
Ijwu: Why bother checking if this is a valid IP?

TShockAPI/DB/BanManager.cs
```diff
@@ -46,7 +46,8 @@ public BanManager(IDbConnection db)
 									new SqlColumn("Reason", MySqlDbType.Text),
 									new SqlColumn("BanningUser", MySqlDbType.Text),
 									new SqlColumn("Date", MySqlDbType.Text),
-									new SqlColumn("Expiration", MySqlDbType.Text)
+									new SqlColumn("Expiration", MySqlDbType.Text),
+									new SqlColumn("AccountName", MySqlDbType.Text)
```
Ijwu: Have we considered a migration plan for existing databases? Bundling with a program to upgrade SQLite or a script for MySQL may be preferable. I don't expect TShock to do it, but I expect us to address it.

TShockAPI/Commands.cs
hakusaro: Because we shouldn't IP ban any random input and call it a day.

TShockAPI/DB/BanManager.cs
hakusaro: System automatically adds new database columns to existing databases with no outside action required.

TShockAPI/Commands.cs
Ijwu: That's not what I meant. Why are you checking numerical ranges in IPs and not `[0-255]\.[0-255]\.[0-255]\.[0-255]`?

TShockAPI/DB/BanManager.cs
Ijwu: 👍 I take it that's a part of the pseudo-ORM we have?

hakusaro: @Ijwu yeah the ```EnsureTableStructure``` method actually will append columns to the database schema if not already present. It only works if you're appending on the end of the schema, though, and it doesn't work in reverse iirc.

PR #1515, opened 2017-12-02
Discourage superadmin use

hakusaro:
This PR attempts to discourage superadmin use. It removes the configuration options for configuring the group's prefix, suffix, and color. To further discourage use, I changed the color to white and made the prefix longer.

This also implements a ```/su``` command which substitutes your current user group with the super admin group for 10 minutes. If you use it again after elevating, you can downgrade back instantly. This lets you temporarily get a permission set that can do more, while simultaneously keeping you from permanently using it.

Since old databases won't have these changes, this is a "going forward" thing. A new database will ideally not have superadmin users, but someone who knows the system might do it anyway. We can't really stop it, but we did discourage new people from going though that process in the first place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added a warning notifying users of the minimum memory required to run TShock (@bartico6)
 * Added /group rename to allow changing group names (@ColinBohn, @ProfessorXZ)
 * Added /region rename and OnRegionRenamed hook (@koneko-nyan, @deadsurgeon42)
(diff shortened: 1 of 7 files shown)


bartico6:
You might want to block setting the group to superadmin using /user group.

hakusaro:
I considered it, but I think that at this point doing that would do more harm than good. The goal is to discourage but not flat out deny a user doing it. If someone is experienced enough with TShock to know what they want to do with /user add, I don't see any reason to block them. Same reason you can login as root on Linux -- though you're highly discouraged from doing so.

bartico6:
I'd argue that setting your group to superadmin is comparable to setting your UID to 0 on linux but in reality - you're correct :-)

hakusaro:
But you can still do that too. It's a really fun trick to play on students that are trying to secure Linux machines. Change some random user ID to 0 and then wait for them to discover that _games_ is root.

hakusaro:
> Superadmin should not exist, you should be able to /su (since /sudo will both a) conflict with Essentials+ and b) sudo in modern versions allows running as any user) and create their main group and later, the user they want.

No, I don't care about:

* Conflicting with essentials (not my problem).
* Supporting substituting other users (I'm not reimplementing unix everything here, just the good parts).

QuiCM:
LGTM

PR #1516, opened 2017-12-03
Modify table creation in MySQL land to support utf8mb4

### LINKED ISSUE #1463: Special character/emoji string message crashes SqlLog logger
**TShock Version**
4.3.24, APIv2.1, OTAPI v2.0.0.27

**Stack Trace**
```
2017-06-27 11:27:36 - SqlLog: ERROR: SQL Log insert query failed: MySql.Data.MySqlClient.MySqlException (0x80004005): Incorrect string value: '\xF0\x9F\x91\x8F.' for column 'Message' at row 1
   at MySql.Data.MySqlClient.MySqlStream.ReadPacket()
   at MySql.Data.MySqlClient.NativeDriver.GetResult(Int32& affectedRow, Int64& insertedId)
   at MySql.Data.MySqlClient.Driver.GetResult(Int32 statementId, Int32& affectedRows, Int64& insertedId)
   at MySql.Data.MySqlClient.Driver.NextResult(Int32 statementId, Boolean force)
   at MySql.Data.MySqlClient.MySqlDataReader.NextResult()
   at MySql.Data.MySqlClient.MySqlCommand.ExecuteReader(CommandBehavior behavior)
   at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
   at TShockAPI.DB.DbExt.Query(IDbConnection olddb, String query, Object[] args)
   at TShockAPI.SqlLog.Write(String message, TraceLevel level)
2017-06-27 11:27:55 - SqlLog: ERROR: SQL Logging disabled due to errors. Reverting to text logging.
2017-06-27 11:27:55 - SqlLog: ERROR: SQL log failed at: 2017-06-27 18:27:36. Message: Utils: INFO: zeff executed: /me [c/ffoooo:jumps, HOOraY!] good for u! 👏.
```

**Reproducibility**
- Running the following in game causes the error:
`/me [c/ffoooo:jumps, HOOraY!] good for u! 👏`
- Logging falls back to textfile afterwards - must reboot server in order to restore database logging again.

**Screenshot**
N/A

hakusaro:
This should in theory fix #1463. Need testing + feedback from MySQL users who want to try this out.

### DIFF
--- a/TShockAPI/DB/DBTools.cs
+++ b/TShockAPI/DB/DBTools.cs
@@ -75,6 +75,9 @@ namespace TShockAPI.DB
                     sb.Append(")");
             }
 
+            if (TShock.Config.StorageType.ToLower() == "mysql")
+                sb.append("CHARACTER SET utf8mb4 COLLATE utf8mb4_bin")
+
             using (var com = database.CreateCommand())
             {
                 com.CommandText = sb.ToString();


hakusaro:
cc @bartico6

bartico6:
This is the same query I use with my server's DB (and that I forced tShock to use) and so far no problems, so LGTM.

hakusaro:
@bartico6 as in, you're saying you've made this exact change and it works?

Or as in you've manually made tables in the past and that's how you've done it?

bartico6:
xD
As in I've made a db, then dumped the statements, then replaced the encodings, then dropped the db and created tables from the new sql scripts, and I'm successfully running it with tShock after importing stuff from the SQLite database we used to run.

hakusaro:
That's like, far from what this does. I was under the impression you could append the character set & collation stuff to the end of a table and it would fly correctly. If not, it has to be in the db creation stage, as well as the column stage, which this code doesn't do either of.

bartico6:
Oh I just realised, you're right...  
I'm afraid the character set and collation have to be set at the database level.  
You can always run `ALTER DATABASE [db] CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;` if you detect it isn't the correct encoding.

hakusaro:
Not doing this.

QuiCM:
We should update our table creation method though, so that future versions are compatible with emojis.
If @bartico6 could set up a small walkthrough we can provide upgrade steps for existing DBs

PR #1520, opened 2017-12-03
Implement revised default permission sets

### LINKED ISSUE #1519: Revise default permission set
This is the issue to track discussion from #1505 post merge of /su and /sudo. I hope this discussion takes less than 24 hours (and I hope to have a proposed new permission set in the next 12 hours). I'm really getting sick of everything in this project taking 7.8 decades to get merged.

Step 1: Generate a matrix of permissions so we can figure out what goes where.

hakusaro:
This fixes #1519. It introduces a revised set of default permissions. Most groups have had significant changes. This accounts for a lot of new or revised permissions that weren't ever updated in the default groups. Following this change, the default owner group is effectively the old superadmin, though with a few caveats and exceptions that I think are sane.

You can view the start and the end permission sets [on google docs](https://docs.google.com/spreadsheets/d/1R5_pTMeY2Z73-DPSdliuEDM38c6VHpA6qQLXY8D9jPM/edit?usp=sharing). The first tab in the spreadsheet is where I started with, along with what the suggestions were. The second sheet is the result of dumping the permission matrix from the newly implemented defaults.

Finally, this change adds a new method to Utils, ```DumpPermissionMatrix()```. This creates a markdown table that you can then use rich text copy paste to put into a spreadsheet system of your choice for later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added /region rename and OnRegionRenamed hook (@koneko-nyan, @deadsurgeon42)
 * Added /su, which temporarily elevates players with the tshock.su permission to super admin. In addition added, a new group, owner, that is suggested for new users to setup TShock with as opposed to superadmin. Finally, /su is implemented such that a 10 minute timeout will occur preventing people from just camping with it on. (@hakusaro)
 * Added /sudo, which runs a command as the superadmin group. If a user fails to execute a command but can sudo, they'll be told that they can override the permission check with sudo. Much better than just telling them to run /su and then re-run the command. (@hakusaro)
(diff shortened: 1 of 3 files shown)


hakusaro:
cc @Brycey92

Brycey92:
Nice markdown dumping code!

hakusaro:
There's probably a better way to do it, but I wanted something quick that worked. The horror is that I can't get dump to work.

hakusaro:
@Pryaxis/tshock requesting code review/merge from basically anybody.

mistzzt:
I'm considering if having a type like `DefaultSettings` containing default permissions would be good.
Next time if we want to change default settings like group permissions, we can directly modify this file.

hakusaro:
@mistzzt I like that idea, but I don’t know if that’s a good thing to implement if we intend to rewrite the database stuff in TS5 or not.

PR #1522, opened 2017-12-03
Allow /savessc to work on players that bypass ssc

### LINKED ISSUE #1506: Bypass SSC shouldn't prevent someone from forcing an SSC save
As discussed in #1503, right now the method that saves SSC data will return true (indicating that the data save was successful) if a save request happens and the player is a super admin. It also denies this request.

This method should be changed to:

- Not return true when the super admin tries to be saved.
- Permit the /savessc command to actually run the save operation as opposed to just ignoring it.

Potential solution:
- Overload the save method to allow forcing a save even if the user is bypassing ssc and permit the save command to work without auto save.

hakusaro:
Fixes #1506 by permitting the command to run the insertion of SSC data if run from a command. Bonus fix: ```InsertPlayerData``` also returns false if normal save operations would not save someone for this reason.


hakusaro:
@Pryaxis/tshock requesting code review/merge from basically anybody.

PR #1527, opened 2017-12-06
WIP: Fragments

QuiCM:
*THIS PR IS PRIMARILY FOR DISCUSSION PURPOSES & REVIEWS DURING THE DEVELOPMENT OF THE FRAGMENTS BRANCH. **DO NOT** MERGE WITHOUT MY APPROVAL*

To contribute to this PR, push commits to the Fragments branch (**with approval from @hakusaro**).


TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+using System.IO.Streams;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Terraria.ID;
+using TShockAPI.Net;
+
+namespace TShockAPI
+{
+	/// <summary>Bouncer - The TShock anti-hack and build guardian system</summary>
```
QuiCM: Comments do not need to (should not, imo) include the name of the object being commented.
A class level summary should provide high level details on what the class does (but more detailed than "anti-hack and build guardian system").

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+	public class Bouncer
+	{
+		/// <summary>Bouncer - Constructor call initializes Bouncer & related functionality.</summary>
+		/// <returns>A new Bouncer.</returns>
```
QuiCM: Returns on a constructor is a bit redundant. If this is to remain, add more details about what setup the returned instance has

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+		public Bouncer(TerrariaPlugin pluginInstance)
```
QuiCM: Do we need a public constructor? Why does it take a `TerrariaPlugin` reference?

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Terraria.ID;
+using TShockAPI.Net;
+
+namespace TShockAPI
+{
+	/// <summary>Bouncer - The TShock anti-hack and build guardian system</summary>
+	public class Bouncer
```
QuiCM: Is Bouncer intended to be extensible? If not this should be `sealed`.
If it is intended to be extensible we should provide an abstract base class or an interface from which this implementation inherits and future implementations can use

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+		/// <returns>A new Bouncer.</returns>
+		public Bouncer(TerrariaPlugin pluginInstance)
+		{
+			// Setup hooks
+
+			GetDataHandlers.SendTileSquare.Register(OnSendTileSquare);
+			GetDataHandlers.HealOtherPlayer.Register(OnHealOtherPlayer);
+		}
+
+		/// <summary>OnHealOtherPlayer - The handler for the HealOther events in Bouncer</summary>
```
QuiCM: Again, summaries don't need the method name either. Description could be more descriptive, especially for the parameters (`<param name />`s below)

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+		/// <param name="sender">sender</param>
+		/// <param name="args">args</param>
+		internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEventArgs args)
```
QuiCM: If Bouncer is intended to be extensible then this should be `public virtual`, or, assuming a base interface provides this method and bouncer inherits, `public override`

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			short amount = args.Amount;
+			byte plr = args.TargetPlayerIndex;
+
+			if (amount <= 0 || Main.player[plr] == null || !Main.player[plr].active)
+			{
+				args.Handled = true;
+				return;
+			}
+
+			if (amount > TShock.Config.MaxDamage * 0.2)
```
QuiCM: Why are we checking max damage * 0.2 for healing others?

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			}
+
+			if (args.Player.HealOtherThreshold > TShock.Config.HealOtherThreshold)
+			{
+				args.Player.Disable("Reached HealOtherPlayer threshold.", DisableFlags.WriteToLogAndConsole);
+				args.Handled = true;
+				return;
+			}
+
+			if (TShock.CheckIgnores(args.Player) || (DateTime.UtcNow - args.Player.LastThreat).TotalMilliseconds < 5000)
```
QuiCM: A comment here would be good.
I assume this rejects the packet if the player is
* 'ignored' due to other hack detection
* has been reprimanded for some other potential hack attempt in the last 5 seconds

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+
+			args.Player.HealOtherThreshold++;
+			args.Handled = false;
+			return;
+		}
+
+		/// <summary>OnSendTileSquare - The handler for SendTileSquare events in Bouncer</summary>
+		/// <param name="sender">sender</param>
+		/// <param name="args">args</param>
+		internal void OnSendTileSquare(object sender, GetDataHandlers.SendTileSquareEventArgs args)
```
QuiCM: Comments and design again, see above review comments. I'll stop marking these from here-on because the feedback is the same

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			int tileX = args.TileX;
+			int tileY = args.TileY;
+
+			if (args.Player.HasPermission(Permissions.allowclientsideworldedit))
+			{
+				args.Handled = false;
+				return;
+			}
+
+			if (size > 5)
```
QuiCM: We discussed this - 5 is a magic number, we should probably figure out exactly why we use 5.
IIRC it's because 5 means a 5x5 square which is normal for a tile square, and anything bigger is a non-vanilla tile modification attempt

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			{
+				args.Handled = true;
+				return;
+			}
+
+			if ((DateTime.UtcNow - args.Player.LastThreat).TotalMilliseconds < 5000)
```
QuiCM: Honestly we should just convert this into a method. Something like `bool IsWarned() => (DateTime.UtcNow - LastThread).TotalMilliseconds < 5000;` in `TSPlayer`

Ijwu:
Will do an official review before merge to ensure I get final code.

So far, it looks good. I appreciate the attempt to compartmentalize various functionality. In the long run this will be a great improvement. 👍

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			{
+				args.Player.SendTileSquare(tileX, tileY, size);
```
QuiCM: A comment on what SendTileSquare does would be good.
Something along the lines of "Undoes the action for the client that performed it. E.G., if we reject a tile break this sends the server's state of the tile so the client doesn't see a broken tile"

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,319 @@
...
+			}
+
+			if (TShock.CheckIgnores(args.Player))
+			{
+				args.Player.SendTileSquare(tileX, tileY, size);
+				args.Handled = true;
+				return;
+			}
+
+			try
```
QuiCM: This is a big logic block that's pretty arcane to look at.
Maybe add a short summary above the `try` statement that explains what happens in the block

TShockAPI/GetDataHandlers.cs
```diff
@@ -532,19 +532,64 @@ public static bool TSCheckNoclip(Vector2 Position, int Width, int Height)
 			return false;
 		}
 
+		/// <summary>HandledEventArgs - The event args object for the HealOtherPlayer event</summary>
```
QuiCM: Comments again

TShockAPI/TShock.cs
```diff
@@ -378,6 +378,8 @@ public override void Initialize()
 				Log.ConsoleInfo("Welcome to TShock for Terraria!");
 				Log.ConsoleInfo("TShock comes with no warranty & is free software.");
 				Log.ConsoleInfo("You can modify & distribute it under the terms of the GNU GPLv3.");
+
+				Bouncer b = new Bouncer(this);
```
QuiCM: This is a local variable. Will go out of scope and get GCed, won't it?

If dotnet magic keeps it alive, it's inaccessible externally and from the rest of the codebase

TShockAPI/Utils.cs
```diff
@@ -1427,5 +1427,66 @@ private void OutputElementsForDump(StringBuilder buffer, List<object[]> elements
 				}
 			}
 		}
+
+		/// <summary>FixChestStacks - Verifies that each stack in each chest is valid and not over the max stack count.</summary>
```
QuiCM: Comment style

TShockAPI/Bouncer.cs
hakusaro: Not.

TShockAPI/Bouncer.cs
hakusaro: Well, it might need one eventually (not done yet).

TShockAPI/Bouncer.cs
hakusaro: I don't want to hear the word "interface" without good justification ever again.

TShockAPI/Bouncer.cs
hakusaro: > Comments do not need to (should not, imo) include the name of the object being commented.

Got it.

> A class level summary should provide high level details on what the class does (but more detailed than "anti-hack and build guardian system").

Given that I don't know how much of our code is anti-hack, it would be quite stupid to predict the future and say what it does without knowing what it does.

TShockAPI/Bouncer.cs
hakusaro: Great question.

cc @Pryaxis/tshock does anyone know why?

TShockAPI/Bouncer.cs
hakusaro: Thank god.

TShockAPI/Bouncer.cs
hakusaro: Noted for future use (I have objectives that are more important than this though).

TShockAPI/TShock.cs
hakusaro: screen shot 2017-12-06 at 16 49 23

hakusaro: I wanted code compiling not being correct for final implementation.

Ijwu: Not for as long as it's hooks are registered, I believe.

The `Bouncer` constructor does that.

hakusaro:
The fragments branch is a pet project by me. If you want to contribute, send me a message on Discord and swear under oath that you will not abandon me ever.

TShockAPI/TShock.cs
hakusaro: @Ijwu it's still not remotely close to done because there's no destructor and no other stuff (this is not final obviously).

Ijwu: Understood. Just commenting on current state. Thanks! :)

TShockAPI/Bouncer.cs
QuiCM: In that case I assume this class is meant for internal use only, and should not be extensible by external parties (eg plugins).
`public sealed`, or even better, `internal sealed` markings should be applied

hakusaro:
I'm gonna update the topic for this issue eventually. Let me summarize what this branch is going to be in brief in the mean time.

This is going to be moving everything that TShock does out into subsystems based on logical groups. For example, the anti-hack code is currently the primary resident of GetDataHandlers, but it has utility functions spread all throughout TShock. These methods, like ```CheckIgnores``` just sit around and have very little descriptive purpose as to what they do. I'm going through and moving all of the code that calls these methods into different ```fragments``` of the original code.

At this point, I expect the first merge of this branch will happen once the anti-hack code is moved into its new home: ```Bouncer```. This project will likely take a non-insignificant amount of time, so I don't know exactly when I'll get there. Hopefully things will pick up as I start to isolate the anti-hack system.

The anti-hack system is decrepit and needs updating and revisiting for 2017. The problem is that updating it right now is difficult because reasoning about what it does is basically impossible. We need to have a hard look at it, but it's actually pretty low on my priority list. I want to get the ban system rewritten first, and that requires getting new database stuff done.

But all of that requires that we have logical groupings of subsystems. Essentially, I'm going the _gizmo_ route except not trying to re-implement all of TShock in one-go or trying to split all the subsystem stuff out in one shot.

The current basis for how I'm acting is by looking at things that call ```CheckIgnores``` and moving them into the Bouncer subsystem. Eventually, that's going to be extended to other methods in the main class and in the utils file. Any anti-hack/anti-cheat stuff should end up in ```Bouncer``` eventually.

The subsystem after ```Bouncer``` for isolation is SSC. Once that's done, I'm going to isolate smaller functional groups into a subsystem that will encompass TShock's minor tweaks. These are like perpetual events, random checks, and the initialize block. My goal is to get ```TShock.cs``` down to as few lines as possible of initialization code of some core subsystems. Those core subsystems will be ordered logically and easily changed by future code maintainers. It should be really stupid obvious just based on file structure what systems go where and how they can be modified.

That's my goal.

TShockAPI/Bouncer.cs
bartico6: Because heal other player only happens when you are using the spectre armor with the hood, and the healing you can do with that is 20% of your damage.

TShockAPI/Bouncer.cs
```diff
@@ -40,11 +41,138 @@ internal Bouncer(TerrariaPlugin pluginInstance)
...
+
+			if (!TShock.Utils.TilePlacementValid(x, y))
+			{
+				args.Handled = true;
+				return;
+			}
+
+			if (args.Player.Dead && TShock.Config.PreventDeadModification)
+			{
+				args.Handled = true;
```
bartico6: This should also send the tiles back to the user. If the user is dead and modifying tiles, this causes desync - the client will see a tunnel that nobody else sees, and appear to be noclipping. A workaround is to also send the tile back *before* 'handling' it. Just add `args.Player.SendTileSquare(x, y, 1);` 👍

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,803 @@
...
+					if (TShock.Itembans.ItemIsBanned(EnglishLanguage.GetItemNameById(selectedItem.netID), args.Player) || editData >= (action == EditAction.PlaceTile ? Main.maxTileSets : Main.maxWallTypes))
+					{
+						args.Player.SendTileSquare(tileX, tileY, 4);
+						args.Handled = true;
+						return;
+					}
+					if (action == EditAction.PlaceTile && (editData == 29 || editData == 97) && Main.ServerSideCharacter)
+					{
+						args.Player.SendErrorMessage("You cannot place this tile because server side characters are enabled.");
+						args.Player.SendTileSquare(tileX, tileY, 3);
```
mistzzt: If it is possible to use a constant to eliminate magic numbers here?

hakusaro: Sure. Do you know what those magic numbers are?

mistzzt: I think different numbers should represent the area of some tiles. 
1 for one tile, 4 for some large objects.
Here, the `editData` is the tile id for piggy bank and safes; Idk why the size of tile square is 3.

QuiCM: We can replace magic numbers with the related `TileID.TileType` constants

QuiCM: Also I imagine it's 3 because that's a 3x3 square (safe is a 2x2, piggy is a 2x1 irc) and we wanted to be safe

TShockAPI/Bouncer.cs
hakusaro: @QuiCM documented in eed7b3fa.

TShockAPI/Bouncer.cs
```diff
@@ -201,12 +201,14 @@ internal void OnPlaceObject(object sender, GetDataHandlers.PlaceObjectEventArgs
 
 			if (!TShock.Utils.TilePlacementValid(x, y))
 			{
+				args.Player.SendTileSquare(tileX, tileY, 4);
```
bartico6: Hate to be such a pain about this, but I'm almost [99%] certain that sending a 1-big box is enough - that's what I do on my server and it has allowed me to completely fix the issue of clientside tunneling by modifying tiles while dead. Since only one tile was modified, you should resend just that one tile rather than four - why waste bandwidth :-)

TShockAPI/Bouncer.cs
```diff
@@ -201,12 +201,14 @@ internal void OnPlaceObject(object sender, GetDataHandlers.PlaceObjectEventArgs
...
 				args.Handled = true;
 				return;
 			}
 
 			if (args.Player.Dead && TShock.Config.PreventDeadModification)
 			{
+				args.Player.SendTileSquare(tileX, tileY, 4);
```
bartico6: Same as above

TShockAPI/Bouncer.cs
```diff
@@ -201,14 +201,14 @@ internal void OnPlaceObject(object sender, GetDataHandlers.PlaceObjectEventArgs
 
 			if (!TShock.Utils.TilePlacementValid(x, y))
 			{
-				args.Player.SendTileSquare(tileX, tileY, 4);
+				args.Player.SendTileSquare(x, y, 4);
```
bartico6: Re-posting since it got insta-outdated.  
```Hate to be such a pain about this, but I'm almost [99%] certain that sending a 1-big box is enough - that's what I do on my server and it has allowed me to completely fix the issue of clientside tunneling by modifying tiles while dead. Since only one tile was modified, you should resend just that one tile rather than four - why waste bandwidth :-)```

File: TShockAPI/Bouncer.cs [hunk 22]
```diff
@@ -201,14 +201,14 @@ internal void OnPlaceObject(object sender, GetDataHandlers.PlaceObjectEventArgs
...
 				args.Handled = true;
 				return;
 			}
 
 			if (args.Player.Dead && TShock.Config.PreventDeadModification)
 			{
-				args.Player.SendTileSquare(tileX, tileY, 4);
+				args.Player.SendTileSquare(x, y, 4);
```
bartico6: Same here, as previously.

TShockAPI/GetDataHandlers.cs
```diff
@@ -2034,6 +2037,9 @@ private static bool HandlePlayerUpdate(GetDataHandlerArgs args)
 		{
 			if (args.Player == null || args.TPlayer == null || args.Data == null)
 			{
+				// Is this really the best option?
+				// If we're getting a packet that doesn't have a player or a TPlayer or data...
+				// Should we really let it through?
 				return false;
```
hakusaro: Looking for input on if this is the best return value here.

TShockAPI/GetDataHandlers.cs
```diff
@@ -2154,6 +2087,7 @@ private static bool HandlePlayerUpdate(GetDataHandlerArgs args)
 				}
 			}
 
+			// Where we rebuild sync data for Terraria?
```
hakusaro: I'm also wondering why this block of code does what it does. Does Terraria not handle this?

TShockAPI/GetDataHandlers.cs
```diff
@@ -2134,6 +2066,7 @@ private static bool HandlePlayerUpdate(GetDataHandlerArgs args)
 					args.Player.SendErrorMessage("You cannot use {0} on this server. Your actions are being ignored.", itemName);
 				}
 
+				// Reimplementation of normal Terraria stuff?
```
hakusaro: Why is this here too?

TShockAPI/GetDataHandlers.cs
bartico6: Packets that come from null players might be packets sent prematurely by unconnected clients, this should be filtered to avoid exceptions further down the execution chain

hakusaro: @bartico6 yeah, but by returning _false_ we're letting the server process it. If we return _true_ the event is cancelled and doesn't get sent to the server.

TShockAPI/GetDataHandlers.cs
bartico6: I think this is to avoid a client just sending a "health update" with arbitrary values to set their maxhealth, so instead you'd manually detect events which increase max hp and replay them against the server character data

TShockAPI/GetDataHandlers.cs
hakusaro: I _think_ this should be _true_

bartico6: If true means handled then yes, it should be true IMO.

hakusaro: @bartico6 I'm waiting on @QuiCM to confirm before changing this. He's the one who implemented the check, and I don't want to change behavior without 100% assurance of what the intention was.

TShockAPI/Bouncer.cs
```diff
@@ -51,6 +52,33 @@ internal Bouncer(TerrariaPlugin pluginInstance)
...
 
+		/// <summary>Handles ProjectileKill events for throttling & out of bounds projectiles.</summary>
+		/// <param name="sender">The object that triggered the event.</param>
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnProjectileKill(object sender, GetDataHandlers.ProjectileKillEventArgs args)
+		{
+			if (args.ProjectileIndex > Main.maxProjectiles || args.ProjectileIndex < 0)
+			{
+				// TODO: Should this be /true/ to stop the server from processing it?
+				args.Handled = false;
```
hakusaro: Should this be true to stop the server from processing it?

TShockAPI/GetDataHandlers.cs
QuiCM: I don't recall why I did false. I can't think of any reason why we shouldn't prevent propagation of a bad packet.

I'm 100% fine with this changing to true

TShockAPI/Bouncer.cs
QuiCM: I wonder if Terraria has any special logic for projectile id = -1.
However this looks like inverted logic again (where someone assumed false meant it was handled)

hakusaro: I whipped out my handy dandy blame tool. Originally implemented by Zido as a `return true`, changed to `return false` by Deathmax in a commit called "merge fail."

screen shot 2017-12-11 at 00 21 19

screen shot 2017-12-11 at 00 21 44

Mind if I change it to true, @QuiCM?

hakusaro: I checked the [merge fail](https://github.com/Pryaxis/TShock/commit/672d360e9dbe6c93997f4d17ac0f5e2c987db6ea) commit and I definitely don't know what to say about it.

hakusaro:
Progress update: 15/22 `CheckIgnores` calls to go (31% complete). I currently expect this to be completed by _Friday (Saturday Australia)_.

hakusaro:
There should be at least 424 more lines that can be chopped out of TShock's main class by the end of fragments (v1). Potentially a ton more if I move all of the stupid `OnSecondUpdate` stuff to Bouncer (if applicable).

hakusaro:
50%, 11 more calls to go.

hakusaro:
Currently, it looks like we need to have a subsystem for regions and a subsystem for item bans, broken out separately.

PR #1528, opened 2017-12-07
Revert "Remove superadmin customization config options"

hakusaro:
This reverts commit 1e68ac22c7d7c191f05e6f86b7847c2b4f885340.

Essentially, I prematurely did this change to discourage superadmin use. But because the direction we went in was basically to obsolete it (and more notably, provided no upgrade path for existing databases), we're just breaking people's existing setups for no reason.

I propose we remove these changes if and only if we write something that migrates people to the new permission set from the old one. If we do, we can keep these. If not, though, existing databases really have no refuge to go to post-removal of these changes.

### DIFF
--- a/TShockAPI/ConfigFile.cs
+++ b/TShockAPI/ConfigFile.cs
@@ -80,11 +80,11 @@ namespace TShockAPI
 
 		/// <summary>SuperAdminChatRGB - The chat color for the superadmin group.</summary>
 		[Description("#.#.# = Red/Blue/Green - RGB Colors for the Admin Chat Color. Max value: 255.")]
-		public int[] SuperAdminChatRGB = { 255, 0, 0 };
+		public int[] SuperAdminChatRGB = { 255, 255, 255 };
 
 		/// <summary>SuperAdminChatPrefix - The superadmin chat prefix.</summary>
 		[Description("Super admin group chat prefix.")]
-		public string SuperAdminChatPrefix = "(Admin) ";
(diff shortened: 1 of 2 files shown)


QuiCM:
I don't understand what that commit has to do with the TShock DB?

hakusaro:
@QuiCM the default permission sets were changed but that only applies if you have a new database. So we'd be screwing over people who don't have a fresh database by giving them no customization on a group they might use daily (and continue to use daily).

(We don't modify the default groups beyond their initial creation).

I'm really trying to save us 300 support queries of "why my superadmin now white?!"

QuiCM:
If we're going this route, I'd suggest adding a message to `PostLogin` if the user is a superadmin stating something like
"You're currently logged in to a superadmin group account. We strongly discourage the use of this group. Consider using a different group and making use of the /sudo command to if you require elevated permissions"

QuiCM:
(Why? Because adjust the config defaults won't change anything on existing servers that have already created a config file, so they won't even notice we're discouraging use of the group)

hakusaro:
Well yeah, I know that about the config file. I don't want people to try to use it afterwards (but also keep them in the loop).

I kinda don't want to too actively desuade people on existing databases. I don't even want to think about how problematic it would be to try to help people try to migrate or update without us doing the backbone first.

hakusaro:
To be clear, I would rather get this merged and discourage more in the future with good ways rather than trying to create more of a support burden for an already burdened staff.

PR #1532, opened 2017-12-09
WIP Updates the commandline parser to not be broken.

### LINKED ISSUE #1521: mono TerrariaServer.exe -dump doesn't dump things
On some latest version of general-devel, ```mono TerrariaServer.exe -dump``` just runs the server, doesn't dump anything, and doesn't exit. No errors in console, nor in any log files. Also, Utils.Dump() is never called because it should error out and die because not all of the dump functions were updated for the newest version of Terraria.

screen shot 2017-12-03 at 16 00 12

QuiCM:
It should now correctly parse a commandline such as `["", "-flag", "-flag", "arg" ... "etc" ]`

If someone could do some testing with an extensive commandline, that would be appreciated.

Fixes #1521.


hakusaro:
Lgtm pending changelog update. Tested with -dump and it worked fine. Makes me wonder if this was related to that one issue with command lines not working.

hakusaro:
@QuiCM per the one man rule, just merge this if no testing happens if it starts to bug you.

ivanbiljan:
Tested and proven to work fine (excluding the crash that occurs with DumpPermissionMatrix - #1533)

PR #1530, opened 2017-12-08
Rename TShock.DB.User to TShock.DB.UserAccount

Ryozuki:
Don't merge.

I wanted to start to work at https://github.com/Pryaxis/TShock/issues/1513
So if anyone wants to contribute you can push to this branch.

What is done actually:
- Rename User to UserAccount to prevent confussion (https://github.com/Pryaxis/TShock/issues/1513#issuecomment-348672961)

Issue #1513, opened 2017-12-02
Rewriting Users


QuiCM:
Users have problems that need fixing, see the attached Project for more details.

This issue is for discussion on how we should go about fixing the mentioned issues

hakusaro:
I would add that we have a lot of bad language. [Player] and [User] really need to be disambiguated so that it's clear that hey, this is something [active on the server playing] and this is the [data store about their account with us].

QuiCM:
Perhaps a rename from User to Account

Ryozuki:
"Relational mappings & sane data typing" (https://github.com/Pryaxis/TShock/projects)
Does this mean moving all the methods that query/insert to each respective class? 
A.k.a getting rid of Database managers such as CharacterManager?

E.g
```csharp
// CharacterManager.cs
public PlayerData GetPlayerData(TSPlayer player, int acctid)
// to
class TSPlayer 
{
    // ...
    public void GetData(); // uses the already stored id to retrieve data from db
}
```

ivanbiljan:
@Ryozuki no. Relational mappings address the issue of the current database design (IPs are stored in a single column as a serialized array, we assume that the 'Group' column matches the Group table...). You should read up a bit on database normalization.
A more acceptable design should look something like this (in my humble opinion):

The 'Users' table: should contain basic information about an account. This includes:
   - **AccountID**: the unique account identifier
   - **Username**: the account's name (should probably be PKd)
   - **UUID**: the account's UUID 
   - **Register Date**: the account's registration date
   - **Last Accessed**: the date that the account was last accessed (logged in to)

The 'UserHasGroup' table: should link the two tables ('Users' and 'Groups') 
   - **User**: foreign key. Holds a reference to the corresponding account in the 'Users' table
   - **Group**: foreign key. Holds a reference to the corresponding group in the 'Groups' table

Similarly we should have a 'UserHasIP' table or such. This is a simple example I came up with to clarify the meaning of "relational mappings", you're free to use your own ideas.

QuiCM:
Blocking this on fragments too

PR #1534, opened 2017-12-09
Add some SQL improvements

Ryozuki:
- Added DateTime datatype support for both sqlite and mysql.
- Added a [DefaultCurrentTimestamp](https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html) field to SqlColumn
- Added a function to check for errors in sqlcolumns:
`public void SqlColumnErrorCheck(List<SqlColumn> columns)`
Which raises a `SqlColumnExcepcion`.
For now it just checks if `DefaultCurrentTimestamp` is set true on a non DateTime datatype column, if you know more possible errors add them there.
- Added some XML comments

_Note: Tested on MySQL and SQLite, it seems to work_


Ryozuki:
Maybe if this gets merged, we could change [UserManager.cs#L44](https://github.com/Pryaxis/TShock/blob/general-devel/TShockAPI/DB/UserManager.cs#L44) in the pr https://github.com/Pryaxis/TShock/pull/1530 / issue https://github.com/Pryaxis/TShock/issues/1513

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -222,4 +389,28 @@ protected static string BuildWhere(List<SqlValue> wheres)
...
 		}
 	}
+
+	/// <summary>
+	/// An excepcion generated by the Column check. 
+	/// </summary>
+	[Serializable]
+	public class SqlColumnExcepcion : Exception
+	{
+		public SqlColumnExcepcion()
```
hakusaro: Needs comments.

TShockAPI/Extensions/DbExt.cs
```diff
@@ -189,6 +192,10 @@ public static SqlType GetSqlType(this IDbConnection conn)
 		                                                                                     			typeof (double?),
 		                                                                                     			(s, i) => s.IsDBNull(i) ? null : (object)s.GetDouble(i)
 		                                                                                     			},
+																									{
+																										typeof (DateTime),
```
hakusaro: Way too many spaces here bro.

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -222,4 +389,28 @@ protected static string BuildWhere(List<SqlValue> wheres)
 			return "WHERE {0}".SFormat(string.Join(", ", wheres.Select(v => v.Name + " = " + v.Value)));
 		}
 	}
+
+	/// <summary>
+	/// An excepcion generated by the Column check. 
+	/// </summary>
+	[Serializable]
+	public class SqlColumnExcepcion : Exception
```
hakusaro: `Exception` should be changed to the English spelling `Exception` instead of the Spanish version for codebase consistency. Sorry, I know this is a shitty thing to ask, but yeah.

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -122,8 +220,15 @@ public override string RenameTable(string from, string to)
 			{ MySqlDbType.Double, "DOUBLE" },
 			{ MySqlDbType.Int32, "INT" },
             { MySqlDbType.Int64, "BIGINT"},
```
hakusaro: Should be changed to indent with tabs instead of spaces if you're modifying this already.

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -73,8 +141,15 @@ public override string RenameTable(string from, string to)
 			{ MySqlDbType.Int32, "INTEGER" },
 			{ MySqlDbType.Blob, "BLOB" },
             { MySqlDbType.Int64, "BIGINT"},
```
hakusaro: This line (again, not modified by you) should be reformatted with tabs.

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -175,11 +295,41 @@ public string AlterTable(SqlTable from, SqlTable to)
 			return "{0}; {1}; {2}; {3};".SFormat(alter, create, insert, drop);
 		}
 
+		/// <summary>
+		/// Check for errors in the columns.
+		/// </summary>
+		/// <param name="columns"></param>
+		/// <exception cref="SqlColumnExcepcion">description</exception>  
+		public void SqlColumnErrorCheck(List<SqlColumn> columns)
```
hakusaro: Not entirely sure that the void return type is great here. But...I don't know the conventions I admit. I think if voids are cool then this can be kept (I still feel like `ValidateSqlColumnType` would be better).

hakusaro:
@Ryozuki it's probably a stretch to say that this database format will stay around long. I want to reiterate that here -> my inclination is to sell us on a less SQL orientated approach, personally. Given that, though, I also support modifying the existing types that use DateTime informally (`BanManager` and `UserManager`) to this type just so that we're not sitting around pretending strings are strings all day. Cause that nearly got a bug in the ban system that broke temp bans (failed validation that a type was a `DateTime` and not a string).

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -16,45 +16,120 @@
...
+		/// <param name="from">Old name of the table</param>
+		/// <param name="to">New name of the table</param>
+		/// <returns>The sql query for renaming the table.</returns>
 		string RenameTable(string from, string to);
 	}
 
+	/// <summary>
+	/// Query Creator for Sqlite
+	/// </summary>
 	public class SqliteQueryCreator : GenericQueryCreator, IQueryBuilder
```
bartico6: In order to be consistent with the naming of the official libraries provided by SQLite authors, the naming should follow the same naming as said libraries: **SQLiteQueryCreator**. Whether you do it, or not, is a matter of choice between following C# naming conventions, or official library naming already in place. I'd personally follow the second for clarity and readability. @Ryozuki

TShockAPI/DB/IQueryBuilder.cs
```diff
@@ -83,28 +165,51 @@ public string DbTypeToString(MySqlDbType type, int? length)
...
 		protected override string EscapeTableName(string table)
 		{
 			return table.SFormat("'{0}'", table);
 		}
 	}
 
+	/// <summary>
+	/// Query Creator for MySQL
+	/// </summary>
 	public class MysqlQueryCreator : GenericQueryCreator, IQueryBuilder
```
bartico6: To follow the naming scheme in libraries provided by Oracle, the naming should follow the same scheme they use: **MySqlQueryCreator**.

hakusaro: imho it should be `MySQLQueryCreator`

TShockAPI/DB/IQueryBuilder.cs
QuiCM: Given that we already have SqliteBlah, to adjust it would only cause yet more breaking changes while adding no value

ivanbiljan: I agree with White on this one.

TShockAPI/DB/IQueryBuilder.cs
bartico6: Same, but their libraries explicitly say "MySqlSomething" so I guess it's just for consistency's sake?

hakusaro:
@Pryaxis/tshock someone needs to look at this, sign off, or do something.

PR #1481, opened 2017-07-24
Add more HintPaths for OTAPI

Leeleo3x:
This will help msbuild to find missing libraries when create release
with `/p:Configuration=Release`.

With this patch, the searching order will be:
  - If OTAPI with same configuration found, use that.
  - Else if OTAPI with Release configuration found, use that.
  - Else use OTAPI with Debug configuration.

This will fix the issue #1480

PR #1545, opened 2017-12-10
Tick the HintPath for OTAPI to 1.3.5.3

hakusaro:
In https://github.com/Pryaxis/TShock/pull/1481, hinting was fixed so that OTAPI could be built correctly no matter where it was. This, however, needs to be kept in sync version wise with Terraria versions or the patch won't work.

Looking for suggestions as to how to make sure we don't forget this in the future.


QuiCM:
Solved in #1547

PR #1555, opened 2017-12-11
Remove /restart & Permissions.updateplugins

hakusaro:
1. Restart is never used in production. Anyone who runs a server at least runs it in a loop and reboots with "exit" causing a relaunch. Bonus: it doesn't work on mono. Bonus: it's always been unreliable, even on Windows.
2. Permission is never going to get used.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -366,10 +366,6 @@ namespace TShockAPI
 			{
 				HelpText = "Reloads the server configuration file."
 			});
-			add(new Command(Permissions.maintenance, Restart, "restart")
-			{
-				HelpText = "Restarts the server."
-			});
 			add(new Command(Permissions.cfgpassword, ServerPassword, "serverpassword")
 			{
 				HelpText = "Changes the server password."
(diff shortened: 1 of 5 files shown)


TShockAPI/Commands.cs
```diff
@@ -366,10 +366,6 @@ public static void InitCommands()
 			{
 				HelpText = "Reloads the server configuration file."
 			});
-			add(new Command(Permissions.maintenance, Restart, "restart")
```
QuiCM: Should identify is Permissions.maintenance is used elsewhere

TShockAPI/Commands.cs
```diff
@@ -1902,25 +1898,6 @@ private static void Off(CommandArgs args)
...
-			}
-
-			if (ServerApi.RunningMono)
-			{
-				TShock.Log.ConsoleInfo("Sorry, this command has not yet been implemented in Mono.");
-			}
-			else
-			{
-				string reason = ((args.Parameters.Count > 0) ? "Server shutting down: " + String.Join(" ", args.Parameters) : "Server shutting down!");
-				TShock.Utils.RestartServer(true, reason);
```
QuiCM: Do we still need Utils.RestartServer?

TShockAPI/Commands.cs
hakusaro: ? It's used in the update checker (and the perm def says that).

QuiCM: Cool, all good then

TShockAPI/Commands.cs
hakusaro: Nope.

QuiCM:
LGTM

PR #1558, opened 2017-12-14
New readme, docs, infra for devs

hakusaro:
This is the WIP branch for my initiative to revamp:

- [x] The readme
- [x] The contributing guidelines
- [x] Developer setup

Please give feedback on my copy and what needs to be improved or changed. I'm trying to make it look top notch 👌🏻

(Fixes #1531.)

PR #1559, opened 2017-12-14
Switch the "auth system" to "initial setup system" everywhere.

hakusaro:
This is a followup from https://github.com/Pryaxis/TShock/pull/1558#discussion_r156857135. It changes the entire "auth system" to the initial setup system and replaces all instances of this verbiage in the system.

This should have no effect on existing servers. If a user exists the authentication system is disabled even if the lockfile is gone. I considered removing the `setup.lock` / `auth.lck` system entirely, but I decided against it. Should I?

This also fixes a bonus problem where the system could turn itself back on if the lock file was removed because the superadmin account wouldn't exist in an ideal scenario (the player created an owner account, not a superadmin account).


QuiCM:
LGTM

hakusaro:
Should hold this issue open until I get some in-progress docs updated for this fact.

hakusaro:
The [documentation for 5.0](https://tshock.readme.io/v5.0/docs) has been updated with the new setup procedure and the removal of the superadmin stuff.

PR #1560, opened 2017-12-16
Fragments Part 1: Implement Bouncer infrastructure

hakusaro:
This represents the completion of work on `fragments` part one, and the discussion from PR #1527.

This first pull request creates the new `Bouncer` subsystem in TShock, attempting to segregate out a large portion of the anti-cheat/anti-exploit/anti-hack code into a single unit of operation (Bouncer). This is going to start a relative trend of subsystem isolation and removal that will progress through TShock 5. No other pull requests should be this large (maybe SSC?), because the vast majority of TShock's `GetDataHandlers` file was just implementing various different anti-cheat mechanisms.

The overall summary of the changes can be found in the changelog, but I've reposted them here:

* `GetDataHandlers.SendTileSquare` hook now sends a `TSPlayer` and a `MemoryStream` of raw data. (@hakusaro)
* Added `GetDataHandlers.HealOtherPlayer` hook. (@hakusaro)
* Added `GetDataHandlers.PlaceObject` hook. (@hakusaro)
* `GetDataHandlers.KillMe` now sends a `TSPlayer` and a `PlayerDeathReason`. (@hakusaro)
* Added `GetDataHandlers.ProjectileKill` hook. (@hakusaro)
* Removed `TShock.CheckProjectilePermission` and replaced it with `TSPlayer.HasProjectilePermission` and `TSPlayer.LacksProjectilePermission` respectively. (@hakusaro)
* Added `TSPlayer` object to `GetDataHandlers.LiquidSetEventArgs`. (@hakusaro)
* Removed `TShock.StartInvasion` for public use (moved to Utils and marked internal). (@hakusaro)
* Fixed invasions started by TShock not reporting size correctly and probably not working at all. (@hakusaro)
* Removed `GetDataHandlers.TileKill` and replaced it with `GetDataHandlers.PlaceChest` as the packet originally designated as tile kill is now only used for chests. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.NPCHome`. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.ChestItemChanged`. (@hakusaro)
* Fixed chest item changes not triggering any range checks, tile checks, or correct chest checks. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.PlayerBuff`. (@hakusaro)
* Added `TSPlayer` and `PlayerDeathReason` to `GetDataHandlers.PlayerDamage`. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.NPCStrike`. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.PlayerAnimation`. (@hakusaro)
* Added `GetDataHandlers.MassWireOperation` hook and related arguments. (@hakusaro)
* Added `GetDataHandlers.PlaceTileEntity` hook and related arguments. (@hakusaro)
* Added `TSPlayer` to `GetDataHandlers.GemLockToggle`. (@hakusaro)
* Added `GetDataHandlers.PlaceItemFrame` hook and related arguments. (@hakusaro)
* Added `TSPlayer.IsBouncerThrottled()`. (@hakusaro)
* Added `TSPlayer.CheckIgnores()` and removed `TShock.CheckIgnores(TSPlayer)`. (@hakusaro)

There are a lot of things that got moved into Bouncer, and some of them are not all entirely appropriate. For example, calls to `TShock.CheckPermissions` do double duty or even triple duty in checking multiple things at once. Some of the item ban system was moved into Bouncer due to complexity issues, and other parts of Bouncer are not 100% polished yet. However, I intend to perform more work _after_ this has been merged in a bigger dig into other subsystem replacement. At some point, I wanted to make sure that this stuff got into `general-devel` before embarking on a huge perfection spree on one system. The biggest gains are to be had by fundamentally rewriting large hooks that TShock has laying around. These are things like `OnSecondUpdate()` which weren't wrapped well to begin with.

A few other misc changes got into this as well. In particular:

* Removal of some old handling code for packet types not used anymore (code duplication through copypasta).
* "Fixing" hooks that were improperly performing `return false` instead of `return true` in handlers. True means a packet has been handled and doesn't need to be dealt with again, whereas false indicates that it isn't our problem (and Terraria still executes the code). Some handlers improperly confused these two.

The type of feedback I'm looking for in particular:

* Use of unknown magic numbers.
* Inconsistencies in code execution pathways.
* Confusing documentation.
* Anything else you can think of within reason?

The types of feedback I want to avoid:

* "But I don't know what CheckIgnores() does!" -- I don't either and I'm not unwrapping that before this gets merged.
* "I don't know what this code block does!" -- chances are I probably don't know 100% either, but I can dig around and see what's up.
* "These are breaking changes!" -- we have already decided to break API compatibility in the next release, and we might as well do all breaking now rather than later.

I've already incorporated most of the feedback where appropriate from #1527. Anything not incorporated was something I didn't know the answer to. In addition, I plan on building out some issues for smaller more quantized Bouncer/fragments/subsystem split up endeavors in the future.

Cheers!


deadsurgeon42:
A gotcha with GetData handlers is that their handler lists might be null during initialization (possibly if plugin loads before tshock):
```cs
GetDataHandlers.NpcStrike += something;
                ^^^^^^^^^ possible System.NullReferenceException
```
and must be initialized first in case it's unset:
```cs
if (GetDataHandlers.NpcStrike == null)
  GetDataHandlers.NpcStrike = new HandlerList<GetDataHandlers.NpcStrikeEventArgs>();

GetDataHandlers.NpcStrike += something;
```

is just a pet peeve. (I would want those lists as events instead but I guess that makes logging which plugin throws exceptions difficult)

hakusaro:
I want specific, in-depth testing on the following:

* Chests being used to hack items into the game (see https://github.com/Pryaxis/TShock/pull/1560/commits/845fc3b7c80d7df870dd241681abfce2ea9a41c9). I'm 99% sure that these hooks were erroneously using `return false` (thus being not-handled and being accepted) which made them null and void.

hakusaro:
@deadsurgeon42 so the trick is to initialize them with the += operator. From `HandlerList.cs`:

```
public static HandlerList<T> operator +(HandlerList<T> hand, HandlerItem obj)
{
	if (hand == null)
		hand = new HandlerList<T>();

	hand.Register(obj);
	return hand;
}
```

TShockAPI/TSPlayer.cs
```diff
@@ -285,6 +285,21 @@ public Group Group
...
+		/// <returns>If the player is currently being throttled by Bouncer, or not.</returns>
+		public bool IsBouncerThrottled()
+		{
+			return (DateTime.UtcNow - LastThreat).TotalMilliseconds < 5000;
+		}
+
+		/// <summary>CheckIgnores - Checks a players ignores...?</summary>
+		/// <param name="player">player - The TSPlayer object.</param>
+		/// <returns>bool - True if any ignore is not none, false, or login state differs from the required state.</returns>
+		public bool CheckIgnores()
```
Ijwu: I'm not sure about the purpose this method serves.

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+			GetDataHandlers.HealOtherPlayer += OnHealOtherPlayer;
+			GetDataHandlers.TileEdit += OnTileEdit;
+		}
+
+		/// <summary>Fired when an item frame is placed for anti-cheat detection.</summary>
+		/// <param name="sender">The object that triggered the event.</param>
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnPlaceItemFrame(object sender, GetDataHandlers.PlaceItemFrameEventArgs args)
+		{
+			if (args.Player.CheckIgnores())
```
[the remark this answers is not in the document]
hakusaro: Well, I got that far, but the problem is that it checks so much and ignores are strings, so...? It's just gonna have to get rebuilt I guess.

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+			{
+				NetMessage.SendData((int)PacketTypes.UpdateTileEntity, -1, -1, NetworkText.Empty, args.ItemFrame.ID, 0, 1);
```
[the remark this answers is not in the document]
hakusaro: > Basically, does args.Handled = true; mean we're disabling further processing of the packet?

It means that we're saying we "processed it" and that the server shouldn't take further action.

> If so, what is the SendData call for?

Sends our view of the world back to the client to indicate that their action didn't happen.

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+		/// <param name="sender">The object that triggered the event.</param>
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnUpdateNPCHome(object sender, GetDataHandlers.NPCHomeChangeEventArgs args)
+		{
+			int id = args.ID;
+			short x = args.X;
+			short y = args.Y;
+			byte homeless = args.Homeless;
+
+			// Calls to TShock.CheckTilePermission need to be broken up into different subsystems
```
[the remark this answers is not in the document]
hakusaro: It's on my list.

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+				}
+					
+				if (args.Player.Dead && TShock.Config.PreventDeadModification)
+				{
+					args.Player.SendTileSquare(tileX, tileY, 4);
+					args.Handled = true;
+					return;
+				}
+
+				if (args.Player.AwaitingName)
```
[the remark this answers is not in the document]
hakusaro: Agreed.

TShockAPI/TSPlayer.cs
hakusaro: Just fwiw, the method CheckIgnores() is just moved here from the TShock main class as a static call. 100% of this is carry over. Hence #1561.

TShockAPI/Bouncer.cs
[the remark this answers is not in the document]
hakusaro: https://github.com/Pryaxis/TShock/issues/1567

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+				{
+					bool includeUnprotected = false;
+					bool includeZIndexes = false;
+					bool persistentMode = false;
+					foreach (string parameter in args.Player.AwaitingNameParameters)
```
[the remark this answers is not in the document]
hakusaro: ?

hakusaro:
Blocking on rerpoduction of Joshwoo's bug, otherwise code is 100% okay with me atm.

TShockAPI/TSPlayer.cs
```diff
@@ -789,6 +804,57 @@ public void Spawn(int tilex, int tiley)
 			}
 		}
 
+		/// <summary>Checks to see if this player object lacks access rights to a given projectile. Used by projectile bans.</summary>
+		/// <param name="index">The projectile index from Main.projectiles (NOT from a packet directly).</param>
+		/// <param name="type">The type of projectile, from Main.projectiles.</param>
+		/// <returns>If the player has lacks access rights to the projectile.</returns>
+		public bool LacksProjectilePermission(int index, int type)
```
QuiCM: This is a needless method as it's just a negation of an existing method

TShockAPI/Utils.cs
```diff
@@ -925,15 +925,15 @@ public bool TryParseTime(string str, out int seconds)
 		/// </summary>
 		/// <param name="identity">identity</param>
 		/// <param name="owner">owner</param>
-		/// <returns>projectile ID</returns>
+		/// <returns>projectile ID or -1 if not found</returns>
 		public int SearchProjectile(short identity, int owner)
```
QuiCM: Just as a note, in future this would be better as a `TrySearch..` or similar:
```C#
public bool TrySearchProjectile(short identity, int owner, out int projId) {... }
```

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1824 @@
...
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnChestItemChange(object sender, GetDataHandlers.ChestItemEventArgs args)
+		{
+			short id = args.ID;
+			byte slot = args.Slot;
+			short stacks = args.Stacks;
+			byte prefix = args.Prefix;
+			short type = args.Type;
+
+			if (args.Player.TPlayer.chest != id)
```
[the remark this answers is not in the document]
QuiCM: `args.Player.TPlayer.chest` should be set to the chest that the player has open

So if the packet is received (player has opened chest) and the player's chest id is not the chest referenced in the packet that they're supposedly opening, then something is wrong

TShockAPI/TSPlayer.cs
hakusaro: Nah it’s okay. It’s for quality of life.

QuiCM: This is the same kind of thing that resulted in `UserException`, `UserNotExistException`, and `UserExistsException`

Joshwoo70:
Video of Reporduction: [Here](https://youtu.be/DVD5OjaRZ1I)
Version: [2008 Build No.][PR: #1568]
Plugins:
plugins
Fresh install, including fresh set of configs

hakusaro:
@Ijwu well, this is somewhat ironic. It turns out that if you range check the prefix on the item drop event, and it's < 1, and you block it, you create item duplication by preventing the client from cleaning up dropped items. I've reverted the change that blocked that and added a comment so we know not to do that again.

This hook is called twice for two different packets (21 and 90) and the [packet docs](https://tshock.readme.io/docs/multiplayer-packet-structure) don't say anything about this behavior. Either that or I can't understand them to make heads or tails of a negative prefix.

Enerdy:
@hakusaro wait I'm confused, can't we just send an Item update packet to the client that spawned the item for cleanup while handling the packet so other clients don't receive it?

hakusaro:
@Enerdy gonna need a little more context here. If you're referring to the prefix thing, I mean, you could, but if it's vanilla functionality there's no reason why we should stop the game from getting it (or other plugins). You only handle an event that you want to prohibit -- why would we prohibit it?

Enerdy:
I missed the part where Vanilla already handles this specific case for us. In that case, then yeah, no reason to double up.

hakusaro:
cc @Pryaxis/tshock

TShockAPI/Bouncer.cs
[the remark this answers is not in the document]
Ijwu: I'm referring to the inclusion of string flags like `-u` in a variable for parameters. There isn't much clarity of type safety behind it.

TShockAPI/GetDataHandlers.cs
```diff
@@ -3318,317 +2538,115 @@ private static bool HandleChestActive(GetDataHandlerArgs args)
...
-				args.Player.SendData(PacketTypes.PlayerUpdate, "", id);
+			if (id < 0 || id > 400)
 				return true;
-			}
 
-			if (TShock.Players[id].GodMode)
+			if (id == 400 && owner == 255)
 			{
-				TShock.Players[id].Heal(args.TPlayer.statLifeMax);
+				args.Player.IgnoreSSCPackets = false;
```
Olink: What does this do?  Seems dangerous.

TShockAPI/Utils.cs
```diff
@@ -1456,5 +1456,93 @@ private void OutputElementsForDump(StringBuilder buffer, List<object[]> elements
...
+		/// <summary>Determines the distance between two vectors.</summary>
+		/// <param name="value1">The first vector location.</param>
+		/// <param name="value2">The second vector location.</param>
+		/// <returns>The distance between the two vectors.</returns>
+		public static float Distance(Vector2 value1, Vector2 value2)
+		{
+			float num2 = value1.X - value2.X;
+			float num = value1.Y - value2.Y;
+			float num3 = (num2 * num2) + (num * num);
+			return (float)Math.Sqrt(num3);
```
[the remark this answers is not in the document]
hakusaro: @Olink ya, it's complicated. We get things like this free from XNA.

TShockAPI/TSPlayer.cs
```diff
@@ -285,6 +285,21 @@ public Group Group
 
 		public bool IgnoreActionsForClearingTrashCan;
 
+		/// <summary>Checks to see if active throttling is happening on events by Bouncer. Rejects repeated events by malicious clients in a short window.</summary>
+		/// <returns>If the player is currently being throttled by Bouncer, or not.</returns>
+		public bool IsBouncerThrottled()
+		{
+			return (DateTime.UtcNow - LastThreat).TotalMilliseconds < 5000;
```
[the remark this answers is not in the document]
hakusaro: That's a good idea. I think this is one of the methods that @QuiCM wants extracted into its own thing.

TShockAPI/GetDataHandlers.cs
hakusaro: Not a clue.

TShockAPI/TSPlayer.cs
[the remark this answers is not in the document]
hakusaro: I'm mildly retarded, actually. This is its own method now. 5k is a good thing to have settable though.

TShockAPI/Utils.cs
```diff
@@ -1456,5 +1456,93 @@ private void OutputElementsForDump(StringBuilder buffer, List<object[]> elements
 				}
 			}
 		}
+
+		/// <summary>Starts an invasion on the server.</summary>
+		/// <param name="type">The invasion type id.</param>
+		internal void StartInvasion(int type)
```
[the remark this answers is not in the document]
hakusaro: I checked into this, but the enum seems...wrong?

```csharp
namespace Terraria.ID
{
  public static class InvasionID
  {
    public const short CachedInvasions = 3;
    public const short CachedFrostMoon = 1;
    public const short CachedPumpkinMoon = 2;
    public const short CachedOldOnesArmy = 3;
    public const short None = 0;
    public const short GoblinArmy = 1;
    public const short SnowLegion = 2;
    public const short PirateInvasion = 3;
    public const short MartianMadness = 4;
    public const short Count = 5;
  }
}
```

hakusaro:
> Maybe for a future commit, but we should derive a class from HandledEventArgs that contains TSPlayer by default.

Good idea.

TShockAPI/Utils.cs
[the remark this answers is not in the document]
DeathCradle: @hakusaro None->Martian are correct, the others are superfluous. **Personally** I do not like many of these constants from Terraria.ID for this reason, rather _I would prefer_ an actual Enum as suggested (Intellisense works better this way too).

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1827 @@
...
+			}
+
+			if (args.Player.CheckIgnores())
+			{
+				args.Player.SendTileSquare(tileX, tileY, 3);
+				args.Handled = true;
+				return;
+			}
+
+			if (flag != 0 && flag != 4 // if no container or container2 placement
```
QuiCM: Should probably write an enum for `flag` here

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1827 @@
...
+
+			if (!TShock.Utils.TilePlacementValid(tileX, tileY) || (args.Player.Dead && TShock.Config.PreventDeadModification))
+			{
+				args.Handled = true;
+				return;
+			}
+
+			if (args.Player.CheckIgnores())
+			{
+				args.Player.SendTileSquare(tileX, tileY, 3);
```
QuiCM: https://www.youtube.com/watch?v=aU4pyiB-kq0

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1827 @@
...
+
+		/// <summary>Handles Bouncer's liquid set anti-cheat.</summary>
+		/// <param name="sender">The object that triggered the event.</param>
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnLiquidSet(object sender, GetDataHandlers.LiquidSetEventArgs args)
+		{
+			int tileX = args.TileX;
+			int tileY = args.TileY;
+			byte amount = args.Amount;
+			byte type = args.Type;
```
QuiCM: Can probably enum up `type`

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1827 @@
...
+						else if (TShock.Config.RequireLogin && !args.Player.IsLoggedIn)
+						{
+							args.Player.SendErrorMessage("Please /register or /login to play!");
+						}
+						else if (args.Player.IgnoreActionsForClearingTrashCan)
+						{
+							args.Player.SendErrorMessage("You need to rejoin to ensure your trash can is cleared!");
+						}
+
+						// ??
```
QuiCM: This is intended to put the player back to where they were last recorded (`LastNetPosition.Y - 48` is to not sink the player 48 pixels into the ground [because `LastNetPosition.Y` is the player's foot, but `Teleport(x, y)` puts your head at `Y`])

TShockAPI/Bouncer.cs
```diff
@@ -0,0 +1,1827 @@
...
+
+			if (y < 0 || y >= Main.maxTilesY)
+			{
+				args.Handled = true;
+				return;
+			}
+
+			//style 52 and 53 are used by ItemID.Fake_newchest1 and ItemID.Fake_newchest2
+			//These two items cause localised lag and rendering issues
+			if (type == TileID.FakeContainers && (style == 52 || style == 53))
```
QuiCM: I'm not sure why the comment refers to ItemID values but then uses magic numbers

TShockAPI/TSPlayer.cs
```diff
@@ -789,6 +804,48 @@ public void Spawn(int tilex, int tiley)
 			}
 		}
 
+		/// <summary>Checks to see if this player object has access rights to a given projectile. Used by projectile bans.</summary>
+		/// <param name="index">The projectile index from Main.projectiles (NOT from a packet directly).</param>
+		/// <param name="type">The type of projectile, from Main.projectiles.</param>
+		/// <returns>If the player has access rights to the projectile.</returns>
+		public bool HasProjectilePermission(int index, int type)
```
QuiCM: This method contains 'Permission' in the name, but as far as I can see there's no permission checks (eg does this block someone with the `*` permission from using a dirt rod if the dirt rod is a banned item?)

Enerdy: It doesn't, and that's because `ItemIsBanned` takes a `TSPlayer` as its second parameter, to perform checks against that specific player I believe. Method name is certainly missleading. But this is why you see us pass the player object (`this`) to it.

PR #1573, opened 2017-12-21
Rebuilt TSPlayer.CheckIgnores() as TSPlayer.IsBeingDisabled(). Fixes #1561.

### LINKED ISSUE #1561: F: Rewrite TSPlayer.CheckIgnores()
Per #1560, I wanted to add a loose end: `TSPlayer.CheckIgnores()` and the related fields it touches are completely and utterly bad. We effectively use strings as multi-purpose data types that cross between boolean and string in different parts of the code. This needs to be unwound and rebuilt in such a way that the replacement method isn't as bad as where we started. This is like, `TSPlayer.IsCheatingFlagged` and a different data structure holding reasons. That's probably all we need to do to make this better.

hakusaro:
This addresses the problem of what `TSPlayer.CheckIgnores()` is actually checking in a clear way. It rewrites all of the "ignores" into booleans, removes some redundant messages, and generally improves code quality.

Each commit message has a pretty good rationale behind it. When reviewing, it's easier to go to the first commit and use the keyboard shortcuts to go in order of commits (n and p) to see how this story was worked. You can read each rationale and see the diff as it was implemented.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -857,7 +857,7 @@ namespace TShockAPI
 						}
 						else if (args.Player.IsDisabledForSSC)
 						{
-							args.Player.SendErrorMessage("Disabled. You need to {0}login to use your server inventory and items.", TShock.Config.CommandSpecifier);
+							args.Player.SendErrorMessage("Disabled. You need to {0}login to load your saved data.", TShock.Config.CommandSpecifier);
 						}
 						else if (TShock.Config.RequireLogin && !args.Player.IsLoggedIn)
 						{
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -863,7 +863,7 @@ namespace TShockAPI
 						{
 							args.Player.SendErrorMessage("Please /register or /login to play!");
 						}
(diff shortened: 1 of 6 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -847,17 +847,17 @@ internal void OnPlayerUpdate(object sender, GetDataHandlers.PlayerUpdateEventArg
...
+							args.Player.SendErrorMessage("Disabled. You went too far with hacked item stacks.");
 						}
 						else if (args.Player.IgnoreActionsForDisabledArmor != "none")
 						{
 							args.Player.SendErrorMessage("Disabled for banned armor: " + args.Player.IgnoreActionsForDisabledArmor);
 						}
 						else if (args.Player.IsDisabledForSSC == true)
 						{
-							args.Player.SendErrorMessage("Disabled. Server side characters is enabled, and you aren't logged in.");
+							args.Player.SendErrorMessage("Disabled. You need to {0}login, since server side characters is enabled.", TShock.Config.CommandSpecifier);
```
QuiCM: >SSC is enabled

>Server side characters **are** enabled

imo use 'are' for grammatical niceness, though I could argue it both ways

PR #1576, opened 2017-12-22
Move TShock.HackedInventory to TSPlayer.HasHackedItemStacks, OnGetSection -> Bouncer

### LINKED ISSUE #1566: F: Rewrite TShock.HackedInventory()
Similar to #1561, this needs to be isolated and put somewhere else related to Bouncer. I'm taking this issue, but using this issue to track progress.

### LINKED ISSUE #1569: Create a HandledEventArgs that includes a player by default
Per https://github.com/Pryaxis/TShock/pull/1560#pullrequestreview-84337221, we should have a `HandledEventArgs` that includes a `TSPlayer` object by default. This would decrease code copypasta in the codebase.

Blocked on #1560 merging before this.

hakusaro:
This fixes #1566 and fixes #1569. First and foremost, it moves the `TShock.HackedInventory` method out of the TShock main class and to `TSPlayer` as `HasHackedItemStacks` with an option to send a player messages for violations. It also simplifies some code in `OnSecondUpdate` (or rather, normalizes it with the rest of TShock) and also moves `OnGetSection` from GDH to `Bouncer`.

In GDH, I implemented a new `GetDataHandledEventArgs` to be the common ancestor to all other event arguments in GDH so that we don't have repeated code and always expose the raw memory stream to plugins if we don't bother to update TShock in a timely manner to supply all hooks with the most up to date data.


hakusaro:
This also fixes #1569.

TShockAPI/TShock.cs
```diff
@@ -1096,17 +1096,10 @@ private void OnSecondUpdate()
...
+						if (!player.HasPermission(Permissions.ignorestackhackdetection))
 						{
-							if (!player.HasPermission(Permissions.ignorestackhackdetection) && (item.stack > item.maxStack || item.stack < 0) &&
-								item.type != 0)
-							{
-								check = "Remove item " + item.Name + " (" + item.stack + ") exceeds max stack of " + item.maxStack;
-								player.SendErrorMessage(check);
-								break;
-							}
+							player.IsDisabledForStackDetection = player.HasHackedItemStacks(true);
```
QuiCM: Personally I'd like to see this as `= player.HasHackedItemStacks(shouldSendMessage: true)`

While this doesn't change any of the behaviour, it does mean someone reading this line can tell what 'true' is doing.
For clarity's sake too, renaming `shouldSendMessage` to `warnPlayer` or somesuch.

Finally, this is getting run every second. If there's no cooldown on the messages then they'll get spammed

hakusaro: Cool, didn’t know you could do that syntactically in C#. I’ll change that.

They’re getting spammed already, and I personally don’t know what to say about that. Right now I’m keeping parity with the existing code, but I don’t believe we should continue running this loop at all (OnSecondUpdate) if we can avoid it.

I would rather players manually tell us if they’ve complied or rely on a different hook. This method probably eats a ton of resources.

Ijwu: Keyword args++

Parity atm is probably a good idea until we can sort out the best improvements to make. This is a big refactoring endeavor.

hakusaro:
@QuiCM @Ijwu thanks guys! <3 fixed it for y'all!

hakusaro:
This issue needs to be merged to fix general-devel disabling all players automatically forever.

PR #1579, opened 2017-12-22
Rewrite TShock.CheckTilePermission

### LINKED ISSUE #1567: F: ReWrite TShock.CheckTilePermission()
Similar to #1561, this method needs to be broken up into multiple subsystem specific checks that return specific things, not just one great check.

hakusaro:
**Requires merging of #1578 first.** **Requires merging of #1576 first.**

Fixes #1567. Implements `TSPlayer.IsInRange` and `TSPlayer.HasBuildPermissions` primarily, with some extra methods for dealing with ice tiles and painting.

General improvements accomplished through this PR:

* Less code in the TShock main class.
* More clear control flow in that `CheckTilePermission` doesn't exist, doesn't have two overloads, and doesn't do about 1.5 million things.
* Faster short circuit logic when checking build permissions. Prioritizes things we have in memory and fast, then things that are in memory and slower, and then things that take logic to process for speed.

Still todo:

- [x] Implement rate throttling on warning messages sent by `HasBuildPermissions`. Remove some of the old throttling code too.


hakusaro:
This is ready to go, but needs the other two dependent pull requests merged first **or merging this branch will merge the other two anyway.**

TShockAPI/TSPlayer.cs
```diff
@@ -321,30 +530,147 @@ public bool IsBeingDisabled()
 
 		public bool SilentJoinInProgress;
 
-		/// <summary>
-		/// A list of points where ice tiles have been placed.
-		/// </summary>
-		public List<Point> IceTiles;
+		/// <summary>Checks if a player has permission to modify a tile dependent on range checks.</summary>
```
QuiCM: Should actually be along the lines of ```<summary>Checks if a given coordinate (x, y) is within a given range of the player's position</summary>```, as there is no permission checks here or any tile references

TShockAPI/TSPlayer.cs
```diff
@@ -321,30 +530,147 @@ public bool IsBeingDisabled()
...
+				IceTiles.Add(new Point(x, y));
+				return true;
+			}
+
+			// The edit wasn't an add, so we check to see if the position matches any of the known ice tiles
+			if (editAction == GetDataHandlers.EditAction.KillTile)
+			{
+				foreach (Point p in IceTiles)
+				{
+					// If they're trying to kill ice or dirt, and the tile was in the list, we allow it.
```
QuiCM: Why is dirt allowed?

hakusaro: I’m guessing because another client might report the melt and have it processed before our client does. I don’t know fully.

TShockAPI/TSPlayer.cs
hakusaro: Oops. I’ll get on fixing that ASAP.

TShockAPI/TSPlayer.cs
```diff
@@ -530,30 +530,147 @@ public bool HasHackedItemStacks(bool shouldWarnPlayer = false)
...
+			{
+				case BuildPermissionFailPoint.GeneralBuild:
+					SendErrorMessage("You lack permission to build on this server.");
+					break;
+				case BuildPermissionFailPoint.SpawnProtect:
+					SendErrorMessage("You lack permission to build in the spawn point.");
+					break;
+				case BuildPermissionFailPoint.Regions:
+					SendErrorMessage("You lack permission to build in this region.");
+					break;
```
mistzzt: Enumerations do make codes more clear here. 
However, this method still has side-effects. When the check method is called, it doesn’t only return whether player has build permission (in this case): the player will receive an error message as well as update the `lastPermissionWarning`. This is unrecommanded since the check methods do more than one thing. Should we change it?

mistzzt: This is likely a (bad?) design pattern from the previous “checkPermission” implementations.

QuiCM: I don't understand what you're trying to convey here

hakusaro: I concur?

hakusaro: Reposting this from Discord: basically my goal right now is to make it understandable what this does and move it out of the TShock main class. I didn’t want to implement a separate warning method and chain it to every one of the 20 some calls to this method for parity and cleanliness. I foresee regions and spawn protection being isolated in the future, which will make splitting this up further easier.

My biggest fear is accidentally introducing a bug in the old code that gets migrated to a new system whilst broken. It’ll be harder to trace bugs if we introduce them now, rather than after the system is split up.

PR #1585, opened 2017-12-25
Fix #1584 / Revert "Change Utils.SearchProjectile to return -1 in error"

### LINKED ISSUE #1584: Normal projectiles will be removed repeatedly in build 2102
*This issue would never happen in stable version (4.3.25).*

There's a check on `NewProjectile` in [Bouncer](https://github.com/Pryaxis/TShock/blob/general-devel/TShockAPI/Bouncer.cs#L981):
```csharp
// from `GetDataHandlers`
// it seems that `SearchProjectile` could never find the proj id here
var index = TShock.Utils.SearchProjectile(ident, owner);

			if (index > Main.maxProjectiles || index < 0)
			{
				args.Player.RemoveProjectile(ident, owner);
				args.Handled = true;
				return;
			}
```
```csharp
		/// <summary>
		/// Searches for a projectile by identity and owner
		/// </summary>
		/// <param name="identity">identity</param>
		/// <param name="owner">owner</param>
		/// <returns>projectile ID or -1 if not found</returns>
		{
			for (int i = 0; i < Main.maxProjectiles; i++)
			{
				if (Main.projectile[i].identity == identity && Main.projectile[i].owner == owner)
					return i;
			}
			return -1;
		}
```
In the master branch, the return value of `Utils.SearchProjectile` if projectile not found is `1000`; it was changed to `-1`. 
Since `Main.maxProjectiles` is 1000, statements in *if-clause* is never executed in stable build. 
However, as it was changed to -1, every new projectile will be removed.

hakusaro:
This reverts commit e19fd22fe0a5600af0c47c07eade94aff119f343.

Not entirely sure why but in #1584 notes that we've regressed in behavior and that the previous function worked fine. This leads me to believe that negatives have an important value in the server somewhere. As a result, I want to revert the changes brought about by Bouncer related to this projectile check.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -978,7 +978,7 @@ namespace TShockAPI
 			short type = args.Type;
 			int index = args.Index;
 
-			if (index > Main.maxProjectiles || index < 0)
+			if (index > Main.maxProjectiles)
 			{
 				args.Player.RemoveProjectile(ident, owner);
 				args.Handled = true;
--- a/TShockAPI/Utils.cs
+++ b/TShockAPI/Utils.cs
@@ -925,7 +925,7 @@ namespace TShockAPI
 		/// </summary>
 		/// <param name="identity">identity</param>
 		/// <param name="owner">owner</param>
-		/// <returns>projectile ID or -1 if not found</returns>
+		/// <returns>projectile ID</returns>
 		public int SearchProjectile(short identity, int owner)
 		{
 			for (int i = 0; i < Main.maxProjectiles; i++)
@@ -933,7 +933,7 @@ namespace TShockAPI
 				if (Main.projectile[i].identity == identity && Main.projectile[i].owner == owner)
 					return i;
 			}
-			return -1;
+			return 1000;
 		}
 
 		/// <summary>


mistzzt:
Removing the index check should solve this problem.
btw, I think there's no need to revert `SearchProjectile`.

hakusaro:
@mistzzt at least 50% of this was just to go back to the previous functionality. What does the index check do or not do?

mistzzt:
[the previous functionality](https://github.com/Pryaxis/TShock/blob/master/TShockAPI/GetDataHandlers.cs):
```csharp
			// always return 1000 here
			var index = TShock.Utils.SearchProjectile(ident, owner);

			if (OnNewProjectile(ident, pos, vel, knockback, dmg, owner, type, index))
				return true;

			// index = 1000, index > Main.maxProjectiles = false
			if (index > Main.maxProjectiles || index < 0)
			{
				// never executed
				args.Player.RemoveProjectile(ident, owner);
				return true;
			}
```
As codes in if-clause were never executed, I don't know what the index check intents to do. Does anyone has ideas?

hakusaro:
My _personal_ thinking is that we should just remove the entire if clause and keep the old search projectile functionality the same.

mistzzt:
The if-clause was introduced in [this commit](https://github.com/Pryaxis/TShock/commit/43b05ec3ef57826729ca946f65bf2e46afa2f998).

As for the search projectile function, don't you think that returning -1 is better than 1000?

hakusaro:
> As for the search projectile function, don't you think that returning -1 is better than 1000?

For sure, but I don't think that changing it arbitrarily if it introduces unexpected consequences is a good plan. It could be detrimental to keep it this way in the event that the function is used by other people or other places in TShock -- I haven't looked recently on the latter. On the former, though, we could be creating a huge number of situations where potentially we've mucked things up. Given that we've already broken one thing by changing this behavior, I'm not entirely sure I want to guarantee we haven't broken something else either.

mistzzt:
Yeah, you're right. 
> My personal thinking is that we should just remove the entire if clause and keep the old search projectile functionality the same.

hakusaro:
> Yeah, you're right.

I looked at the commit you linked, and unfortunately, I can't discern if that's the actual origin of the check or not. @cheahjs can probably chime in here if he remembers back to 2011. I want to hope this was just a check to prevent a _client_ crash, in which case we can probably remove it as it will have probably been fixed in the 6 years since Terraria came out.

hakusaro:
Ready for someone to review/merge this.

PR #1587, opened 2017-12-27
Remove old, deprecated, and broken code

hakusaro:
This is the general pull request for removal of code in any of the following capacities:

* Deprecated.
* Obsolete.
* Terrible.
* Broken.

- [x] Remove string nulls from that one method in utils.


hakusaro:
Should be ready to go.

PR #1593, opened 2018-01-01
Fix tile square checking in Bouncer

mistzzt:
Based on latest codes, players without build permissions can place objects like *item frame* everywhere.
tl;dr The tile square event should be handled inside bouncer.

References:
https://github.com/mistzzt/TShock/blob/85df1533d7b97db3f9fd1010ede29fe10f2cc736/TShockAPI/GetDataHandlers.cs#L1893


hakusaro:
I'm really sorry, I don't know how that happened.

PR #1602, opened 2018-02-18
Revert "Remove IP reset code in TShock init"

### LINKED ISSUE #1601: REST API fails to launch
In the current gen-dev build we have an issue whereby TShock is loaded before `Terraria.Netplay.ServerIP` is filled.
Unfortunately, the following is passed to the REST constructor: `new SecureRest(Netplay.ServerIP, Config.RestApiPort)`.
When `RestApi.Start()` is called, the null value of `Netplay.ServerIP` results in the entire server failing to start.

This can be worked around by starting REST **after** the server has fully initialized (`ServerApi.Hooks.GamePostInitialize`), however this requires that a world is loaded before REST can be used.

It appears that https://github.com/Pryaxis/TShock/pull/1583 is the culprit for this regression.
@hakusaro can you confirm if the vanilla server *does* actually provide `Netplay.ServerIP` with a value before TShock initializes?
Every travis build after #2089 apparently experiences this issue

hakusaro:
Reverts Pryaxis/TShock#1583. Fixes #1601.

### DIFF
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -220,6 +220,10 @@ namespace TShockAPI
 
 				Main.ServerSideCharacter = ServerSideCharacterConfig.Enabled;
 
+				//TSAPI previously would do this automatically, but the vanilla server wont
+				if (Netplay.ServerIP == null)
+					Netplay.ServerIP = IPAddress.Any;
+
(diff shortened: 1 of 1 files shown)


hakusaro:
No changelog required.

PR #1606, opened 2018-03-06
Kicks characters with keys that cannot be obtained without using numpad codes.

djpiper28:
Does exactly what the title says.

Yes, I have.


bartico6:
other than a minor grammar issue ("Illegal character in **you** name" in kick section) this seems okay code-wise, but it should be a configurable option

djpiper28:
I shall try to make it configurable  and then make another pull request.

bartico6:
Just edit this one mate

djpiper28:
Look I am a github noob (LOL), how do I push the latest changes I made to my fork to this pull request.

LOL I can develop but not use github. 100% quality.

djpiper28:
Lol I just did it. I am a fool sometimes.

TShockAPI/Bouncer.cs
```diff
@@ -36,6 +36,13 @@ namespace TShockAPI
 	/// <summary>Bouncer is the TShock anti-hack and anti-cheat system.</summary>
 	internal sealed class Bouncer
 	{
+		private ConfigFile config; //Used for UsernameHasBannedChar checks.
+
+		public void setConfig(ConfigFile configuration)
```
QuiCM: C# Pascal Case conventions dictate this should be SetConfig - however TShock.Config is a static reference so there is no need to pass the object to Bouncer

hakusaro:
@mistzzt can Chinese/non latin IMEs have non-english names?

TShockAPI/Bouncer.cs
```diff
@@ -65,6 +72,20 @@ internal Bouncer()
 			GetDataHandlers.TileEdit += OnTileEdit;
 			GetDataHandlers.MassWireOperation += OnMassWireOperation;
 		}
+		// As in a letter
+		internal Boolean UsernameHasBannedChar(String wholeusername)
```
QuiCM: Preferred to use primitive type `string` rather than object type `String`

hakusaro:
Thanks for your PR! I'll code review and ask for changes pending sign-off that Chinese servers aren't broken with this change, and after you've fixed the compiler error.

TShockAPI/Bouncer.cs
```diff
@@ -65,6 +72,20 @@ internal Bouncer()
...
+		{
+			String[] letters = wholeusername.ToLower().Select(x => x.ToString()).ToArray();
```
QuiCM: Can be simplified to `wholeusername.ToCharArray()`

TShockAPI/Bouncer.cs
```diff
@@ -65,6 +72,20 @@ internal Bouncer()
...
+			String[] allowedCharacters = "abcdefghijklmnopqrstuvwxyz1234567890!\"�$%^&*()-_=/\\+{}[]';@:,.<>?|#~".Select(x => x.ToString()).ToArray();
```
QuiCM: Should be class-scoped - as is this is created and populated in every single call to the variable.

This string is also not configurable by the end user and disallows spaces in usernames.
A regex would be better suited for the use case

TShockAPI/Bouncer.cs
QuiCM: `bool` instead of `Boolean` too

TShockAPI/Bouncer.cs
```diff
@@ -81,6 +102,13 @@ internal void OnGetSection(object sender, GetDataHandlers.GetSectionEventArgs ar
 				args.Handled = true;
 				return;
 			}
+			
+			if (UsernameHasBannedChar(args.Player.Name) && this.config.KickInvalidUsernames)
+			{
+				args.Player.Kick("Illegal character in you name.", true, true);
```
QuiCM: you => your.

Kick message could use further revising

TShockAPI/Bouncer.cs
```diff
@@ -453,6 +481,11 @@ internal void OnItemDrop(object sender, GetDataHandlers.ItemDropEventArgs args)
 			}
 		}
 
+		internal void setConfig(ConfigFile config)
```
QuiCM: Unnecessary method

TShockAPI/TShock.cs
```diff
@@ -331,6 +332,7 @@ public override void Initialize()
 				var systemRam = StatTracker.GetFreeSystemRam(ServerApi.RunningMono);
 				if (systemRam > -1 && systemRam < 2048)
 				{
+					//Little note windows 10 or higher needs 4GB and that should be enough for TShock to use after windows 10 eats several GB f it.
```
QuiCM: [citation needed]

TShockAPI/ConfigFile.cs
```diff
@@ -30,6 +30,10 @@ namespace TShockAPI
 	/// <summary>ConfigFile - The config file class, which contains the configuration for a server that is serialized into JSON and deserialized on load.</summary>
 	public class ConfigFile
 	{
+		/// <summary>True- a username with a symbol like ☺ or ╦  will be kicked. Allowed symbols (other than numbers and letters) !2£$%^&*(){}:@~?,./;'#[]-=\|</summary>
```
QuiCM: This summary doesn't display the full range that is actually blocked by the hardcoded string

TShockAPI/ConfigFile.cs
```diff
@@ -30,6 +30,10 @@ namespace TShockAPI
...
+		[Description("When enabled a user joining with a username that has characters thataren't on the keyboard are kicked.")]
```
QuiCM: thataren't => that aren't

djpiper28:
Will fix stuff when I have the time

Ijwu:
This seems rather out of scope of TShock. This would be better suited as a plugin. The TShock config file has too many options as it is. Plugins are inherently optional, anyway. I suggest transplanting your logic to a plugin and submitting it to the TShock.co resource manager.

ivanbiljan:
I second @QuiCM and @Ijwu 's comments.

mistzzt:
@hakusaro Players could have non-ASCII names with their IME enabled.

hakusaro:
I concur with @mistzzt and @Ijwu and @ivanbiljan given that non-English character names would be effectively shafed from this feature. About 50% of TShock users are Chinese, so...

Ijwu:
The argument may be made that this PR provides a configuration and means to disable the feature. I would argue that less things need to be in the TShock config, not more. Bouncer should get its own config that's separate from TShock's administration features.

hakusaro:
No activity/response from the author, code didn't compile, and unlikely to be merged.

PR #1617, opened 2018-04-20
Remove the stat tracker

hakusaro:
This pull request removes the stat tracker.


QuiCM:
Do we want to remove it entirely or disable it completely for potential re-adding at a later date?

hakusaro:
If we want to re-add it, we can always git revert this commit or write a better plugin on this side. Since we know for a fact that we've been arguing about what it should and shouldn't do, and we knew it needed changes, I don't see a reason to keep it.

PR #1619, opened 2018-04-26
Don't send PlayerTeam packet twice in SetTeam

DankRank:
Discord discussion:
image
image

---
I tested this change with a plugin of mine which uses SetTeam, and it seems to work.

I don't think changelog entry is needed, because this is just an optimization, with no significant behavioural changes. Please tell me if I'm wrong.

### DIFF
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1520,7 +1520,6 @@ namespace TShockAPI
 		{
 			Main.player[Index].team = team;
 			NetMessage.SendData((int)PacketTypes.PlayerTeam, -1, -1, NetworkText.Empty, Index);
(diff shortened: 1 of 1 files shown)


QuiCM:
Based on [Terraria code](https://github.com/Pryaxis/Sources/blob/master/1.3.5.3/Terraria/MessageBuffer.cs#L1516):
```C#
MessageBuffer.cs
    public void GetData(int start, int length, out int messageType)
    {
        ...
        case 45: //team change
            if (Main.netMode != 2)
              break;
            NetMessage.SendData(45, -1, this.whoAmI, (NetworkText) null, number21, 0.0f, 0.0f, 0.0f, 0, 0, 0);
            ...
            break;
        ...
    }
```
I'm going to make an assumption that once upon a time the second send was necessary for clients to understand team changes, however it appears that this is no longer the case

LGTM 👍

PR #1620, opened 2018-04-27
Make Kick/BanOnMediumcoreDeath behave like their Hardcore equvivalents

### LINKED ISSUE #1618: KickOnMediumcoreDeath has false-positives
TShock version: 035e9d114b4bd31d5ace1be98277a981a6fb1202
This issue also applies to BanOnMediumcoreDeath.

It kicks Mediumcore players when they respawn after:
- they used Ice Mirror (id 3199)
- they used Recall Potion (id 2350)
- a plugin called Spawn() on a TSPlayer

Also, it (theorethically) does *not* kick them [when inventory slot 50 is selected](https://github.com/Pryaxis/TShock/blob/035e9d114b4bd31d5ace1be98277a981a6fb1202/TShockAPI/GetDataHandlers.cs#L2447) (which is the first ammo slot, I think). This was probably meant to be a check for Magic Mirror (id 50), which was properly added in cf7fd6da23aa2c8e424f103df06274ec76ca9947

Is there a reason why it can't be done in HandlePlayerKillMeV2 just like the Hardcore equivalents?

DankRank:
fixes #1618

---
Should we get rid of TSPlayer.InitSpawn? It's not used anywhere else, but it's a public field.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2380,16 +2380,21 @@ namespace TShockAPI
 				}
 			}
 
-			if (args.TPlayer.difficulty == 2 && (TShock.Config.KickOnHardcoreDeath || TShock.Config.BanOnHardcoreDeath))
+			if (args.TPlayer.difficulty != 0)
 			{
-				if (TShock.Config.BanOnHardcoreDeath)
-				{
-					if (!args.Player.Ban(TShock.Config.HardcoreBanReason, false, "hardcore-death"))
+				bool mediumcore = args.TPlayer.difficulty == 1;
+				bool shouldBan = mediumcore ? TShock.Config.BanOnMediumcoreDeath : TShock.Config.BanOnHardcoreDeath;
+				bool shouldKick = mediumcore ? TShock.Config.KickOnMediumcoreDeath : TShock.Config.KickOnHardcoreDeath;
+				string banReason = mediumcore ? TShock.Config.MediumcoreBanReason : TShock.Config.HardcoreBanReason;
+				string kickReason = mediumcore ? TShock.Config.MediumcoreKickReason : TShock.Config.HardcoreKickReason;
+				string banAdminName = mediumcore ? "mediumcore-death" : "hardcore-death";
+
+				if(shouldBan) {
+					if (!args.Player.Ban(banReason, false, banAdminName))
 						args.Player.Kick("You died! Normally, you'd be banned.", true, true);
 				}
-				else
-				{
-					args.Player.Kick(TShock.Config.HardcoreKickReason, true, true, null, false);
+				else if(shouldKick) {
+					args.Player.Kick(kickReason, true, true, null, false);
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


hakusaro:
Ya, go for it.

hakusaro:
(The fewer undocumented stupidly named pointless public fields we have, the better)

TShockAPI/GetDataHandlers.cs
```diff
@@ -2380,16 +2380,21 @@ private static bool HandlePlayerKillMeV2(GetDataHandlerArgs args)
...
 			{
-				if (TShock.Config.BanOnHardcoreDeath)
-				{
-					if (!args.Player.Ban(TShock.Config.HardcoreBanReason, false, "hardcore-death"))
+				bool mediumcore = args.TPlayer.difficulty == 1;
+				bool shouldBan = mediumcore ? TShock.Config.BanOnMediumcoreDeath : TShock.Config.BanOnHardcoreDeath;
+				bool shouldKick = mediumcore ? TShock.Config.KickOnMediumcoreDeath : TShock.Config.KickOnHardcoreDeath;
+				string banReason = mediumcore ? TShock.Config.MediumcoreBanReason : TShock.Config.HardcoreBanReason;
+				string kickReason = mediumcore ? TShock.Config.MediumcoreKickReason : TShock.Config.HardcoreKickReason;
+				string banAdminName = mediumcore ? "mediumcore-death" : "hardcore-death";
```
hakusaro: Stylistically, I don't like hard coding in `mediumcore-death` or `hardcore-death` as names. They look really bleh. How about just like...`World` or `Terraria` or `GameRule` or something? Does it really need to be distinguished between medium/hard if the ban reason covers it?

TShockAPI/GetDataHandlers.cs
```diff
@@ -2380,16 +2380,21 @@ private static bool HandlePlayerKillMeV2(GetDataHandlerArgs args)
 				}
 			}
 
-			if (args.TPlayer.difficulty == 2 && (TShock.Config.KickOnHardcoreDeath || TShock.Config.BanOnHardcoreDeath))
+			if (args.TPlayer.difficulty != 0)
```
hakusaro: For the next person who gets here, can you leave a comment as to what this code block does, and why you have a magic number (0)?

TShockAPI/GetDataHandlers.cs
```diff
@@ -2380,16 +2380,21 @@ private static bool HandlePlayerKillMeV2(GetDataHandlerArgs args)
...
 			{
-				if (TShock.Config.BanOnHardcoreDeath)
-				{
-					if (!args.Player.Ban(TShock.Config.HardcoreBanReason, false, "hardcore-death"))
+				bool mediumcore = args.TPlayer.difficulty == 1;
+				bool shouldBan = mediumcore ? TShock.Config.BanOnMediumcoreDeath : TShock.Config.BanOnHardcoreDeath;
+				bool shouldKick = mediumcore ? TShock.Config.KickOnMediumcoreDeath : TShock.Config.KickOnHardcoreDeath;
```
hakusaro: First: If you get to this point and both shouldBan and shouldKick are false, then we can short circuit and not continue doing stuff. It's a small time optimization.

Second: Is there any way to make this flow control not 6 ternary operators? They all just kinda look bleh to me. Does an if statement or a case/switch look better?

DankRank: 1) I'm 99% sure compiler would inline those. I aimed for readability when writing this code, so I decided to do it like that because it makes the ternaries align nicely. I could easily add a check though.

2) This is how it would look with if statements (switch/case with only two cases would look dumb, imo)
```cs
bool shouldBan, shouldKick;
string banReason, kickReason, banAdminName;
if(args.TPlayer.difficulty == 1) // Player is mediumcore
{
        shouldBan = TShock.Config.BanOnMediumcoreDeath;
        shouldKick = TShock.Config.KickOnMediumcoreDeath;
        banReason = TShock.Config.MediumcoreBanReason;
        kickReason = TShock.Config.MediumcoreKickReason;
        banAdminName = "mediumcore-death";
}
else  // Player is hardcore
{
        shouldBan = TShock.Config.BanOnHardcoreDeath;
        shouldKick = TShock.Config.KickOnHardcoreDeath;
        banReason = TShock.Config.HardcoreBanReason;
        kickReason = TShock.Config.HardcoreKickReason;
        banAdminName = "hardcore-death";
}
```

If I do both 1 and 2, it will be kinda ugly, because it would require one of the ifs appear in both branches of another if. (e.g. in the code above, I'd have to wrap both assignments of kickReason in `if(shouldKick)`, etc)

I honestly would prefer to just do 2 - I think it's the most readable, and doesn't scrafice much (if any) efficency. What do you think?

TShockAPI/GetDataHandlers.cs
DankRank: It should probably be something unique, so that it stands out from any possible admin names. How about `@TSHOCK@`?

hakusaro: TShock is never capitalized like that. That's a silly thing to do stylistically.

TShockAPI/GetDataHandlers.cs
hakusaro: Honestly it all sucks stylistically. I guess the way it is now is fine -- but ugh.

TShockAPI/GetDataHandlers.cs
DankRank: I'll just go with `TShock` then.

Issue #1633, opened 2018-08-28
Update checker should use a TLS endpoint instead of raw HTTP


hakusaro:
As mentioned in #1632, we need to change the update checker to use TLS instead of HTTP. This means both a code change and a verification on my end that the DNS settings in CloudFlare won't prohibit us from making automated GETs (because of the bot filtering).

This is a great first time issue for someone who wants to get involved on the project (it requires adding an `s` to a URL).

hakusaro:
This was solved in #1638.

PR #1638, opened 2018-10-03
Update Manager now uses TLS.

pandabear41:
Simple PR for Hacktoberfest.

Some things to change on your Cloudflare or I need to add code to allow the specific self-signed Cloudflare certificate. 

> * [x]  Need to configure cloudflare to use the Universal SSL cert. https://support.cloudflare.com/hc/en-us/articles/200170566-Why-isn-t-SSL-working-for-my-site- (@hakusaro)
> * [x]  Test code in Mono and .NET to see if it works or generates an HttpRequestException exception. (@pandabear41)

Solves issue #1633

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 This is the rolling changelog for TShock for Terraria. Use past tense when adding new entries; sign your name off when you add or change something. This should primarily be things like user changes, not necessarily codebase changes unless it's really relevant or large.
 
 ## Upcoming Changes
+* Update tracker now uses TLS (@pandabear41) 
 * API: Added return in OnNameCollision if hook has been handled. (@Patrikkk)
 * API: Added hooks for item, projectile and tile bans (@deadsurgeon42)
 * API: Changed `PlayerHooks` permission hook mechanisms to allow negation from hooks (@deadsurgeon42)
--- a/TShockAPI/UpdateManager.cs
+++ b/TShockAPI/UpdateManager.cs
@@ -33,7 +33,7 @@ namespace TShockAPI
 	/// </summary>
 	public class UpdateManager
 	{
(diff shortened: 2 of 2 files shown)


hakusaro:
Backstory: The reason why TLS on update.tshock.co wasn't working is because the certificate on that host is the CloudFlare origin certificate, which enables us to have encrypted connections to upstream CloudFlare servers, but not much else. The `update.tshock.co` endpoint was actually hitting the backend server directly, rather than going through CloudFlare at all, simply as a result of a quirk of our infrastructure. Since it's an API endpoint, the requests to it come in as obviously not browsers. This means that CloudFlare's automatic security system (which is set to high) would likely throw out these requests as they're obviously automated.

In order to disable the security system on a per app basis, we needed to add a page rule that would whitelist that subdomain. But we were all out of page rules until now -- when I realized that we had a page rule for a domain that was no longer in use. By removing that page rule, we were able to put `update.tshock.co` behind CloudFlare, allowing it to be proxied with TLS successfully.

screen shot 2018-10-02 at 20 18 07

hakusaro:
@pandabear41 FYI: This works on Mono a-okay.

Tested on:

```
Mono JIT compiler version 5.4.1.6 (tarball Mon Dec 11 14:59:42 GMT 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
```

pandabear41:
Tested with .NET, no more unsigned cert exceptions.

Ijwu:
Thanks for helping us out!

PR #1644, opened 2018-10-23
Added PlayerZone handler and bouncer, detect lunar towers cheat

AxeelAnder:
If a client send zone packet contains lunar towers, server would spawn corresponding monsters even no towers really exist.


mistzzt:
As far as I know, client would send *playerzone* packets quite frequently -- thus i wonder if it's possible for us to have some "cache" mechanisms to prevent every linear searching on NPCs?

Also, you may skip the test in case of players not in any zone.

AxeelAnder:
> 
> 
> As far as I know, client would send _playerzone_ packets quite frequently -- thus i wonder if it's possible for us to have some "cache" mechanisms to prevent every linear searching on NPCs?

image
Profiled by resharper, it cost less than I expected.

PR #1666, opened 2019-03-30
Backport removal of stat tracker to master branch

hakusaro:
This will update the master branch version of TShock to comply with GDPR by removing the stat tracker.

This fixes #1634 when it's released.

PR #1667, opened 2019-03-30
Backport (frontport) version number from future master to gen-dev

hakusaro:
Since general-devel always has to be up to date as compared to master, we should update the version on general-devel to be what master will be if #1666 is merged.


hakusaro:
@QuiCM this is also good to go

PR #1681, opened 2019-10-02
Add GitHub Actions build

sgkoishi:
Fix the build script and add the badge to the readme.


.github/workflows/msbuild-test.yml
```diff
@@ -7,7 +7,35 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@master
-    - uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: '4.5.1'
-    - run: dotnet build --configuration Release ./TShockAPI/TShockAPI.csproj
+        submodules: recursive
+    - name: Download MSBuild installer
+      run: powershell "$source = (New-Object System.Net.WebClient).DownloadString(\"https://gist.githubusercontent.com/sgkoishi/87e613881f9164bf03849603dd4f4266/raw/\") ; Add-Type
```
hakusaro: Can you add this script to the repo itself? We can't really link to a random gist from inside this repo for security purposes.

hakusaro (changes requested):
Thank you so so much! I really appreciate this. Can you address the one comment? If we went forward with this, we would actually love to have artifact downloading as part of GitHub actions CI, possibly to just eliminate AppVeyor/Travis entirely, too. If you can't do that, that's okay, but it would be an awesome add.

According to the docs, I think msbuild should actually be included in the virtual environment actions uses on windows? I could be entirely wrong, but it might be worth checking to see if you need to install it.

sgkoishi:
> 
> 
> Thank you so so much! I really appreciate this. Can you address the one comment? If we went forward with this, we would actually love to have artifact downloading as part of GitHub actions CI, possibly to just eliminate AppVeyor/Travis entirely, too. If you can't do that, that's okay, but it would be an awesome add.
> 
> According to the docs, I think msbuild should actually be included in the virtual environment actions uses on windows? I could be entirely wrong, but it might be worth checking to see if you need to install it.

I've added the gist to repo, and the artifact is already included. ([example](https://github.com/sgkoishi/TShock/runs/245623098)) ([code](https://github.com/Pryaxis/TShock/pull/1681/commits/79b092b9c13e21df684f342708ef0f74bb72cd4e#diff-df91aecff22f5180353b36928fc4042aR38-R41))

MSBuild: I'm not quite sure if there is a msbuild in the environment - but at least it's not in the PATH environment variable. NuGet is also required since all these projects use `packages.config` file instead of `PackageReference` in csproj, which msbuild could not handle it.

.github/workflows/msbuild-test.yml
sgkoishi: [Updated](https://github.com/Pryaxis/TShock/pull/1681/commits/fdbc958b3481680f3ca575e9c24fe4debb100597#diff-24fc619a40b157250cc2bc1b212b4689).

hakusaro:
> I've added the gist to repo, and the artifact is already included. (example) (code)

Thank you so, so much! I totally just didn't see it. :D

hakusaro (approved):
lgtm_6

hakusaro:
@sgkoishi actually, before I forget: if you want, you can update the name + file to something like "Build Server" or something. It doesn't need to be called `MSBuild Test` because it works thanks to you :)

QuiCM:
LGTM :)

hakusaro:
@all-contributors please add @sgkoishi for code, infra

hakusaro:
@sgkoishi I'm sure you're very well aware, but [hacktoberfest](https://hacktoberfest.digitalocean.com/) is happening and you definitely deserve to be rewarded.

PR #1686, opened 2020-02-07
Fix BoC false positive kick in NPCAddBuffTimeMax

moisterrific:
360 ticks = 6 seconds, according to the wiki, the max possible duration is proportional to dmg taken which goes up to 7.5 when taking up to 300 dmg, so 7.5 seconds would be the max duration (aka 450 ticks)

https://terraria.gamepedia.com/Brain_of_Confusion (see the debuff equation for yourselves, extracted from the 1.3.5.3 source code)

Didn't know this was an issue until players on my server were getting bounced. They were messing around and jumping off tall buildings, which did ~300 fall damage. One of them had BoC (Brain of Confusion) equipped, so the confused debuff was likely applied for >6 seconds to a nearby mob and caused that player to be force kicked.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1869,7 +1869,7 @@ namespace TShockAPI
 			{ BuffID.DryadsWardDebuff, 120 },
 			{ BuffID.BetsysCurse, 600 },
 			{ BuffID.Oiled, 540 },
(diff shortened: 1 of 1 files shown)


QuiCM:
I don't see this causing any problems. LGTM ✔️

bartico6:
Looks good, also 👍

PR #1689, opened 2020-02-10
Fix GitHub Actions with updated Windows Server 2019 builder

### LINKED ISSUE #1688: Build Action cannot complete
Seems like something changed in GitHub Actions and now the Build action is failing:

```
Run actions/checkout@master
  > Run actions/checkout@master
  ##[error]The input 'submodules' is not supported in actions/checkout@v2

##[error]Node run failed with exit code 1
```

hakusaro:
This changeset fixes several problems with the github actions build. It
removes the dependency on a repo script to install msbuild + VS. It
switches the shell back to cmd.exe, so that the scripts in build.yml
continue to run. It installs nuget via chocolatey, since that's not on
the path by default. Finally, it also uses built in MSBUILD to satisfy
the first statement.

Fixes #1688.


hakusaro:
This fixes #1688.

QuiCM:
LGTM.
I'd prefer not to have to install nuget via choco but you gotta do what you gotta do

QuiCM (approved):

PR #1691, opened 2020-02-12
Publish artifacts in github for debug builds from CI

hakusaro:
It's helpful to have debug builds available for people to download if
they need them.

Also includes the OTAPI bootstrapper, because why not?


QuiCM:
Is there a reason for 'not debug' vs 'release'?
I'm not against the terminology, it's just a variation from what we've used historically

hakusaro:
Trying to avoid anyone coming to us saying they downloaded the latest "TShock Release" from github, and then finding out it's an experimental build and they don't know what to do with it. We definitely use the terminology everywhere right now but I think making it less visible is a good step in the right direction.

hakusaro:
Mostly because we distribute the actual software on github too, so it can be ambiguous. If they got it from travis, that's ez to ask about.

QuiCM:
Fair enough

QuiCM (approved):

PR #1728, opened 2020-05-07
Tile entity fixes

### LINKED ISSUE #1693: Item Frame and Target Dummy broken
## TShock Version
4.3.26 (Mintaka), API 2.1

## Issue Description
- **Item Frame** and **Target Dummy** disappears to the client when placed, but are still visible to other players online
- **Target Dummy** cannot be interacted with by other players at all

(tested with stock TShock default group, it is worth mentioning this issue does not occur for superadmin group, nor any group with the `ignore.sendtilesquare` permission)

## Steps to Reproduce
1. craft Item Frame and Target Dummy
2. place them in a non-protected area (while in a group that does not have ignore.sendtilesquare perm)

## Supporting Evidence
https://cdn.discordapp.com/attachments/479657350043664386/689561252896440355/18_05_49_42_784.mp4
https://cdn.discordapp.com/attachments/479657350043664386/689561999025242151/18_05_52_39_150.mp4

QuiCM:
This pull request resolves a number of issues, primarily surrounding item frames but also with some touches on the other Tile Entity objects.
* Item frames place properly on the first try
* Item frames return items instead of eating them 
* Objects that span multiple tiles (eg Item Frames which are a 2x2 tile) are now first class citizens for region protection. See ASCII art diagram below
* Resolves #1693 and others of its ilk

Old region protection.
Region borders marked with | & -, item frame marked with x & *
```
 |------|
*|x    x|*
*|x    x|*
 |      |
 |------|
```
In this diagram, the x tiles of the item frame were safe, however the frame itself could still be destroyed by destroying the * sections.
In the new protection, the * parts are recognized as part of the object and the region's protection is extended to them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming Changes
 
+## TShock 4.3.27
+* Fixed a number of bugs relating to Tile Entities (Item Frames, Target Dummies, Logic Sensors)
+
 ## TShock 4.3.26
 * Removed the stat tracking system. (@hakusaro)
 * Updated SQLite binaries. (@hakusaro)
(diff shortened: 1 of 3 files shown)


hakusaro:
I believe this has been superseded by changes on `fix-sts` merged earlier today.

PR #1743, opened 2020-05-14
Provide warning message on invalid group command

### LINKED ISSUE #1742: Provide warning when using group command that doesn't exist
TShock user in Discord reported that `/group addperms Guest tshock.ignore.damage` didn't do anything. This is because they had a typo and `addperms` doesn't exist, but we didn't give a warning message so they didn't know it worked. The group command should give an error message when attempting to run a subcommand that doesn't exist so that they get feedback that something isn't right.

hakusaro:
This fixes #1742 where users could run group commands that didn't exist and the server would silently fail rather than giving an error.


mistzzt (approved):
lgtm

PR #1751, opened 2020-05-17
Ready to go 1.4

AxeelAnder:


TShockAPI/Commands.cs
```diff
@@ -2144,7 +2144,12 @@ private static void ClearAnglerQuests(CommandArgs args)
 
 		private static void ToggleExpert(CommandArgs args)
 		{
-			Main.expertMode = !Main.expertMode;
+			const int NormalMode = 0;
```
QuiCM, line 2147: Constants for this exist in Terraria.ID.GameModeID - We should update to use those once we're happy with this command. For now this is fine

QuiCM (dismissed):
Noted one minor adjustment for when things are stable.
LGTM :)

PR #1757, opened 2020-05-17
1.4.0.2

Patrikkk:


TShockAPI/DB/CharacterManager.cs
```diff
@@ -52,7 +52,7 @@ public CharacterManager(IDbConnection db)
 									 new SqlColumn("shirtColor", MySqlDbType.Int32),
 									 new SqlColumn("underShirtColor", MySqlDbType.Int32),
 									 new SqlColumn("shoeColor", MySqlDbType.Int32),
-									 new SqlColumn("hideVisuals", MySqlDbType.Int32),
+									 new SqlColumn("hideVisibleAccessorys", MySqlDbType.Int32),
```
hakusaro: We can't just change a database column like this because it will break the SSC database for all existing characters. If the data structure for `hideVisibleAccessorys` is just the same as `hideVisuals` then we can just insert the new data into the database and keep the old column name. If the data structure has changed we need to provide a migration path so that people's saved character data is carried over to the new data structure without data loss. It might be better if you just don't update the SSC system yet.

TShockAPI/DB/CharacterManager.cs
```diff
@@ -104,7 +104,7 @@ public PlayerData GetPlayerData(TSPlayer player, int acctid)
 						playerData.shirtColor = TShock.Utils.DecodeColor(reader.Get<int?>("shirtColor"));
 						playerData.underShirtColor = TShock.Utils.DecodeColor(reader.Get<int?>("underShirtColor"));
 						playerData.shoeColor = TShock.Utils.DecodeColor(reader.Get<int?>("shoeColor"));
-						playerData.hideVisuals = TShock.Utils.DecodeBoolArray(reader.Get<int?>("hideVisuals"));
+						playerData.hideVisibleAccessorys = TShock.Utils.DecodeBoolArray(reader.Get<int?>("hideVisibleAccessorys"));
```
hakusaro: Same thing here -- if we do this the character database will break and not be migratable which will orphan all SSC character data.

TShockAPI/DB/CharacterManager.cs
```diff
@@ -174,8 +174,8 @@ public bool InsertPlayerData(TSPlayer player)
...
-						player.User.ID, playerData.health, playerData.maxHealth, playerData.mana, playerData.maxMana, String.Join("~", playerData.inventory), playerData.extraSlot, player.TPlayer.SpawnX, player.TPlayer.SpawnY, player.TPlayer.skinVariant, player.TPlayer.hair, player.TPlayer.hairDye, TShock.Utils.EncodeColor(player.TPlayer.hairColor), TShock.Utils.EncodeColor(player.TPlayer.pantsColor),TShock.Utils.EncodeColor(player.TPlayer.shirtColor), TShock.Utils.EncodeColor(player.TPlayer.underShirtColor), TShock.Utils.EncodeColor(player.TPlayer.shoeColor), TShock.Utils.EncodeBoolArray(player.TPlayer.hideVisual), TShock.Utils.EncodeColor(player.TPlayer.skinColor),TShock.Utils.EncodeColor(player.TPlayer.eyeColor), player.TPlayer.anglerQuestsFinished);
+						"INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, extraSlot, spawnX, spawnY, skinVariant, hair, hairDye, hairColor, pantsColor, shirtColor, underShirtColor, shoeColor, hideVisibleAccessorys, skinColor, eyeColor, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8, @9, @10, @11, @12, @13, @14, @15, @16, @17, @18, @19, @20);",
```
hakusaro: Same here.

TShockAPI/DB/CharacterManager.cs
```diff
@@ -188,8 +188,8 @@ public bool InsertPlayerData(TSPlayer player)
...
-						playerData.health, playerData.maxHealth, playerData.mana, playerData.maxMana, String.Join("~", playerData.inventory), player.User.ID, player.TPlayer.SpawnX, player.TPlayer.SpawnY, player.TPlayer.hair, player.TPlayer.hairDye, TShock.Utils.EncodeColor(player.TPlayer.hairColor), TShock.Utils.EncodeColor(player.TPlayer.pantsColor), TShock.Utils.EncodeColor(player.TPlayer.shirtColor), TShock.Utils.EncodeColor(player.TPlayer.underShirtColor), TShock.Utils.EncodeColor(player.TPlayer.shoeColor), TShock.Utils.EncodeBoolArray(player.TPlayer.hideVisual), TShock.Utils.EncodeColor(player.TPlayer.skinColor), TShock.Utils.EncodeColor(player.TPlayer.eyeColor), player.TPlayer.anglerQuestsFinished, player.TPlayer.skinVariant, player.TPlayer.extraAccessory ? 1 : 0);
+						"UPDATE tsCharacter SET Health = @0, MaxHealth = @1, Mana = @2, MaxMana = @3, Inventory = @4, spawnX = @6, spawnY = @7, hair = @8, hairDye = @9, hairColor = @10, pantsColor = @11, shirtColor = @12, underShirtColor = @13, shoeColor = @14, hideVisibleAccessorys = @15, skinColor = @16, eyeColor = @17, questsCompleted = @18, skinVariant = @19, extraSlot = @20 WHERE Account = @5;",
```
hakusaro: Same here.

TShockAPI/DB/CharacterManager.cs
```diff
@@ -244,7 +244,7 @@ public bool InsertSpecificPlayerData(TSPlayer player, PlayerData data)
 				try
 				{
 					database.Query(
-						"INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, extraSlot, spawnX, spawnY, skinVariant, hair, hairDye, hairColor, pantsColor, shirtColor, underShirtColor, shoeColor, hideVisuals, skinColor, eyeColor, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8, @9, @10, @11, @12, @13, @14, @15, @16, @17, @18, @19, @20);",
+						"INSERT INTO tsCharacter (Account, Health, MaxHealth, Mana, MaxMana, Inventory, extraSlot, spawnX, spawnY, skinVariant, hair, hairDye, hairColor, pantsColor, shirtColor, underShirtColor, shoeColor, hideVisibleAccessorys, skinColor, eyeColor, questsCompleted) VALUES (@0, @1, @2, @3, @4, @5, @6, @7, @8, @9, @10, @11, @12, @13, @14, @15, @16, @17, @18, @19, @20);",
```
hakusaro: & here

TShockAPI/DB/CharacterManager.cs
```diff
@@ -262,7 +262,7 @@ public bool InsertSpecificPlayerData(TSPlayer player, PlayerData data)
 						TShock.Utils.EncodeColor(playerData.shirtColor),
 						TShock.Utils.EncodeColor(playerData.underShirtColor),
 						TShock.Utils.EncodeColor(playerData.shoeColor),
-						TShock.Utils.EncodeBoolArray(playerData.hideVisuals),
+						TShock.Utils.EncodeBoolArray(playerData.hideVisibleAccessorys),
```
hakusaro, line 265: & here

TShockAPI/DB/CharacterManager.cs
```diff
@@ -278,7 +278,7 @@ public bool InsertSpecificPlayerData(TSPlayer player, PlayerData data)
 				try
 				{
 					database.Query(
-						"UPDATE tsCharacter SET Health = @0, MaxHealth = @1, Mana = @2, MaxMana = @3, Inventory = @4, spawnX = @6, spawnY = @7, hair = @8, hairDye = @9, hairColor = @10, pantsColor = @11, shirtColor = @12, underShirtColor = @13, shoeColor = @14, hideVisuals = @15, skinColor = @16, eyeColor = @17, questsCompleted = @18, skinVariant = @19, extraSlot = @20 WHERE Account = @5;",
+						"UPDATE tsCharacter SET Health = @0, MaxHealth = @1, Mana = @2, MaxMana = @3, Inventory = @4, spawnX = @6, spawnY = @7, hair = @8, hairDye = @9, hairColor = @10, pantsColor = @11, shirtColor = @12, underShirtColor = @13, shoeColor = @14, hideVisibleAccessorys = @15, skinColor = @16, eyeColor = @17, questsCompleted = @18, skinVariant = @19, extraSlot = @20 WHERE Account = @5;",
```
hakusaro: & here

TShockAPI/DB/CharacterManager.cs
```diff
@@ -295,7 +295,7 @@ public bool InsertSpecificPlayerData(TSPlayer player, PlayerData data)
 						TShock.Utils.EncodeColor(playerData.shirtColor),
 						TShock.Utils.EncodeColor(playerData.underShirtColor),
 						TShock.Utils.EncodeColor(playerData.shoeColor),
-						TShock.Utils.EncodeBoolArray(playerData.hideVisuals),
+						TShock.Utils.EncodeBoolArray(playerData.hideVisibleAccessorys),
```
hakusaro, line 298: & here

TShockAPI/GetDataHandlers.cs
```diff
@@ -1439,8 +1439,8 @@ private static bool HandlePlayerInfo(GetDataHandlerArgs args)
 			string name = args.Data.ReadString();
 			byte hairDye = args.Data.ReadInt8();
 
-			BitsByte hideVisual = args.Data.ReadInt8();
-			BitsByte hideVisual2 = args.Data.ReadInt8();
+			BitsByte hideVisibleAccessory = args.Data.ReadInt8();
```
hakusaro, line 1442: This is ok to change

TShockAPI/GetDataHandlers.cs
```diff
@@ -1439,8 +1439,8 @@ private static bool HandlePlayerInfo(GetDataHandlerArgs args)
...
+			BitsByte hideVisibleAccessory2 = args.Data.ReadInt8();
```
hakusaro, line 1443: This is ok to chagne

TShockAPI/GetDataHandlers.cs
```diff
@@ -1488,11 +1488,11 @@ private static bool HandlePlayerInfo(GetDataHandlerArgs args)
 				args.Player.TPlayer.shirtColor = shirtColor;
 				args.Player.TPlayer.underShirtColor = underShirtColor;
 				args.Player.TPlayer.shoeColor = shoeColor;
-				args.Player.TPlayer.hideVisual = new bool[10];
+				args.Player.TPlayer.hideVisibleAccessory = new bool[10];
```
hakusaro, line 1491: This is okayish

TShockAPI/PlayerData.cs
```diff
@@ -44,7 +44,7 @@ public class PlayerData
 		public Color? shoeColor;
 		public Color? skinColor;
 		public Color? eyeColor;
-		public bool[] hideVisuals;
+		public bool[] hideVisibleAccessorys;
```
hakusaro, line 47: This is the SSC backing data model so I'm inclined to say that we can't under any circumstances change this without a migration path/strategy to upgrade the data so that we don't lose people's stuff.

TShockAPI/SaveManager.cs
```diff
@@ -126,10 +126,10 @@ private void SaveWorker()
 								if (task.direct)
 								{
 									OnSaveWorld(new WorldSaveEventArgs());
-									WorldFile.saveWorld(task.resetTime);
+									WorldFile.SaveWorld(task.resetTime);
```
hakusaro, line 129: This is good.

TShockAPI/TSPlayer.cs
```diff
@@ -902,7 +902,7 @@ public virtual void GiveItem(int type, string name, int width, int height, int s
 			Main.item[itemid].wet = Collision.WetCollision(Main.item[itemid].position, Main.item[itemid].width,
 															 Main.item[itemid].height);
 			Main.item[itemid].stack = stack;
-			Main.item[itemid].owner = Index;
+			Main.item[itemid].playerIndexTheItemIsReservedFor = Index;
```
hakusaro, line 905: This is good.

TShockAPI/Commands.cs
```diff
@@ -2109,7 +2109,12 @@ private static void ClearAnglerQuests(CommandArgs args)
 
 		private static void ToggleExpert(CommandArgs args)
 		{
-			Main.expertMode = !Main.expertMode;
+			const int NormalMode = 0;
```
hakusaro, line 2112: These should really be an enum imho.

hakusaro (changes requested):
This is pretty good so far but the problem is that it alters the SSC database in such a way that the existing SSC database tables would be incompatible with it and thus hose everyone who's currently using SSC. I don't know enough about the changes yet to know if the `hideVisibleAccessorys` field is the same in structure as `hideVisuals` but simply renamed? Either way this would cause problems because existing schemas would be adversely affected by it.

Other changes seem relatively okay.

Issue #1759, opened 2020-05-18
Firecracker Whip debuff kicks player


kirides:
* TShock version: `TShock 4.4.0 Pre-release 1`
* TShock build number (if known):

#### Reproduction steps (if applicable)?

1. Start the server
2. Have a Firecracker (Whip) equipped
3. Attack an enemy
4. get kicked because of `Added Buff to NPC abnormally`

#### Any stack traces or error messages (if known)?

- 

#### Any screenshots?
![image](https://user-images.githubusercontent.com/13602143/82194580-53e3bf00-98f7-11ea-8fb3-b60f9a09a8b1.png)
![image](https://user-images.githubusercontent.com/13602143/82194632-6958e900-98f7-11ea-8c22-42b051ab7c84.png)

GoobyCorp:
Looks like it's related to this:https://github.com/Pryaxis/TShock/blob/b56f1fc35c0179f4196e827a9a0e327117aa8428/TShockAPI/Bouncer.cs#L1872-L1895
I don't see the effect or it's maximum duration listed in there for the whip.

dly2424:
Also the Snapthorn whip kicks the player for the same reason. Seems tshock doesnt have these buffs/debuffs accounted for.

breeece:
I am also having an issue with getting kicked with the same error for using the durendal. I have my summons active, then I use the Durendal to attack an enemy and I get kicked

hakusaro:
This should be fixed in the latest experimental builds. Please let me know if it isn't!

PR #1768, opened 2020-05-18
Fixed whips kicking players

ndragon798:
This fixes issue #1759. 
Where you would get kicked for using a whip on an enemy since the debuff wasn't allowed.


ndragon798:
Do I have to have the change log update as part of the pull request?

PR #1769, opened 2020-05-18
Fixed getting kicked for whipping mobs

ndragon798:
This fixes issue #1759.
Where you would get kicked for using a whip on an enemy since the debuff wasn't allowed.


Tomo3214:
Which file do i have to overwrite?

ndragon798:
You would have to clone my fork and then compile it with the instructions in the readme.

hakusaro:
Thank you for your contribution! We're going to add #1770 in favour of this though, because the times match the Terraria source as opposed to just being `300` in all cases.

@all-contributors add ndragon798 for code

hakusaro:
@all-contributors add @ndragon798 for code

PR #1811, opened 2020-05-20
Make sqlite db path configurable

AxeelAnder:
https://github.com/Pryaxis/TShock/issues/1810


Patrikkk:
lgtm 👌

hakusaro (approved):

hakusaro:
I think ideally we would make this a command line flag too, but arguably it's not required.

Issue #1809, opened 2020-05-20
TShock 4.4.0 Pre4 Desync upon death


Minescence:
* TShock version: [TShock 4.4.0 Pre4](https://github.com/Pryaxis/TShock/releases/tag/v4.4.0-pre4)
* TShock build number (if known): 

#### Reproduction steps (if applicable)?

1. Die to anything
2. Now you're desynced until you reconnect to server

#### Any stack traces or error messages (if known)?

```
Nothing in console.
```

#### Any screenshots?

![image](https://user-images.githubusercontent.com/51507768/82434461-2b90c780-9a93-11ea-87a0-d128d8f81a07.png)

Client data before dying
![image](https://user-images.githubusercontent.com/51507768/82434949-e751f700-9a93-11ea-9c30-1ec5cf2f5aa4.png)

Client data after dying
![image](https://user-images.githubusercontent.com/51507768/82434537-42cfb500-9a93-11ea-95ea-1af7ff854a82.png)

#### Any log messages from files that end in `.log` or `.txt`?

**[2020-05-19_19-21-01.log](https://github.com/Pryaxis/TShock/files/4656114/2020-05-19_19-21-01.log)**

#### What plugins and what versions of those plugins are you running?

No plugins

dly2424:
Yup, entire fresh install of tshock pre 4 still leads to desyncs on death (still using same world though. Shouldnt matter however)

Ubuntu - using sudo permissions in screen
`sudo screen -L -Logfile log -dmS terraria mono-sgen /path/to/tshock/TerrariaServer.exe -ip (ipv4here) -port 7777 -maxplayers 20 -world /path/to/world.wld -autocreate 0 -pass password`

Ill do some testing in the config file/OS environment and looking around for errors/issues. If i find anything ill post about it.

dly2424:
Interestingly, not specifiying the -world flag seems to rid of my desyncs, this is also using another world. Still looking into it.

Things that i changed around and havent affected it are:
Sudo permissions
Screen
Some Configurations

EDIT:
Further testing shows that -world flag doesnt matter, but instead the world seems to.
World with desyncs can be found below, using any pre versions so far up to pre 4 seems to cause desyncs.

A fresh world does not seem to have this issue.

[world.zip](https://github.com/Pryaxis/TShock/files/4657537/world.zip)

(All done on an untouched install of Pre 4 - Please attempt to replicate)

Patrikkk:
[...]

Issue #1800, opened 2020-05-20
Server Desync On Respawn 


TheCipher01:
* TShock version: Latest pre release
* TShock build number (if known):

#### Reproduction steps (if applicable)?

1) Die
2) Respawn and other players won't see you moving

I'm honestly not sure how to reproduce this because this issue just showed up less than an hour ago. One friend first noticed it earlier in the day but relogging fixed it and we didn't have the issue again until recently. Also after you die you can't seem to open chests till you relog. The only thing I could think of is we had a friend start playing golf, and then it seemed to cause more issues. No errors at all in the server console. It's running on a vps running debian. This is kind of based on #1799 Been watching the CPU usage all day  really haven't had anything above 50% on any one thread at any given time. We're at 2.07/5.82 G of RAM usage. 

#### Any stack traces or error messages (if known)?

```
PUT SUPER LONG ERROR MESSAGES IN THE TICK MARKS
```

#### Any screenshots?

 
Server Console has no errors. 

This is right after I died I was unable to pick up my money or anything and my friend didn't see me moving. 
https://gyazo.com/1fc4da7e976c69dde6961fff9b9b9c04

#### What plugins and what versions of those plugins are you running?
No plugins

Patrikkk:
This is resolved by AxeelAnder@53d211b 
Thank you for the reports guys! Helped us fix it.

TheCipher01:
Well I just died on pre4 and again was desynced. Is that fix not in pre4?

Patrikkk:
Sorry, my message wasn't clear enough. This will be fixed in the next pre-release which drops in a couple of hours!

Issue #1779, opened 2020-05-18
Big Desync Issue Between Players and Server


SnakeASaur:
* TShock version: 1.4
* TShock build number (if known):

#### Reproduction steps (if applicable)?

Just normal playing, building interacting with doors and chests

Once it happens, everyone else is moving around as normal on your screen, but to them you are frozen in one spot until you die at which point any progress made during that time is lost.

#### Any stack traces or error messages (if known)?

```
PUT SUPER LONG ERROR MESSAGES IN THE TICK MARKS
```

#### Any screenshots?
![image0](https://user-images.githubusercontent.com/65572887/82266708-46422f80-991f-11ea-951d-b21bdb48a165.jpg)

Don't have access to server console, using hosting provider

#### Any log messages from files that end in `.log` or `.txt`?

 - TShock: INFO: SnakeASaur (73.15.24.252) from 'default' group joined. (1/8)
2020-05-18 15:07:49 - Utils: INFO: Broadcast: SnakeASaur has joined.
2020-05-18 15:07:59 - GetDataHandlers: INFO: You authenticated successfully as user You.
2020-05-18 15:08:00 - TShock: INFO: You (67.164.41.107) from 'default' group joined. (2/8)
2020-05-18 15:08:00 - Utils: INFO: Broadcast: You has joined.
2020-05-18 15:08:08 - Utils: INFO: Broadcast: You has left.
2020-05-18 15:08:08 - TShock: INFO: You disconnected.
2020-05-18 15:08:24 - GetDataHandlers: INFO: You authenticated successfully as user You.
2020-05-18 15:08:24 - TShock: INFO: You (67.164.41.107) from 'default' group joined. (2/8)
2020-05-18 15:08:24 - Utils: INFO: Broadcast: You has joined.
2020-05-18 15:11:43 - Utils: INFO: SnakeASaur tried to execute /group add'.
2020-05-18 15:17:46 - Utils: INFO: Broadcast: Saving world. Momentary lag might result from this.
2020-05-18 15:18:04 - Utils: INFO: Broadcast: SnakeASaur has left.
2020-05-18 15:18:04 - TShock: INFO: SnakeASaur disconnected.
2020-05-18 15:18:06 - GetDataHandlers: INFO: SnakeASaur authenticated successfully as user SnakeASaur.
2020-05-18 15:18:07 - TShock: INFO: SnakeASaur (73.15.24.252) from 'default' group joined. (2/8)
[...]

PR #1812, opened 2020-05-20
Fix player spawn stuff

AxeelAnder:
This might fix the big desync bug from these issue: 
https://github.com/Pryaxis/TShock/issues/1779
https://github.com/Pryaxis/TShock/issues/1800
https://github.com/Pryaxis/TShock/issues/1809

I can't reproduce this bug in my local environment anyway, but these changes certainly will fix something

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 This is the rolling changelog for TShock for Terraria. Use past tense when adding new entries; sign your name off when you add or change something. This should primarily be things like user changes, not necessarily codebase changes unless it's really relevant or large.
 
 ## Upcoming changes
-* Update player spawn related things to 1.4. (@AxeelAnder)
+* Update player spawn related things to 1.4. `Terraria.Player.Spawn` method now has a required argument, `PlayerSpawnContext context`. (@AxeelAnder)
 
 ## TShock 4.4.0 (Pre-release 4)
 * Debug logging now provides ConsoleDebug and ILog has been updated to support the concept of debug logs. Debug logs are now controlled by `config.json` instead of by preprocessor debug flag. (@hakusaro)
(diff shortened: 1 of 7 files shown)


Patrikkk:
Indeed my man. Great find! 👍 
`Terraria.Player.Spawn` method now has a required argument, `PlayerSpawnContext`.  `TShock.Net.SpawnMsg` was not updated to pass spawncontext and respawntime. This resulted in clients reading invalid data which surely causes issues.
Not sure if it is related to #1779 though.
Since you are actively tending to this issue, can I ask you to open up two clients and try to replicate it, without your patch? Join with both clients, die with one of them and see if the desync happens when your player respawns.

Patrikkk:
#1813 Might be related to this as well.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 This is the rolling changelog for TShock for Terraria. Use past tense when adding new entries; sign your name off when you add or change something. This should primarily be things like user changes, not necessarily codebase changes unless it's really relevant or large.
 
+## Upcoming changes
+* Update player spawn related things to 1.4. (@AxeelAnder)
```
Patrikkk: Please add a more informative changelog, since changing .Spawn() is a breaking change on projects that reference TShock and access the method.
`Terraria.Player.Spawn method now has a required argument, PlayerSpawnContext.`

Patrikkk (changes requested):

CHANGELOG.md
AxeelAnder: ty, I'm not good in English so I'm mostly unable to do doc or instruction things well

Patrikkk: Don't put yourself down. :) Keep up the good work!

Patrikkk (approved):

hakusaro:
@AxeelAnder I have a rerpo case I can send you but it looks like you fixed it on our consistent test case. Wow. You're incredible! 💚

PR #1816, opened 2020-05-20
Missing backtick from RegionManager Group table.

Patrikkk:
Group is a reserved keyword in MySQL 8. Requires backtick.
Missed this from previous commit. Version 8 would error on table creation.


hakusaro:
Did you check to see if the server creates dB normally on SQLite with this?

Patrikkk:
I checked now. It does create the table, but it fails to execute INSERT query.

hakusaro:
What?

hakusaro (changes requested):
Needs to be made to work.

hakusaro (approved):

PR #1821, opened 2020-05-20
Remove mentions of a superadmin group during /setup

ColinBohn:
Small messaging changes around the /setup command to discourage use of the superadmin group.


Vednix:
Think that the superuser is the root of a server, you have several management members from different areas, would you root them all?

hakusaro:
@all-contributors add @ColinBohn for code, support

hakusaro:
@ColinBohn sorry, but I ninja'd you: https://github.com/Pryaxis/TShock/commit/f25ed62599cf78a9c1d95e620318e702c7ebc449

PR #1830, opened 2020-05-21
Fix player respawn again

AxeelAnder:
Thanks to test materials provided by @hakusaro. Confirmed the desync issue has been solved.

Terraria's respawn timer's unit is frame/tick while tshock's is second. So I correct the unit.

I noticed when set respawn time longer than normal respawn time in config.json, client will send a spawn packet to revive thus our config is of no use, should we forbid this or add a permission/config or do nothing?

### DIFF
--- a/TShockAPI/ConfigFile.cs
+++ b/TShockAPI/ConfigFile.cs
@@ -441,12 +441,12 @@ namespace TShockAPI
 		[Description("Whether or not to log REST API connections.")]
 		public bool LogRest = false;
 
-		/// <summary>The number of seconds a player must wait before being respawned.</summary>
-		[Description("The number of seconds a player must wait before being respawned.")]
+		/// <summary>The number of seconds a player must wait before being respawned. Cannot be longer than normal value now. Use at your own risk.</summary>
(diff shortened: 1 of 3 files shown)


QuiCM:
Can we swallow the client generated spawn packet and only respawn when we're ready to?

hakusaro:
I would just ignore the client packet and respawn when we want them to.

AxeelAnder:
I tried to prevent client from respawning but failed, there's no way to change client's respawn time

AxeelAnder:
If we ignore all spawn packets sent by client, client will be "false alive", aka the desync state

hakusaro (approved):
Honestly, probably worth merging

hakusaro:
@AxeelAnder I force pushed your fork of TShock to change your commit message to include the original description of your PR. Since I've been making heavy use of git I wanted to make sure that the reason why this change was made was encoded in the message. I would normally ask you to change it but I didn't want to bother you since it's kinda late.

AxeelAnder:
Even have tried to hack the mechanism, still failed. So I add a warning on config description.

PR #1879, opened 2020-05-24
Fix wind command

### LINKED ISSUE #1763: Wind command no longer functions
From Discord:
image

Wind command seems to set wind speed to 0 now. We should probably fix this

AxeelAnder:
Fixes https://github.com/Pryaxis/TShock/issues/1763

I'm sorry, this bug is my masterpiece :sob: , I simply swapped fields to change without looking into them.


hakusaro (approved):
You can merge this whenever @AxeelAnder?

PR #1880, opened 2020-05-24
Optimize workflow

sgkoishi:
Overall time halved. Not a big change so no changelog update :)
* Use official NuGet installer (choco takes two minutes to download NuGet)
* Add names for tasks
* Re-order tasks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
       uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: Installer NuGet client
+    - name: Install NuGet client
       uses: nuget/setup-nuget@v1
     - name: Restore NuGet packages
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,11 +10,6 @@ jobs:
       uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: Add MSBuild to environment variable
-      shell: pwsh
-      run: |
-        $msbuildPath = Split-Path (& "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -requires Microsoft.Component.MSBuild -find MSBuild\Current\Bin\amd64\MSBuild.exe | Select-Object -First 1) -Parent
-        echo "::set-env name=PATH::$msbuildPath;$env:PATH"
     - name: Installer NuGet client
       uses: nuget/setup-nuget@v1
     - name: Restore NuGet packages
@@ -24,19 +19,19 @@ jobs:
     - name: Build OTAPI (Debug)
       shell: cmd
       run: |
-        MSBuild.exe .\TerrariaServerAPI\TShock.4.OTAPI.sln /p:Configuration=Debug
+        "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\TerrariaServerAPI\TShock.4.OTAPI.sln /p:Configuration=Debug
         cd .\TerrariaServerAPI\TShock.Modifications.Bootstrapper\bin\Debug
         TShock.Modifications.Bootstrapper.exe
     - name: Build TerrariaServerAPI (Debug)
       shell: cmd
       run: |
         cd .\TerrariaServerAPI
-        MSBuild.exe .\TerrariaServerAPI\TerrariaServerAPI.csproj /p:Configuration=Debug
+        "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\TerrariaServerAPI\TerrariaServerAPI.csproj /p:Configuration=Debug
     - name: Build TShock (Debug)
       shell: cmd
       run: |
         nuget restore TShock.sln
-        MSBuild.exe .\TShockAPI\TShockAPI.csproj /p:Configuration=Debug
+        "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\TShockAPI\TShockAPI.csproj /p:Configuration=Debug
     - name: Prepare packaging
       shell: cmd
       run: |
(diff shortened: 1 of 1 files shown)


hakusaro (commented):
Do not merge without approval from me -- need time to review/consider.

.github/workflows/build.yml
```diff
@@ -6,67 +6,83 @@ jobs:
...
+        cd .\TerrariaServerAPI\TShock.Modifications.Bootstrapper\bin\Release
+        TShock.Modifications.Bootstrapper.exe
+    - name: Build TerrariaServerAPI (Release)
       shell: cmd
       run: |
         cd .\TerrariaServerAPI
-        "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\TerrariaServerAPI\TerrariaServerAPI.csproj /p:Configuration=Release
-    - name: TShock Release
+        MSBuild.exe .\TerrariaServerAPI\TerrariaServerAPI.csproj /p:Configuration=Release
+    - name: Build TShock (Release)
```
hakusaro: We intentionally don't ever say the word "release" in this process because we don't want to convey that it is a _release_. That's why we say "not debug" and "debug."

.github/workflows/build.yml
```diff
@@ -6,67 +6,83 @@ jobs:
...
+    - name: Git checkout
+      uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: Install nuget
-      run: choco install nuget.commandline
-    - name: OTAPI Debug
-      shell: cmd
+    - name: Add MSBuild to environment variable
+      shell: pwsh
```
hakusaro: I really don't want to use Powershell for this. Powershell isn't platform agnostic even when it tries to be. If we want to later switch to linux (which we might, I'd want to switch back).

.github/workflows/build.yml
```diff
@@ -6,67 +6,83 @@ jobs:
...
-    - name: TShock Release
+        MSBuild.exe .\TerrariaServerAPI\TerrariaServerAPI.csproj /p:Configuration=Release
+    - name: Build TShock (Release)
       shell: cmd
       run: |
         nuget restore TShock.sln
-        "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\TShockAPI\TShockAPI.csproj /p:Configuration=Release
-    - name: Normalize release packaging
+        MSBuild.exe .\TShockAPI\TShockAPI.csproj /p:Configuration=Release
+    - name: Prepare packaging
```
hakusaro: Why?

.github/workflows/build.yml
```diff
@@ -6,67 +6,83 @@ jobs:
...
       shell: cmd
       run: |
         xcopy /Y prebuilts\*.* TShockAPI\bin\Release
-        xcopy /Y prebuilts\*.* TShockAPI\bin\Debug
-        mkdir TShockAPI\bin\Debug\ServerPlugins
         mkdir TShockAPI\bin\Release\ServerPlugins
         move TShockAPI\bin\Release\TShockAPI.dll TShockAPI\bin\Release\ServerPlugins
-        move TShockAPI\bin\Debug\TShockAPI.dll TShockAPI\bin\Debug\ServerPlugins
-    - uses: actions/upload-artifact@master
+    - name: Upload TShock (Release)
```
hakusaro: Same on the "Release" verbiage.

.github/workflows/build.yml
```diff
@@ -6,67 +6,83 @@ jobs:
...
-    - uses: actions/upload-artifact@master
-      with:
-        name: Experimental TShock (debug)
-        path: TShockAPI\bin\Debug
-    - uses: actions/upload-artifact@master
-      with:
-        name: Experimental (debug) OTAPI Bootstrapper
-        path: .\TerrariaServerAPI\TShock.Modifications.Bootstrapper\bin\Debug\TShock.Modifications.Bootstrapper.exe
-    - uses: actions/upload-artifact@master
+    - name: Upload OTAPI Bootstrapper (Release)
```
hakusaro: Same thing here too

hakusaro (dismissed):
I think you could probably get by with changing it so we use the official nuget installer action and softcoding the path to VS, but I'm not convinced that reordering this or shifting to PowerShell for the build is necessary. I'm the one who has to maintain the build infra and while the current script is definitely not super optimized, I wouldn't say that it bothers me in the slightest and the organization is simple.

.github/workflows/build.yml
sgkoishi: I'm using Powershell here to put MSBuild's folder into `PATH` environment variable - so that we can replace the long `"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe"` with `MSBuild`. I can switch it back if you want.

.github/workflows/build.yml
sgkoishi: Get it, replaced with `Not Debug`

.github/workflows/build.yml
sgkoishi: Well I guess this is still not the *release*? ~~We need better text for this!!!~~

hakusaro: Technically we made this only to prepare for releases ;)

sgkoishi: Get it.

hakusaro (commented):
I think this is mostly good to go. I have been thinking about this. Do you have opinions about splitting into two workflow files (one for debug mode, one for release mode)? It might result in faster stuff since the job would complete faster?

sgkoishi:
> 
> 
> I think this is mostly good to go. I have been thinking about this. Do you have opinions about splitting into two workflow files (one for debug mode, one for release mode)? It might result in faster stuff since the job would complete faster?

Splitting them will restore packages twice, but it should still be faster (like two 80-seconds vs. one 120-seconds).
There are 15 jobs in the single workflow which is kinda long. I would prefer splitting them into two 9-jobs workflows (still not short, but more acceptable).

sgkoishi:
Hmmm... Using build matrix syntax instead of splitting into two files. 
There is still a `Release` in the name of run, but all artifacts are `not debug`. Is that acceptable?

hakusaro (changes requested):
This is looking great, but it looks like some artifacts aren't getting uploaded: 
Screen Shot 0002-05-30 at 14 28 28

sgkoishi:
> 
> 
> This is looking great, but it looks like some artifacts aren't getting uploaded:
> Screen Shot 0002-05-30 at 14 28 28

Aha yeah, because the build configurations are `Debug` and `Release` while artifacts' names are `debug` and `not debug`.

Currently, it's trying to upload 8 artifacts (`debug`/`not debug`, TShock/OTAPI, `Debug`/`Release`), and half is empty so finally 4 artifacts will be uploaded.

Splitting them into two workflow files will fix that (with much more duplicated code).


Wait, I got a nice fix

hakusaro (changes requested):
> `The workflow is not valid. .github/workflows/build.yml (Line: 44, Col: 11): Unexpected symbol: '"Debug"'. Located at position 16 within expression: matrix.mode == "Debug",.github/workflows/build.yml (Line: 50, Col: 11): Unexpected symbol: '"Debug"'. Located at position 16 within expression: matrix.mode == "Debug"`

Looks syntactically sad :(

sgkoishi:
No idea why they accept only `'Debug'` but no `"Debug"` :(

.github/workflows/build.yml
```diff
@@ -5,68 +5,62 @@ on: [push, pull_request]
...
-    - uses: actions/checkout@v1
+    - name: Git checkout
+      uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: Install nuget
-      run: choco install nuget.commandline
-    - name: OTAPI Debug
-      shell: cmd
+    - name: Installer NuGet client
```
hakusaro:
```suggestion
    - name: Install NuGet client
```

hakusaro (approved):
Last change before we can merge this!

hakusaro (approved):

PR #1902, opened 2020-05-24
Journeymode permissions

Patrikkk:
This is the proper way to tie TShock permissions to Journey mode creative powers.
I tried a different approach before this, where I use the vanilla permissions that are sent from server to client to limit their creative power usage.
This did not work for unknown reasons. The difference between the two approaches would have been that the first failed approach would not have been able to display messages to the player, as the block is happening in the client (buttons would not react at all unlike in this approach). And for this reason, it would have been an insecure approach, as client mods could have overwritten those client-sided permissions.
And yes... I over analyzed this simple implementation.  😅 

Please give me feedback with permission names and descriptions.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -3107,6 +3107,130 @@ namespace TShockAPI
 
 		private static bool HandleLoadNetModule(GetDataHandlerArgs args)
 		{
+			short moduleId = args.Data.ReadInt16();
+			if (moduleId == 6) // Power module.
+			{
+				short powerId = args.Data.ReadInt16();
+				switch (powerId)
+				{
+					case 0:
+						{
+							if (!args.Player.HasPermission(Permissions.journey_timefreeze))
+							{
+								args.Player.SendErrorMessage("You have no permission to freeze the time of the server!");
+								return true;
+							}
+							break;
+						}
+					case 1:
+					case 2:
+					case 3:
+					case 4:
+						{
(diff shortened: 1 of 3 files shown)


TShockAPI/GetDataHandlers.cs
```diff
@@ -3116,6 +3116,130 @@ private static bool HandlePlaceObject(GetDataHandlerArgs args)
...
+					case 0:
+						{
+							if (!args.Player.HasPermission(Permissions.journey_timefreeze))
+							{
+								args.Player.SendErrorMessage("You have no permission to freeze the time of the server!");
+								return true;
+							}
+							break;
+						}
+					case 1:
```
hakusaro: Can you use an enumeration for this so that these aren't magic numbers?

TShockAPI/GetDataHandlers.cs
```diff
@@ -3116,6 +3116,130 @@ private static bool HandlePlaceObject(GetDataHandlerArgs args)
 
 		private static bool HandleLoadNetModule(GetDataHandlerArgs args)
 		{
+			short moduleId = args.Data.ReadInt16();
+			if (moduleId == 6) // Power module.
```
hakusaro: Can you use an enumeration for this because different modules are identifiable?

hakusaro (changes requested):
Can you give these permissions to groups by default so that the suggested groups make sense? I've added a couple comments -- in particular, there are a lot of magic numbers that should be enums instead for long term maintenance.

hakusaro (approved):

PR #1940, opened 2020-05-28
Implement FoodPlatter placing event.

Patrikkk:
This is called when a player is placing a fruit (item) in a plate.
Adding checks to see if they have permission to place or replace food in the plate.
Checks if they are within range. And a check to see if they are legitimately placing the item from their hand, and not by sending a raw packet.

This PR does not fix the current issue, where removing food from a plate does not occur in item drop.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -67,6 +67,7 @@ namespace TShockAPI
 			GetDataHandlers.MassWireOperation += OnMassWireOperation;
 			GetDataHandlers.PlayerDamage += OnPlayerDamage;
 			GetDataHandlers.KillMe += OnKillMe;
+			GetDataHandlers.FoodPlatterTryPlacing += OnFoodPlatterTryPlacing;
 		}
 
 		internal void OnGetSection(object sender, GetDataHandlers.GetSectionEventArgs args)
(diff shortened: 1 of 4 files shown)


TShockAPI/TShockAPI.csproj
```diff
@@ -31,7 +31,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>..\..\..\Debug\ServerPlugins\</OutputPath>
```
Olink: I don't think this can change, right?
Looks like it broke the CI plan.

TShockAPI/Bouncer.cs
```diff
@@ -2062,6 +2063,54 @@ internal void OnKillMe(object sender, GetDataHandlers.KillMeEventArgs args)
...
 		}
 
+		/// <summary>
+		/// Called when a player is trying to place an item into a food plate.
+		/// </summary>
+		/// <param name="sender"></param>
+		/// <param name="args"></param>
+		internal void OnFoodPlatterTryPlacing(object sender, GetDataHandlers.FoodPlatterTryPlacingEventArgs args)
+		{
+			if (args.Player.ItemInHand.type != args.ItemID)
```
Olink, line 2073: Is it possible to place food from the hotbar, rather than using the item in hand?

TShockAPI/TShockAPI.csproj
Patrikkk: I'm sorry, I did not mean to push this file...

TShockAPI/Bouncer.cs
Patrikkk: I did think of that, even asked around people but the only way I could find was right-clicking when having it in hand.

Olink (approved):
LGTM 👍

PR #1941, opened 2020-05-28
Adding additional FoodPlatter event check. Modifying range check.

Patrikkk:
There are two ways to place food into a plate. One is by having it in hand (mouse) and right-clicking, the other is by having the item selected in the... "inventory bar"(?) and right-clicking the plate.

Tested range, if the player is outside the range, they should not get their item back.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2070,7 +2070,7 @@ namespace TShockAPI
 		/// <param name="args"></param>
 		internal void OnFoodPlatterTryPlacing(object sender, GetDataHandlers.FoodPlatterTryPlacingEventArgs args)
 		{
-			if (args.Player.ItemInHand.type != args.ItemID)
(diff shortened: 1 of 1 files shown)


hakusaro (changes requested):
Commits should be in present imperative tense and should update the changelog.

TShockAPI/Bouncer.cs
```diff
@@ -2099,12 +2099,9 @@ internal void OnFoodPlatterTryPlacing(object sender, GetDataHandlers.FoodPlatter
 				return;
 			}
 
-			if (!args.Player.IsInRange(args.TileX, args.TileY))
+			if (!args.Player.IsInRange(args.TileX, args.TileY, range: 10)) // To my knowledge, max legit tile reach with accessories.
```
Olink: May want to account for any lag in player position.

hakusaro (approved):

PR #1943, opened 2020-05-29
Fishoutnpc event

Patrikkk:
Add FishOutNPC event handler, which is called whenever a player fishes out an NPC using a fishing rod. Added anti-hack to Bouncer, to prevent unauthorized and invalid mob spawning, by checking player action, NPC IDs and range.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -67,6 +67,7 @@ namespace TShockAPI
 			GetDataHandlers.MassWireOperation += OnMassWireOperation;
 			GetDataHandlers.PlayerDamage += OnPlayerDamage;
 			GetDataHandlers.KillMe += OnKillMe;
+			GetDataHandlers.FishOutNPC += OnFishOutNPC;
 			GetDataHandlers.FoodPlatterTryPlacing += OnFoodPlatterTryPlacing;
 		}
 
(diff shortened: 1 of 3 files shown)


TShockAPI/GetDataHandlers.cs
```diff
@@ -1864,6 +1865,45 @@ private static bool OnKillMe(TSPlayer player, MemoryStream data, byte plr, byte
...
+			/// </summary>
+			public short NpcID { get; set; }
+		}
+		/// <summary>
+		/// Called when a player fishes out an NPC.
+		/// </summary>
+		public static HandlerList<FishOutNPCEventArgs> FishOutNPC = new HandlerList<FishOutNPCEventArgs>();
+		private static bool OnFishOutNPC(TSPlayer player, MemoryStream data, ushort tileX, ushort tileY, short npcID)
+		{
+			if (FoodPlatterTryPlacing == null)
```
Olink: This seems....incorrect.

TShockAPI/Bouncer.cs
```diff
@@ -2071,16 +2071,9 @@ internal void OnKillMe(object sender, GetDataHandlers.KillMeEventArgs args)
...
-				if (Main.projectile[recentProjectile.Index] != null && Main.projectile[recentProjectile.Index].Name == "Bobber")
-				{
-					projectile = Main.projectile[recentProjectile.Index];
-					break;
-				}
-			}
-			if (!FishingRodItemIDs.Contains(args.Player.SelectedItem.type) || projectile == null || !FishableNpcIDs.Contains(args.NpcID))
+			var projectile = args.Player.RecentlyCreatedProjectiles.FirstOrDefault(p => Main.projectile[p.Index] != null && Main.projectile[p.Index].Name == "Bobber");
+
+			if (!FishingRodItemIDs.Contains(args.Player.SelectedItem.type) || Main.projectile[projectile.Index] == null || !FishableNpcIDs.Contains(args.NpcID))
```
Olink, line 1895:
```csharp
if (!FishingRodItemIDs.Contains(args.Player.SelectedItem.type) || projectile == null || !FishableNpcIDs.Contains(args.NpcID))
```

IEnumerable.FirstOrDefault() will return null if no items in the collection match the predicate.  You've potentially dereferenced a null here.

hakusaro:
This now needs additional work due to conflicts in code files -- soz. Please rebase/merge when you have time.

TShockAPI/Bouncer.cs
Patrikkk: `var projectile` is a struct (`ProjectileStruct`) so it cannot be null. It would return a struct with default values.
I won't merge until you reply:)

Olink: Fair enough.
On that note though, I don't recall if I put a default constructor into the projectile struct...if I didn't the projectile.Index will default to 0, which happens to be a valid index.

Patrikkk: That is indeed true. That would mean that one check does nothing because it could never be null if I'm correct..

Patrikkk:
Resolved CHANGELOG conflict.

TShockAPI/GetDataHandlers.cs
```diff
@@ -1865,6 +1866,45 @@ private static bool OnKillMe(TSPlayer player, MemoryStream data, byte plr, byte
 			return args.Handled;
 		}
 
+		/// <summary>
+		/// For use in a FishOutNPC event.
+		/// </summary>
+		public class FishOutNPCEventArgs : GetDataHandledEventArgs
+		{
+			/// <summary>
+			/// The X world position of the spawning NPC.
```
hakusaro, line 1875: Is this on the tile grid or on the pixel grid?

hakusaro (approved):
lgtm on paper, re-requested review from Olink with one minor comment before merge

hakusaro:
for clarity if @QuiCM wants to step in for olink that's cool too

PR #1960, opened 2020-05-31
Fix doors in the lamest way possible (!)

### LINKED ISSUE #1774: TShock 4.4.0 Pre 4, Doors disappear
[Worlds 5-18-2020-11-52pm.zip](https://github.com/Pryaxis/TShock/files/4646291/Worlds.5-18-2020-11-52pm.zip)

* TShock version: 4.4.0 Pre 1
* TShock build number (if known): b56f1fc https://github.com/Pryaxis/TShock/releases/tag/v4.4.0-pre1

Happening to a user that is overseas so I'm assuming it's due to high latency but when opening or closing doors it will disappear and if there is a trap door next to it then it will duplicate that trap door. I haven't tested this on any other item to see if it duplicates it too. 

This doesn't happen to all door but only some of them too.

No error I can tell 

#### Any screenshots?

https://imgur.com/a/CjxSQGz

#### Any log messages from files that end in `.log` or `.txt`?

No error I can tell 

#### What plugins and what versions of those plugins are you running?

No Plugins

hakusaro:
Fixes #1774.

This commit is designed to fix the clientside door desync issue. Based
on the order of events that I've been able to see, the way that door
opening works is like this:

1. Client sends a door open request.
2. Server echoes request back to client.
3. Both server and client simulate door opening.
4. The client that requests the initial door open sends a tile square to
the server for some reason.

In TShock, under all circumstances, we send a tile square back to the
client that sends one in, unless you have the
`tshock.ignore.sendtilesquare` permission. Therefore, this deviates from
Terraria behavior where a tile square is broadcast to all other players.
Rather, the tile square is sent back to the player who requested it.

This does not really solve the underlying problem or answer the question
as to why sending a tile square back to the client seems to throw it
off, but it does. I was not able to replicate the desync issue anymore
with this branch. I expect that it will be safe to keep, because the
improved logic will only happen if the tile square had no effective
changes in addition to the door changes.

Video: https://files.catbox.moe/67rp9k.mov

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -560,6 +560,7 @@ namespace TShockAPI
 
 			bool changed = false;
 			bool failed = false;
+			bool doorRelated = false;
 			try
 			{
 				var tiles = new NetTile[size, size];
@@ -589,6 +590,11 @@ namespace TShockAPI
 							!args.Player.IsInRange(realx, realy))
 						{
 							continue;
+						}
+
+						if (newtile.Active && Terraria.ID.TileID.Sets.RoomNeeds.CountsAsDoor.Contains(newtile.Type))
+						{
+							doorRelated = true;
 						}
 
 						// Fixes the Flower Boots not creating flowers issue
@@ -704,7 +710,8 @@ namespace TShockAPI
 				}
 				else
 				{
-					args.Player.SendTileSquare(tileX, tileY, size);
+					if (!doorRelated)
+						args.Player.SendTileSquare(tileX, tileY, size);
 				}
 			}
 			catch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * The default group that gets this permission is `Guest` for the time being.
   * To add this command to your guest group, give them `tshock.synclocalarea`, with `/group addperm guest tshock.synclocalarea`.
   * This command may be removed at any time in the future (and will likely be removed when send tile square handling is fixed).
+* Fixed smart door automatic door desync and deletion issue. (@hakusaro)
 
 ## TShock 4.4.0 (Pre-release 8)
 * Update for OTAPI 2.0.0.36 and Terraria 1.4.0.4. (@hakusaro, @Patrikkk, @DeathCradle)


QuiCM:
Does this still set `args.Handled = true` on the handler?

hakusaro:
@QuiCM yes, the control loop is essentially the same except we don't network back to the client in this one case. The server data is accurate at the point that we decide not to send data to the player, which means that they won't be desynchronized.

hakusaro:
Honestly it's probably possible to just handle the event and return here, instead of sending any data to anyone. The server state is correct at this point but sending something to the client breaks it.

QuiCM:
> 
> 
> Honestly it's probably possible to just handle the event and return here, instead of sending any data to anyone. The server state is correct at this point but sending something to the client breaks it.

The vanilla process here is to send the tile square to other players, so I think it makes sense to echo it back to everyone. Should test that it doesn't make doors disappear on other clients though

TShockAPI/Bouncer.cs
```diff
@@ -704,7 +710,14 @@ internal void OnSendTileSquare(object sender, GetDataHandlers.SendTileSquareEven
 				}
 				else
 				{
-					args.Player.SendTileSquare(tileX, tileY, size);
+					if (doorRelated)
+					{
+						NetMessage.SendData((int) PacketTypes.TileSendSquare, -1, args.Player.TPlayer.whoAmI, NetworkText.FromLiteral(""), size, tileX, tileY);
```
QuiCM: Instead of going directly through netmessage, can we use `TSPlayer.All.SendTileSquare(...)`?

hakusaro:
@QuiCM as demonstrated in my video, it definitely works. Can't use `TSPlayer.All` because I'm specifically using the `ignoreClient` feature to ignore the existing client in that particular iteration.

I tested with two clients on the version I've just pushed (which doesn't send any data at all) and the data still syncs.

TShockAPI/Bouncer.cs
hakusaro: While this has been validated to work, we can't do that because `TSPlayer.All` sends to all players, and we only want to send to all players that _aren't_ the current player. In any case, the existing commit on the tip of this branch removes this branch entirely, so it's kinda moot unless you want to restore it. However, sending data to the current client doesn't work for sure. That's why I'm using `ignoreClient` to ignore the sender.

hakusaro:
This video demonstrates that the second iteration (which doesn't send _any_ data on a STS following door change) still synchronizes state properly. Packet 20, as shown, does not increment on the client except sporadically (not correlated with all door events, but other events): https://files.catbox.moe/y13x2b.mov

hakusaro:
I will admit, it does look a little more choppy I think on the receiving client when it doesn't get regular STS updates. It looks kind of like the player that's opening the door is "snapping" to one or the other side of the door. This could be client correction/interpolation of some kind (in Terraria?!) or it could just be the usual network stack issues.

QuiCM:
I'm inclined to say that even if it is a little choppier, its probably worth it given that we don't sync back a whole 5x5 tile square every time a door is opened or closed. That's not a cheap network operation and if there's no severe negatives from not sending it then cutting it out is a good deal

QuiCM (approved):
Code looks good

QuiCM:
As a note: I think this flow should also be implemented in the fix-sts branch as the fix for doors, as I believe the split of STS handling into its own class is worthwhile

hakusaro:
@QuiCM I admittedly haven't really looked into `fix-sts` on a cursory level lately. I think we should definitely keep it in some form in that branch. I tried to distance myself from that branch when working this problem so as to not skip over details (I re-learned STS from scratch a bit).

I'll merge after some prep.

PR #1961, opened 2020-05-31
Add door debug visualizer to Bouncer

hakusaro:
This is my own take on a STS debug visualizer. It pads and prints things
nicely so that it looks like the game representation and might be
helpful for other people. Obviously this is like very low priority, but
I wanted to save it off in the event that someone other than me likes
it.

It looks like this:
Screen Shot 0002-05-31 at 00 00 24

It shows the current server state as well as the new server state after the change and highlights doors as open spaces with curly braces indicating that the door is "active."

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -526,6 +526,88 @@ namespace TShockAPI
 			int tileX = args.TileX;
 			int tileY = args.TileY;
 
+			int startX = tileX;
+			int startY = tileY;
+
+			var tiles = new NetTile[size, size];
+			for (int x = 0; x < size; x++)
+			{
+				for (int y = 0; y < size; y++)
+				{
+					tiles[x, y] = new NetTile(args.Data);
+				}
+			}
+
+			if (TShock.Config.DebugLogs)
+			{
+				char pad = '0';
+				for (int y = 0; y < size; y++)
+				{
+					int realY = y + startY;
+					for (int x = 0; x < size; x++)
+					{
+						int realX = x + startX;
+						ushort type = Main.tile[realX, realY].type;
+						string type2;
+						switch (type)
+						{
+							case 10:
+								type2 = "[X]";
+								break;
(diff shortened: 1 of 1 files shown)


QuiCM:
I like this, but I'm not sure I'm a fan of the symbols marking the door.
Not sure what `[ ]` or `{ }` represents

hakusaro:
@QuiCM it's supposed to be an "open gap" e.g., `[ ]` is an open door, and `{ }` is an open active door and `[x]` is a "filled in" door (a closed door). I can remove those if you want though, if doors don't need debugging anymore.

QuiCM:
> 
> 
> @QuiCM it's supposed to be an "open gap" e.g., `[ ]` is an open door, and `{ }` is an open active door and `[x]` is a "filled in" door (a closed door). I can remove those if you want though, if doors don't need debugging anymore.

for clarity my preference would be to remove it or have a legend explaining what special symbols mean

hakusaro:
> for clarity my preference would be to remove it or have a legend explaining what special symbols mean

Ya, if you're at all interested I can change it to not do that. I'm not sure if it would help with other STS debugging things though, because it doesn't show all of the different attributes on a square's state changes.

hakusaro:
(Maybe better suited to packet monitor honestly)

hakusaro:
Sorry to triple post, but another easy thing could just be to not handle the event and then return without sending anything either. I did at least go through and deduce that the STS event isn't worth messing with (it changes nothing about the state of the world), and flag it off so that it's not going to enter a fall through state if a user were to send a door change to bypass all other STS checks.

hakusaro:
@QuiCM it's kinda still messy but I think it's still novel (iterates y before x, writes in nice lines, pads for fixed widths)

QuiCM (approved):

QuiCM:
I pulled a sneaky and pushed this in with #1965

hakusaro:
@QuiCM :D

PR #1962, opened 2020-05-31
Projectilestruct addition

Patrikkk:
Can we store the type of recent projectile as well? This could be used in my upcoming PR regarding the golf packet where I check for recently created projectiles, and their types. Having only the index is not useful in this scenario, because the projectile can already be killed, so cannot access the Main.projectile array to get the projectile type.

Can this be public instead of internal? Developers could make good use of it, instead of having to implement their own version of RecentProjectiles in their plugins.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2398,6 +2398,7 @@ namespace TShockAPI
 					args.Player.RecentlyCreatedProjectiles.Add(new GetDataHandlers.ProjectileStruct()
 					{
 						Index = index,
+						Type = type,
 						CreatedAt = DateTime.Now
 					});
 				}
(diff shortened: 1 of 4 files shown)


TShockAPI/GetDataHandlers.cs
```diff
@@ -3770,9 +3771,10 @@ public enum EditType
 			{TileID.MinecartTrack, 3}
 		};
 
-		internal struct ProjectileStruct
+		public struct ProjectileStruct
```
hakusaro, line 3876: When changing access type to public, can you add docs? Prefer to have less compiler warnings than start on any PR.

hakusaro (dismissed):

TShockAPI/GetDataHandlers.cs
QuiCM: I have added some brief documentation for the purposes of merging this tonight - a quick post-merge review by @Patrikkk and @hakusaro to confirm the doco is accurate would be appreciated

QuiCM (approved):

TShockAPI/GetDataHandlers.cs
Patrikkk: Thank you<3 Looks good to me

PR #1965, opened 2020-06-01
Fix sts

QuiCM:
This PR introduces a new class for handling the SendTileSquare packet, pulling the complex logic out of Bouncer.
Additionally it should fix issues relating to:
* Door syncing
* Tile entity syncing
* Desyncs due to malformed tile squares being sent to client

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,11 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * The default group that gets this permission is `Guest` for the time being.
   * To add this command to your guest group, give them `tshock.synclocalarea`, with `/group addperm guest tshock.synclocalarea`.
   * This command may be removed at any time in the future (and will likely be removed when send tile square handling is fixed).
-* Fixed smart door automatic door desync and deletion issue. (@hakusaro)
+* Fixed smart door automatic door desync and deletion issue. (@hakusaro, @QuiCM, @Olink)
 * Bump default max damage received cap to 42,000 to accommodate the Empress of Light's instant kill death amount. (@hakusaro, @moisterrific, @Irethia, @Ayrawei)
+* Fixed a longstanding issue with SendTileSquare that could result in desyncs and visual errors (@QuiCM)
+* Fixed placement issues with Item Frames, Teleportation Pylons, etc (@QuiCM)
+
 
 ## TShock 4.4.0 (Pre-release 8)
 * Update for OTAPI 2.0.0.36 and Terraria 1.4.0.4. (@hakusaro, @Patrikkk, @DeathCradle)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 This is the rolling changelog for TShock for Terraria. Use past tense when adding new entries; sign your name off when you add or change something. This should primarily be things like user changes, not necessarily codebase changes unless it's really relevant or large.
 
 ## Upcoming release
-* Your change goes here!
+* Fixed a longstanding issue with SendTileSquare that could result in desyncs and visual errors (@QuiCM)
+* Fixed placement issues with Item Frames, Teleportation Pylons, etc (@QuiCM)
+* Doors are good now for real probably (@QuiCM, @Hakusaro, @Olink)
 
 ## TShock 4.4.0 (Pre-release 9)
 * Fixed pet licenses. (@Olink)
@@ -31,10 +33,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * The default group that gets this permission is `Guest` for the time being.
   * To add this command to your guest group, give them `tshock.synclocalarea`, with `/group addperm guest tshock.synclocalarea`.
   * This command may be removed at any time in the future (and will likely be removed when send tile square handling is fixed).
-* Fixed smart door automatic door desync and deletion issue. (@hakusaro, @QuiCM, @Olink)
+* Fixed smart door automatic door desync and deletion issue. (@hakusaro)
 * Bump default max damage received cap to 42,000 to accommodate the Empress of Light's instant kill death amount. (@hakusaro, @moisterrific, @Irethia, @Ayrawei)
-* Fixed a longstanding issue with SendTileSquare that could result in desyncs and visual errors (@QuiCM)
-* Fixed placement issues with Item Frames, Teleportation Pylons, etc (@QuiCM)
+
 
 
 ## TShock 4.4.0 (Pre-release 8)
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -43,12 +43,12 @@ namespace TShockAPI
 		internal Bouncer()
 		{
 			STSHandler = new Handlers.SendTileSquareHandler();
+			GetDataHandlers.SendTileSquare += STSHandler.OnReceiveSendTileSquare;
 
 			// Setup hooks
 			GetDataHandlers.GetSection += OnGetSection;
 			GetDataHandlers.PlayerUpdate += OnPlayerUpdate;
 			GetDataHandlers.TileEdit += OnTileEdit;
-			GetDataHandlers.SendTileSquare += STSHandler.OnReceiveSendTileSquare;
 			GetDataHandlers.ItemDrop += OnItemDrop;
 			GetDataHandlers.NewProjectile += OnNewProjectile;
 			GetDataHandlers.NPCStrike += OnNPCStrike;
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -36,15 +36,19 @@ namespace TShockAPI
 	/// <summary>Bouncer is the TShock anti-hack and anti-cheat system.</summary>
 	internal sealed class Bouncer
 	{
+		internal Handlers.SendTileSquareHandler STSHandler { get; set; }
+
 		/// <summary>Constructor call initializes Bouncer and related functionality.</summary>
 		/// <returns>A new Bouncer.</returns>
 		internal Bouncer()
 		{
+			STSHandler = new Handlers.SendTileSquareHandler();
+
 			// Setup hooks
 			GetDataHandlers.GetSection += OnGetSection;
 			GetDataHandlers.PlayerUpdate += OnPlayerUpdate;
 			GetDataHandlers.TileEdit += OnTileEdit;
-			GetDataHandlers.SendTileSquare += OnSendTileSquare;
+			GetDataHandlers.SendTileSquare += STSHandler.OnReceiveSendTileSquare;
 			GetDataHandlers.ItemDrop += OnItemDrop;
 			GetDataHandlers.NewProjectile += OnNewProjectile;
 			GetDataHandlers.NPCStrike += OnNPCStrike;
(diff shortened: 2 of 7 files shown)


QuiCM:
Quick note: this doesn't fix trapdoors. They're even more broken than doors

CHANGELOG.md
```diff
@@ -31,8 +31,11 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * The default group that gets this permission is `Guest` for the time being.
   * To add this command to your guest group, give them `tshock.synclocalarea`, with `/group addperm guest tshock.synclocalarea`.
   * This command may be removed at any time in the future (and will likely be removed when send tile square handling is fixed).
-* Fixed smart door automatic door desync and deletion issue. (@hakusaro)
+* Fixed smart door automatic door desync and deletion issue. (@hakusaro, @QuiCM, @Olink)
 * Bump default max damage received cap to 42,000 to accommodate the Empress of Light's instant kill death amount. (@hakusaro, @moisterrific, @Irethia, @Ayrawei)
+* Fixed a longstanding issue with SendTileSquare that could result in desyncs and visual errors (@QuiCM)
```
hakusaro: This is the changelog entry for an already released version. You should put your changes in the next block (upcoming changes).

TShockAPI/Bouncer.cs
```diff
@@ -36,15 +36,19 @@ namespace TShockAPI
 	/// <summary>Bouncer is the TShock anti-hack and anti-cheat system.</summary>
 	internal sealed class Bouncer
 	{
+		internal Handlers.SendTileSquareHandler STSHandler { get; set; }
+
 		/// <summary>Constructor call initializes Bouncer and related functionality.</summary>
 		/// <returns>A new Bouncer.</returns>
 		internal Bouncer()
 		{
+			STSHandler = new Handlers.SendTileSquareHandler();
```
hakusaro, line 45: While Bouncer can handle this, I think it's a bit confusing that `SendTileSquareHandler` lives in `Handlers` and not some sort of `Bouncer` equivalent namespace. If it isn't in like a `Bouncer` namespace, it should probably not be initialized by `Bouncer`. I think the operation is a bit nebulous so it probably needs to go in Bouncer though.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+				TileEntity.PlaceEntityNet(realx, realy, _tileEntityIdToTileIdMap[tileType]);
+			}
+		}
+
+		/// <summary>
+		/// Invoked when a SendTileSquare packet is received
+		/// </summary>
+		/// <param name="sender"></param>
+		/// <param name="args"></param>
+		public void OnReceiveSendTileSquare(object sender, GetDataHandlers.SendTileSquareEventArgs args)
```
hakusaro: Since this is the "entrypoint" of this object, (i.e., it is the thing that is being called from an event handler and where all processing begins), it should be the first method in the class. This means that someone coming in to debug issues with it (or just to look at it) will start at the top of the file and things will logically flow down.

In general, it's easier to understand things with a single entrypoint if the file structure looks like this:

1. The constructor comes first.
2. The most important method or most important thing in the class comes next.
3. Helper methods that help accomplish main goal that are procedural come next in order of appearance.

This means that a hypothetical signature structure would look like:
```
void OnReceiveSendTileSquare(...);
void ProcessTileObject(...);
void UpdateTile(...);
void ProcessSingleTile(...);
void HandleFlowerBoots(...);
void HandleConversionSpreads(...);
```

In general this means that the file can, for the most part, be read from top to bottom with minimal backtracking to follow the whole control flow. For heavily procedural code, this means that a reader can quickly understand the structure as it goes on.

Obviously this is not required for any type of actual code correctness, but is more of a style preference.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+		{
+			short size = args.Size;
+			int tileX = args.TileX;
+			int tileY = args.TileY;
+
+			// By default, we'll handle everything
+			args.Handled = true;
```
hakusaro: I feel like there is some merit to making this be explicit at every return path. The default way we handle things in TShock is a "blacklist approach" where we're only handling a set number of things. STS is the reverse. While we can keep this principle, it may be easier to follow long into the future if we push this down into every return block so that it keeps with the existing "convention" of changing the handled state and returning at the same time.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+			// At this point we should send our state back to the client so they remain in sync with the server
+			if (args.Handled == true)
+			{
+				args.Player.SendTileSquare(tileX, tileY, size);
+				TShock.Log.ConsoleDebug("Bouncer / SendTileSquare reimplemented from spaghetti from {0}", args.Player.Name);
+			}
+		}
+
+
+		class Debug
```
hakusaro, line 441: I know C# permits this, but I hate embedding things like this. The overall structure of the debug class is good. It's just surprising to me every time I see a class halfway down a file (line 373).

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+			{
+				for (int y = 0; y < size; y++)
+				{
+					// Do not process already processed tiles
+					if (processed[x, y])
+					{
+						continue;
+					}
+
+					int realx = tileX + x;
```
hakusaro: Should these be `realX` and `realY`?

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+			{ TileID.LogicSensor,        TELogicSensor._myEntityID },
+			{ TileID.DisplayDoll,        TEDisplayDoll._myEntityID },
+			{ TileID.WeaponsRack2,       TEWeaponsRack._myEntityID },
+			{ TileID.HatRack,            TEHatRack._myEntityID },
+			{ TileID.FoodPlatter,        TEFoodPlatter._myEntityID },
+			{ TileID.TeleportationPylon, TETeleportationPylon._myEntityID }
+		};
+
+		/// <summary>
+		/// Syncs a single tile on the server with a tile from the tile square packet
```
hakusaro:
```suggestion
		/// Syncs a single tile's canonical state on the server with a change from the tile square packet
```

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+		/// </summary>
+		/// <param name="tile">The tile to update</param>
+		/// <param name="newTile">The NetTile containing new tile properties</param>
+		public static void UpdateTile(ITile tile, NetTile newTile)
```
hakusaro: I think if you called this `UpdateServerState` or something to convey that it's a "destructive" action (i.e., it is changing the core data store for the world state), that would be useful in terms of communicating to future readers what precisely it's doing.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+		public void OnReceiveSendTileSquare(object sender, GetDataHandlers.SendTileSquareEventArgs args)
+		{
+			short size = args.Size;
+			int tileX = args.TileX;
+			int tileY = args.TileY;
+
+			// By default, we'll handle everything
+			args.Handled = true;
+
+			if (args.Player.HasPermission(Permissions.allowclientsideworldedit))
```
hakusaro, line 410: If we're going to keep this code super procedural, you could break the if statements here into small helper methods for each one. A hyper modular version could be like:
```
SkipProcessingForPrivilegedUsers();
SkipNonVanillaEdit();
SkipThrottledEdit();
SkipDisabledEdit();
```

The result of doing this is that the method names would encapsulate the reason and intent of the check. The checks are pretty minimal and uninteresting, so they can live somewhere else. Overall, doing this minimizes the number of lines between the "entrypoint" of the method, variable initialization, and then the core processing. This means that a future reader can quickly jump to the "interesting" part and not have to muddle through trivial checks.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+
+			if (args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug("Bouncer / SendTileSquare rejected from being disabled from {0}", args.Player.Name);
+				args.Player.SendTileSquare(tileX, tileY, size);
+				return;
+			}
+
+			bool[,] processed = new bool[size, size];
+			NetTile[,] tiles = new NetTile[size, size];
```
hakusaro: The upshot to making this a non-local variable would be that you could convert the for loop below into some helper method that is just like `DeserializeSendTileSquarePacket()`.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+			}
+
+			if (args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug("Bouncer / SendTileSquare rejected from being disabled from {0}", args.Player.Name);
+				args.Player.SendTileSquare(tileX, tileY, size);
+				return;
+			}
+
+			bool[,] processed = new bool[size, size];
```
hakusaro: It might be worth renaming this to something like `synchronized` or `persisted` -- "processed" is a really nebulous term. The `Process` verbiage is kinda nebulous too -- I tried to read this from scratch and understand it, and it with no preconceptions, and it threw me off. It's okay if there's no better thing to call this, but if you want to make an adjustment now, this would be an opportune time to.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+					{
+						continue;
+					}
+
+					NetTile newTile = tiles[x, y];
+					TileObjectData data;
+
+					// If the new tile has an associated TileObjectData object, we take the tile and the surrounding tiles that make up the tile object
+					// and process them as a tile object
+					if (newTile.Type < TileObjectData._data.Count && (data = TileObjectData._data[newTile.Type]) != null)
```
hakusaro: Maybe I'm just really old fashioned but I really can't stand assigning things in conditionals because I always think it's an error and then realize it's not. I never read this correctly on the first time through.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+					{
+						NetTile[,] newTiles = new NetTile[data.Width, data.Height];
+						for (int i = 0; i < data.Width; i++)
```
hakusaro: Given that we're iterating over a 2d array that represents an x, y plane, it might be better to use locals that reflect that this is an x, y plane. `offsetX` and `offsetY`?

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+						{
+							for (int j = 0; j < data.Height; j++)
+							{
+								newTiles[i, j] = tiles[x + i, y + j];
+								processed[x + i, y + j] = true;
+							}
+						}
+
+						ProcessTileObject(newTile.Type, data, newTiles, realx, realy, args);
```
hakusaro: I'm really thankful that this is separated into a helper method.

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+			// Uncommenting this function will send the same tile square 10 blocks above you for visualisation. This will modify your world and overwrite existing blocks.
+			// Use in test worlds only.
+			//Debug.DisplayTileSetInGame(tileX, tileY - 10, size, size, tiles, args.Player);
+
+			// If we are handling this event then we have updated the server's Main.tile state the way we want it.
+			// At this point we should send our state back to the client so they remain in sync with the server
+			if (args.Handled == true)
+			{
+				args.Player.SendTileSquare(tileX, tileY, size);
+				TShock.Log.ConsoleDebug("Bouncer / SendTileSquare reimplemented from spaghetti from {0}", args.Player.Name);
```
hakusaro:
```suggestion
				TShock.Log.ConsoleDebug("Bouncer / SendTileSquare reimplemented from new greatness from {0}", args.Player.Name);
```

Since the original comment was named after entirely inscrutable spaghetti, this should be different now!

TShockAPI/TSPlayer.cs
```diff
@@ -679,6 +680,31 @@ public bool HasBuildPermission(int x, int y, bool shouldWarnPlayer = true)
 			return false;
 		}
 
+		/// <summary>
+		/// Determines if the player can build a tile object on a given point.
```
hakusaro: I think there's a good % chance people don't know what a tile object is off the top of their head. If you can remind them here, that would be good.

QuiCM:
Trapdoors now fixed as well

TShockAPI/TShockAPI.csproj
```diff
@@ -211,7 +212,7 @@
   </PropertyGroup>
   <ProjectExtensions>
     <VisualStudio>
-      <UserProperties BuildVersion_UpdateAssemblyVersion="True" BuildVersion_UpdateFileVersion="True" BuildVersion_BuildAction="Both" BuildVersion_BuildVersioningStyle="None.None.None.MonthAndDayStamp" BuildVersion_StartDate="2011/6/17" BuildVersion_IncrementBeforeBuild="False" />
+      <UserProperties BuildVersion_IncrementBeforeBuild="False" BuildVersion_StartDate="2011/6/17" BuildVersion_BuildVersioningStyle="None.None.None.MonthAndDayStamp" BuildVersion_BuildAction="Both" BuildVersion_UpdateFileVersion="True" BuildVersion_UpdateAssemblyVersion="True" />
```
hakusaro, line 215: Eh?

TShockAPI/Handlers/SendTileSquareHandler.cs
```diff
@@ -0,0 +1,445 @@
...
+{
+	/// <summary>
+	/// Provides processors and handling for Tile Square packets
+	/// </summary>
+	public class SendTileSquareHandler
+	{
+		/// <summary>
+		/// Maps grass-type blocks to flowers that can be grown on them with flower boots
+		/// </summary>
+		Dictionary<ushort, List<ushort>> _grassToPlantMap = new Dictionary<ushort, List<ushort>>
```
hakusaro, line 22: I'm a big fan of shoving as much of this stuff in the bottom as possible and marking it as constant if it is, just because it's usually not that interesting. Granted, I think this is a deviation from the norm and most people disagree with me here, but eh.

hakusaro (approved):
I've made a number of minor suggestions, most of which are stylistic. This type of review is atypical for TShock, but I think it was worth doing because this could be a precedent setting file for other developers on how to properly encapsulate complex behavior in a way that's easy to understand. I approached this as open minded as possible and tried to view it like an outsider would and tried to note as many of the things that might confuse people.

The most notable thing I'd suggest and advocate for is reordering methods in order of importance, starting with the "entrypoint," and then working in order of appearance down. I'd also suggest renaming some of the `Process` named methods to something more powerful, with `UpdateTile` being renamed to something that indicates that it's dangerously updating the Terraria world state as it runs.

Less important things are like extracting out methods into smaller blocks. This would help in overall readability of the main entrypoint but can be complex to do.

Ultimately, if you're down for it, I would suggest trying to make this as non-procedural as possible and as self contained as possible so as to be an example we can give other people in the future.

But this is a volunteer project, and the code _works_, so I don't really think anything blocks merging here. If you've had enough of STS, feel free to just get this in ASAP.

Thank you so much for working on this for so long. Even if you change absolutely nothing (which you have every god given right to), I really appreciate this and understand that it's a huge step forward in terms of understanding of `SendTileSquare` in Terraria. Thank you for putting in the time and effort and dedication to understand this on a really fundamental level. Thank you for also determining that our `SendTileSquare` method in `TSPlayer` was totally bonkers and destroyed all doors for ages. 💚

CHANGELOG.md
QuiCM: oh i can't read

TShockAPI/TShockAPI.csproj
QuiCM: Visual Studio and project files do strange things. I think this slipped in when I merged gen-dev into STS

TShockAPI/Handlers/SendTileSquareHandler.cs
QuiCM: Given we have a number of non-native English speakers my preference is to avoid words like canonical.
`Syncs a single tile's state on the server with a change from the tile square packet` I think flows smoothly and encompasses the best of both worlds

hakusaro: You could change it to `world state`. "on disk" or "in memory" or anything is better than "on the server" because, frankly, everything is on the server.

TShockAPI/Handlers/SendTileSquareHandler.cs
QuiCM: At the moment, in this implementation, there is only one spot where Handled becomes false. I'm inclined to agree that it differs from our general bouncer handling where we explicitly set true before returning, but that also comes down to the flow of the handler. In this particular handler we are whitelisting rather than blacklisting so our default state is to handle the packet

hakusaro: Ya, fair. I think that "we'll handle everything" is probably good enough here as a defense there

TShockAPI/Handlers/SendTileSquareHandler.cs
hakusaro: (This isn't something you have to change, I just hate C#)

TShockAPI/Handlers/SendTileSquareHandler.cs
QuiCM: I did consider this but I don't really see a clean way to implement, as those methods would still need to return a value to break the control flow of the method. You'd end up with
```csharp
if (SkipProcessingForPrivilegedUsers())
{
    ...
    return;
}
```
Which doesn't add any clarity imo

TShockAPI/Handlers/SendTileSquareHandler.cs
QuiCM: It's certainly not something I'm a fan of, but it's a very clean way of moving debug method signatures somewhere else where they don't conflict with non-debug methods without dumping them into new classes

TShockAPI/Handlers/SendTileSquareHandler.cs
hakusaro: If you made it a oneshot, it would be
```
if (SkipProcessingGarbo())
    return;
```

TShockAPI/Bouncer.cs
QuiCM: I agree. I'm not a fan of pushing this into GetDataHandlers either though. I did consider making SendTileSquareHandler a static so that instantiation isn't explicit and it doesn't need to live anywhere much, but not sure

TShockAPI/Handlers/SendTileSquareHandler.cs
QuiCM: My personal preference is still 'processed' - simply because we read every tile in the tile square, but we don't necessarily do anything to them. We process them and action them only if required.
Persisted or synchronized in that case becomes counter-intuitive to me

QuiCM: The mixed use of `HandleXYZ` and `ProcessXYZ` doesn't help

hakusaro (approved):
fly 🚀 :shipit:

hakusaro:
lgtm_6

I forgot this

PR #1968, opened 2020-06-01
Handlegolfpacket

Patrikkk:
Added LandGolfBallInCup event which is accessible for developers to work with, as well as LandGolfBallInCup handler to handle exploits where players could send direct packets to trigger and imitate golf ball cup landing anywhere in the game world. Added two public lists in Handlers.LandGolfBallInCupHandler: GolfBallProjectileIDs and GolfClubItemIDs.

### DIFF
--- a/TShockAPI/Handlers/LandGolfBallInCupHandler.cs
+++ b/TShockAPI/Handlers/LandGolfBallInCupHandler.cs
@@ -1,0 +1,115 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Terraria;
+using Terraria.ID;
+
+namespace TShockAPI.Handlers
+{
+	/// <summary>
+	/// Handles client side exploits of LandGolfBallInCup packet.
+	/// </summary>
+	public class LandGolfBallInCupHandler
+	{
+		/// <summary>
+		/// List of golf ball projectile IDs.
+		/// </summary>
+		public static readonly List<int> GolfBallProjectileIDs = new List<int>()
+		{
+			ProjectileID.DirtGolfBall,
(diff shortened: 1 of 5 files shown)


TShockAPI/Handlers/LandGolfBallInCupHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+		/// <summary>
+		/// Invoked when a player lands a golf ball in a cup.
+		/// </summary>
+		/// <param name="sender"></param>
+		/// <param name="args"></param>
+		public void OnLandGolfBallInCup(object sender, GetDataHandlers.LandGolfBallInCupEventArgs args)
+		{
+			if (args.PlayerIndex != args.Player.Index)
+			{
+				TShock.Log.ConsoleError($"LandGolfBallInCupHandler: Packet is spoofing to be player ID {args.PlayerIndex}! - From [{args.Player.Index}]{args.Player.Name}");
```
Patrikkk: Should these be ConsoleDebug instead?

QuiCM: Yep.
You can also be less formal with ConsoleDebug messages because the intention is that we ask people to send specific sections of their log to capture specific messages.
I'd suggest a rephrasing to something like `LandGolfBallInCup rejected for ID spoofing. Expected {playerId} , received {actualId} from {player}`

QuiCM (changes requested):

TShockAPI/Handlers/LandGolfBallInCupHandler.cs
Patrikkk: Sounds, good  thanks! I sometimes have a hard time with wording when I'm coding. Will send a commit soon.

QuiCM (approved):

PR #1969, opened 2020-06-01
Emojipacketevent

Patrikkk:
This is received from the client when they are trying to display an emoji, this comes into the server, gets processed, and the server sends back the Emote Bubble packet to the clients.

I know, this is not something important, but I'm going through the new packets one by one and adding events developers can work with, patching exploits, and thought this could be a core permission.

### DIFF
--- a/TShockAPI/Handlers/EmojiHandler.cs
+++ b/TShockAPI/Handlers/EmojiHandler.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TShockAPI.Handlers
+{
+	/// <summary>
+	/// Handles an exploit and checks for permissions.
+	/// </summary>
+	public class EmojiHandler
+	{
+		public void OnEmoji(object sender, GetDataHandlers.EmojiEventArgs args)
+		{
+			if (args.PlayerIndex != args.Player.Index)
+			{
+				TShock.Log.ConsoleError($"EmojiHandler: Packet is spoofing to be player ID {args.PlayerIndex}! - From [{args.Player.Index}]{args.Player.Name}");
+				args.Handled = true;
+				return;
+			}
+
+			if (!args.Player.HasPermission(Permissions.sendemoji))
+			{
+				args.Player.SendErrorMessage("You have no permission to send emotes!");
+				args.Handled = true;
+				return;
+			}
+		}
+	}
+}
--- a/TShockAPI/Handlers/EmojiHandler.cs
+++ b/TShockAPI/Handlers/EmojiHandler.cs
@@ -7,22 +7,22 @@ using System.Threading.Tasks;
 namespace TShockAPI.Handlers
 {
 	/// <summary>
-	/// Handles an exploit and checks for permissions.
+	/// Handles emoji packets and checks for validity and permissions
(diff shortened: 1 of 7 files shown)


TShockAPI/Handlers/EmojiHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+	/// <summary>
+	/// Handles an exploit and checks for permissions.
+	/// </summary>
+	public class EmojiHandler
+	{
+		public void OnEmoji(object sender, GetDataHandlers.EmojiEventArgs args)
+		{
+			if (args.PlayerIndex != args.Player.Index)
+			{
+				TShock.Log.ConsoleError($"EmojiHandler: Packet is spoofing to be player ID {args.PlayerIndex}! - From [{args.Player.Index}]{args.Player.Name}");
```
Patrikkk: Should I change this to ConsoleDebug instead?

QuiCM: Yep.
You can also be less formal with ConsoleDebug messages because the intention is that we ask people to send specific sections of their log to capture specific messages.
I'd suggest a rephrasing to something like `Emoji packet rejected for ID spoofing. Expected {playerId}, received {actualId} from {player}`

TShockAPI/Handlers/EmojiHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+
+namespace TShockAPI.Handlers
+{
+	/// <summary>
+	/// Handles an exploit and checks for permissions.
+	/// </summary>
+	public class EmojiHandler
+	{
+		public void OnEmoji(object sender, GetDataHandlers.EmojiEventArgs args)
```
QuiCM: My preference for naming here would be `OnReceiveEmoji`, but I may add some frameworking around the `Handlers` namespace and revisit this

TShockAPI/Handlers/EmojiHandler.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TShockAPI.Handlers
+{
+	/// <summary>
+	/// Handles an exploit and checks for permissions.
```
QuiCM: This could be more informative.
`Handles emoji packets and checks for validity and permissions` perhaps?

TShockAPI/Handlers/EmojiHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+			if (args.PlayerIndex != args.Player.Index)
+			{
+				TShock.Log.ConsoleError($"EmojiHandler: Packet is spoofing to be player ID {args.PlayerIndex}! - From [{args.Player.Index}]{args.Player.Name}");
+				args.Handled = true;
+				return;
+			}
+
+			if (!args.Player.HasPermission(Permissions.sendemoji))
+			{
+				args.Player.SendErrorMessage("You have no permission to send emotes!");
```
QuiCM: I believe our normal verbiage is `You do not have permission to ...`

QuiCM (changes requested):
Minor nitpicks and phrasing. Overall looks good

QuiCM (approved):

PR #1970, opened 2020-06-01
Add HandleSyncCavernMonsterType

Patrikkk:
This packet is never sent from the client to the server in a normal scenario. Although with modded clients, a packet can be sent to modify the cavernMonsterType of the server world and have the world spawn defined NPC types. Can be used to have the server randomly spawn bosses on players in caverns.
Is it okay to have this simple handling in GetDataHandlers? A separate class felt like an overkill.

Moved the HandleSyncRevengeMarker packet handler to it's "correct" position, so I won't have merge issues between my last PR. As I've mentioned there, we have the Packets in their numerical order.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -151,9 +151,10 @@ namespace TShockAPI
 					{ PacketTypes.CrystalInvasionStart, HandleOldOnesArmy },
 					{ PacketTypes.PlayerHurtV2, HandlePlayerDamageV2 },
 					{ PacketTypes.PlayerDeathV2, HandlePlayerKillMeV2 },
+					{ PacketTypes.SyncRevengeMarker, HandleSyncRevengeMarker },
 					{ PacketTypes.FishOutNPC, HandleFishOutNPC },
 					{ PacketTypes.FoodPlatterTryPlacing, HandleFoodPlatterTryPlacing },
-					{ PacketTypes.SyncRevengeMarker, HandleSyncRevengeMarker }
+					{ PacketTypes.SyncCavernMonsterType, HandleSyncCavernMonsterType }
 				};
 		}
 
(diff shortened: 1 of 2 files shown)


QuiCM:
Is it worth kicking if this packet is received?
Assuming the vanilla client never sends this in any way shape or form, we can be certain that anyone sending it to the server is cheating and should be kicked

TShockAPI/GetDataHandlers.cs
```diff
@@ -3671,6 +3672,21 @@ private static bool HandlePlayerKillMeV2(GetDataHandlerArgs args)
 			return false;
 		}
 
+		private static bool HandleSyncRevengeMarker(GetDataHandlerArgs args)
+		{
+			int uniqueID = args.Data.ReadInt32();
+			Vector2 location = args.Data.ReadVector2();
+			int netId = args.Data.ReadInt32();
+			float npcHpPercent = args.Data.ReadSingle();
+			int npcTypeAgainstDiscouragement = args.Data.ReadInt32(); //tfw the argument is Type Against Discouragement
```
QuiCM, line 3612: what does the comment here mean?

Patrikkk: I didn't touch it, not my comment. I only moved the whole method up so things are ordered by packet type.

QuiCM: ah fair enough

Patrikkk:
I'm fine with kicking if this is received. I'll add it.
Just to back up my statement. The only time packet is sent is in messagebuffer GetData packet type case 8, when player requests more stuff from server on join. That's when server sends this packet.

QuiCM:
> 
> 
> I'm fine with kicking if this is received. I'll add it.
> Just to back up my statement. The only time packet is sent is in messagebuffer GetData packet type case 8, when player requests more stuff from server on join. That's when server sends this packet.

I think a case can be made for kicking, but I'm fine to accept this without a kick in it currently and we can discuss further later on :)

QuiCM (approved):

PR #1975, opened 2020-06-02
Move netmodule handling into handlers namespace

QuiCM:
Moves the existing netmodule handling into new namespace `TShock.Handlers.NetModules`.
Adds a new netmodule handler for the Teleport Pylon net module, along with a new permission `tshock.tp.pylon` to enable/disable pylon teleports.
Also adds new netmodule handlers for all the netmodules we should process or instantly reject.
As far as I am aware the only unhandled modules are now the NetAmbienceModule and NetPingModule. Both could potentially be abused via spam, but are otherwise 'safe'

### DIFF
--- a/TShockAPI/Handlers/NetModules/CreativePowerHandler.cs
+++ b/TShockAPI/Handlers/NetModules/CreativePowerHandler.cs
@@ -1,0 +1,113 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.IO.Streams;
+using static TShockAPI.GetDataHandlers;
+
+namespace TShockAPI.Handlers.NetModules
+{
+	/// <summary>
+	/// Provides handling for the Creative Power net module. Checks permissions on all creative powers
+	/// </summary>
+	public class CreativePowerHandler : INetModuleHandler
+	{
+		/// <summary>
+		/// The power type being activated
+		/// </summary>
+		public CreativePowerTypes PowerType { get; set; }
+
+		/// <summary>
+		/// Reads the power type from the stream
+		/// </summary>
+		/// <param name="data"></param>
+		public void Deserialize(MemoryStream data)
+		{
+			PowerType = (CreativePowerTypes)data.ReadInt16();
+		}
+
+		/// <summary>
+		/// Determines if the player has permission to use the power type
+		/// </summary>
+		/// <param name="player"></param>
+		/// <param name="rejectPacket"></param>
+		public void HandlePacket(TSPlayer player, out bool rejectPacket)
+		{
+			if (!CheckPermission(PowerType, player))
+			{
+				rejectPacket = true;
+				return;
+			}
+
+			rejectPacket = false;
+		}
+
+		/// <summary>
+		/// Determines if a player has permission to use a specific creative power
+		/// </summary>
+		/// <param name="powerType"></param>
+		/// <param name="player"></param>
+		/// <returns></returns>
+		public static bool CheckPermission(CreativePowerTypes powerType, TSPlayer player)
+		{
+			if (!PowerToPermissionMap.ContainsKey(powerType))
+			{
+				TShock.Log.ConsoleDebug("CreativePowerHandler received permission check request for unknown creative power");
+				return false;
+			}
+
+			string permission = PowerToPermissionMap[powerType];
+
+			if (!player.HasPermission(permission))
+			{
+				player.SendErrorMessage("You do not have permission to {0}.", PermissionToDescriptionMap[permission]);
+				return false;
+			}
+
+			return true;
+		}
+
+
+		/// <summary>
+		/// Maps creative powers to permission nodes
+		/// </summary>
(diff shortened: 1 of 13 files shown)


TShockAPI/Handlers/NetModules/CreativePowerHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+			rejectPacket = false;
+		}
+
+		/// <summary>
+		/// Determines if a player has permission to use a specific creative power
+		/// </summary>
+		/// <param name="powerType"></param>
+		/// <param name="player"></param>
+		/// <returns></returns>
+		public static bool CheckPermission(CreativePowerTypes powerType, TSPlayer player)
```
hakusaro: `Check` is ambiguous (prefer `Has`)

TShockAPI/Handlers/NetModules/CreativeUnlocksHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+		/// </summary>
+		public ushort Amount { get; set; }
+
+		/// <summary>
+		/// Reads the unlock data from the stream
+		/// </summary>
+		/// <param name="data"></param>
+		public void Deserialize(MemoryStream data)
+		{
+			UnknownField = data.ReadInt8();
```
hakusaro, line 36: This is worth annotating with a comment as to why it's de-serialized and why it's unknown.

TShockAPI/Handlers/NetModules/NetModulePacketHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+			{
+				// As of 1.4.x.x, this is now used for more things:
+				//	NetCreativePowersModule
+				//	NetCreativePowerPermissionsModule
+				//	NetLiquidModule
+				//	NetParticlesModule
+				//	NetPingModule
+				//	NetTeleportPylonModule
+				//	NetTextModule
+				// I (particles) have disabled the original return here, which means that we need to
```
hakusaro: This comment probably needs updating with remaining netmodules and removal of already implemented ones

TShockAPI/Handlers/NetModules/PylonHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+		/// </summary>
+		/// <param name="player"></param>
+		/// <param name="rejectPacket"></param>
+		public void HandlePacket(TSPlayer player, out bool rejectPacket)
+		{
+			if (PylonEventType == SubPacketType.PlayerRequestsTeleport)
+			{
+				if (!player.HasPermission(Permissions.pylon))
+				{
+					rejectPacket = true;
```
hakusaro, line 53: this is heckin good, super clear "rejectPacket" verbiage is nice

hakusaro (approved):
I much prefer this approach over the original one we went with. I think this is great.

The point that was raised that `LoadNetModule` is a bad name is probably still true though. `Load` always gave me the impression that we were initializing it or something. In truth it's really like `TransferNetModule` or something. If we wanted to change it, now would be the time to change it.

`INetModuleHandler` is a breath of fresh air. I think it could be improved if we were able to twist it into an object with a single purpose of determining rejection. Right now, the classes are basically zombies: they're driven entirely by external function calls. This means that outside knowledge is required to call `Deserialize` before `HandlePacket`. Obviously, making these entirely focused on rejection/acceptance would probably be non-ideal from an external user PoV. I can definitely see arguments both ways.

Either way, this is definitely a 10x improvement over what we've got now.

QuiCM:
> 
> 
> I much prefer this approach over the original one we went with. I think this is great.
> 
> The point that was raised that `LoadNetModule` is a bad name is probably still true though. `Load` always gave me the impression that we were initializing it or something. In truth it's really like `TransferNetModule` or something. If we wanted to change it, now would be the time to change it.


I've renamed `LoadNetModule` to `ReadNetModule` which I think covers the purpose of the getdatahandler event and the handlers pretty well

QuiCM:
> `INetModuleHandler` is a breath of fresh air. I think it could be improved if we were able to twist it into an object with a single purpose of determining rejection. Right now, the classes are basically zombies: they're driven entirely by external function calls. This means that outside knowledge is required to call `Deserialize` before `HandlePacket`. Obviously, making these entirely focused on rejection/acceptance would probably be non-ideal from an external user PoV. I can definitely see arguments both ways.
> 
> Either way, this is definitely a 10x improvement over what we've got now.

I'm not a huge fan of needing to call Deserialize before HandlePacket - from an external perspective it sounds like you could just call HandlePacket and it would deal with it. But I dislike the idea of the developer needing to know to call Deserialize inside HandlePacket even more. Not sure what an elegant solution would really be.
However given that the flow is controlled entirely inside of NetModulePacketHandler currently, and new handlers can be added by adding/editing `NetModulePacketHandler.NetModulesToHandlersMap` I think the current implementation ends up being fairly smooth

PR #1982, opened 2020-06-03
Rewrite ssc

QuiCM:
This is wildly untested and smashed out in a night.
Some hardcore review (and testing if possible!) would be appreciated.

Pulled as much of the SSC subsystem out of the various handlers and methods it was in as I could and turned it into a first class system. It now hooks things on its own and is hopefully a lot cleaner to look at


QuiCM:
No idea if overridessc and uploadssc commands still work.
No idea how stable anything is.
No idea if there's other login/logout/account/etc use cases I haven't considered yet
Haven't tested that it works with an aged SSC database, but not much has changed in terms of data in/out

TShockAPI/ServerSideCharacters/ServerSideCoordinator.cs
```diff
@@ -0,0 +1,495 @@
...
+using TShockAPI.DB;
+using TShockAPI.Hooks;
+using TShockAPI.Net;
+
+namespace TShockAPI.ServerSideCharacters
+{
+	/// <summary>
+	/// Provides and executes CRUD operations on server side player data
+	/// </summary>
+	public class ServerSideCoordinator
```
QuiCM, line 16: I'm not sure about this name or this class.
Feels like its trying to do too much at once (hence an ambiguous name like 'Coordinator'), so suggestions on design here would be appreciated

rickojp:
> No idea if overridessc and uploadssc commands still work.
> No idea how stable anything is.
> No idea if there's other login/logout/account/etc use cases I haven't considered yet
> Haven't tested that it works with an aged SSC database, but not much has changed in terms of data in/out

Some test results:

-Overridessc and uploadssc commands do not work (Log: Command: ERROR: System.NullReferenceException)
-Logout account does not ignore the player the same way as when the player enters the server the first time
-I used a clean database created by that branch and in another test an existing database and apparently there were no problems
-Superadmin accounts or groups with permission 'tshock.ignore.ssc', usually ignore SSC inventory, but some errors occur. 
As I don't usually use this function, I don't know if in the old way it is intentional to save the current items in the database. So one must treat or add the check if the player has permission in [SaveAllPlayersData](https://github.com/Pryaxis/TShock/blob/a8532ca750625062cc34b1a96be72fa5e09c6003/TShockAPI/ServerSideCharacters/ServerSideCoordinator.cs#L212l), [OnPlayerLogout](https://github.com/Pryaxis/TShock/blob/a8532ca750625062cc34b1a96be72fa5e09c6003/TShockAPI/ServerSideCharacters/ServerSideCoordinator.cs#L80), [OnPlayerHpChange](https://github.com/Pryaxis/TShock/blob/a8532ca750625062cc34b1a96be72fa5e09c6003/TShockAPI/ServerSideCharacters/ServerSideCoordinator.cs#L135), [OnPlayerManaChange](https://github.com/Pryaxis/TShock/blob/a8532ca750625062cc34b1a96be72fa5e09c6003/TShockAPI/ServerSideCharacters/ServerSideCoordinator.cs#L147)

Thank for this work!

PR #2020, opened 2020-06-07
Reorganize ConfigFile.cs for more orderly config.json output

kubedzero:
Every time the server is loaded from scratch or the `/reload` command is called, the `config.json` file will populate missing values to be in line with `ConfigFile.cs` and reorder the JSON to match the order defined in the file. As discussed in https://github.com/Pryaxis/TShock/discussions/1964 this PR is a refactor of the order of the variables in the file, which should update the order of the JSON output on each server load. In addition, I added a couple missing summary lines and updated the changelog. 

I tried to categorize the parameters into a few main sections. Server, backup/save, login/ban, chat, MySQL/REST, world settings. Within those I moved  a few settings around to be closer to their "siblings."

I don't have a dev environment set up for this so I won't be able to compile and ensure the updated order works as intended. However,  I did a text comparison of the existing production ConfigFile.cs at https://raw.githubusercontent.com/Pryaxis/TShock/07f0dc1540bf91ad9a47f3ed63e6d2ebbb560402/TShockAPI/ConfigFile.cs and my own version, stripping out the methods at the bottom and whitespace and stuff. I was able to confirm:

* All lines beginning with `[Description]` stayed unchanged in count and content
* All lines beginning with `/// <summary>` stayed unchanged in content. Two new lines were added for `BombExplosionRadius` and `DebugLogs` as they were missing
*  There are 132 variables in the new file, which is unchanged from the ConfigFile.cs file in production. 

I'm submitting this PR with these validations and the understanding that this change should only affect the order of the `config.json` output and no other functional change.

Sorry for the big diff, it's almost exclusively blocks of code moving around so if there's a way to diff the compiled output, only the summary lines should have changed (if those are even compiled)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming changes
 * Installed new sprinklers!
+* Organized parameters by category and relevance in the `config.json` file. (@kubedzero)
 
 ## TShock 4.4.0 (Pre-release 11)
 * New permission `tshock.tp.pylon` to enable teleporting via Teleportation Pylons (@QuiCM)
--- a/TShockAPI/ConfigFile.cs
+++ b/TShockAPI/ConfigFile.cs
@@ -31,7 +31,7 @@ namespace TShockAPI
 	public class ConfigFile
 	{
 
-		// Server
+		/* ~~~~~ Server ~~~~~ */
 
 		/// <summary>The server password required to join the server.</summary>
 		[Description("The server password required to join the server.")]
@@ -74,7 +74,7 @@ namespace TShockAPI
 		public bool IgnoreChestStacksOnLoad = false;
 
 
-		// Backup and Save
+		/* ~~~~~ Backup and Save ~~~~~ */
 
 		/// <summary>Enable or disable Terraria's built-in world auto save.</summary>
 		[Description("Enable or disable Terraria's built-in world auto save.")]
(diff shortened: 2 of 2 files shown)


kubedzero:
TODO: I see there are some build artifacts from https://github.com/Pryaxis/TShock/pull/2020/checks?check_run_id=747924860, maybe I can download those to confirm the config file is sorted and that the server still starts up

Patrikkk:
Thank you for your PR:) 
Putting things in region blocks would be better to keep things organized since these blocks can be collapsed and expanded in Visual Studio.
```
#region  
data here
#endregion
```

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-region

kubedzero:
> Putting things in region blocks would be better

Sure let me give that a shot and post an update. Thanks!

kubedzero:
Great, here's the new output order of the JSON. I pulled the non-debug build artifact, unzipped and started up, and then this Paste is the contents of the config.json: https://pastebin.com/c0FdATqr

Patrikkk:
My only concern is how it would affect an existing config. But it's about to be 4 am so that's a question for tomorrow.

QuiCM:
> 
> 
> My only concern is how it would affect an existing config. But it's about to be 4 am so that's a question for tomorrow.

shouldn't affect existing configs. JSON is great that way  :)

kubedzero:
Customized JSON from a pre-11 running instance https://pastebin.com/vpsnKyuS

I then copied it over into the test build and started the server. The resulting JSON can be seen here, in line with reorganization expectations and equal in other aspects to the original: https://pastebin.com/3kTQsWCS

QuiCM (approved):

hakusaro:
The organization of this looks good to me, but I haven't had time to review for correctness to make sure anything got transposed in terms of copying and pasting around things. If someone else can sign off that it's technically the same as before, just reordered, I'm good.

Issue #2017, opened 2020-06-07
Replace blocks and 798 projectile (Mini Nuke Proximity Mine II)


Ap6y3uk:
1) If I put a banned block(tile) HIVE - Server disable me.
But if I replace the block with HIVE - HIVE place
2) If I create 798 projectile - Server disable me.
But projectile is being created and after a while it explodes (checked: it explodes the dirt blocks)

-> TShock 4.4.0 Pre-release 11 for Terraria 1.4.0.5

VIDEO (I cut the gaps of the video to make it faster): https://youtu.be/EKZB6h1IPCQ

Ap6y3uk:
+ do not save History (history plugin) if place banned tile.

QuiCM:
Replace resolved, still pending work on projectile 798

PR #2024, opened 2020-06-08
Bouncer OnTileEdit - Add tile replace action to tileban check.

Patrikkk:
Tiny change, but a pressing matter. My smallest PR ever? hahah.
Referencing #2017 , as this patches the tile replace bypass issue that is mentioned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming changes
 * Installed new sprinklers!
-* Fix issue where players could replace tiles with banned tiles without permission. (@Patrikkk)
+* Fix multiple holes in Bouncer OnTileData. (@Patrikkk, @hakusaro)
+	* Issue where players could replace tiles with banned tiles without permission. 
+	* Including replace action in TilePlace threshold incrementation, so players cannot bypass the threshold while replacing tiles/walls.
+	* Including check for maxTileSets when player is replacing tiles, so players cannot send invalid tile data through the replace tile action.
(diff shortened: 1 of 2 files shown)


hakusaro (approved):
This looks great! Can you also update these checks, because they're related to updating thresholds?
* https://github.com/Pryaxis/TShock/pull/2024/files#diff-3086bd09c54e93b5f9a34786586f710cL539
* https://github.com/Pryaxis/TShock/pull/2024/files#diff-3086bd09c54e93b5f9a34786586f710cR548

Your changes help enforce replacing tiles in terms of ban state, but it doesn't look like replacements are covered by TileKill/TilePlace thresholds. If you don't want to, I can happily merge this without them :)

hakusaro:
Bah, those links are not good. There are two threshold checks in the same method -- TilePlaceThreshold and TileKillThreshold. Both deserve to be ticked and currently they only happen on Place/Kill events, not replace events.

Patrikkk:
My bad:) I just quickly patched the issue that someone reported to me, I didn't look for any leftovers. 
The links didn't work indeed. I tried to spot the lines you mentioned, but not sure I got both.
At the same time, I didn't want to spam in redundant checks or ones that don't exactly make sense. 
I just sent another commit that adds two more checks. One is for tile Threshold. Other is for the maxtile check.
Aaand I'm sending another because I forgot to update changelog on this.

hakusaro:
L548 is still needed, I think.

Patrikkk:
Indeed, thank you:) That made me notice a few other checks regarding walls and tile killing.
Gotta be honest the checks in the method are quite badly documented, makes it hard to understand what some of them do. If I find time I might clean things up a bit.

PR #2041, opened 2020-06-11
Bouncer OnTileEdit - Fix valid replacment action getting caught.

Patrikkk:
In a previous commit, I've added the checks for TileReplace where I believed it was necessary.  By placing a check for the tile replace action, in another check, this action got falsely caught and ignored by the player.
This commit fixes that.
I've also moved the block that checks for max valid edit data closer to the start of the method, where we have a check for edit data not being smaller than 0.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -231,6 +231,14 @@ namespace TShockAPI
 					args.Handled = true;
 					return;
 				}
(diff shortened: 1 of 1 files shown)


Patrikkk:
@hakusaro  Was I supposed to write a changelog for this? I'm asking because I feel like I didn't need to, as there was no prerelease yet containing the issue. This is rather a followup commit to my previous one, as I've left out a simple check that created an issue that didn't exist previously. 
So I wasn't sure what I was supposed to write in the changelog, apart from what already exists regarding adding replace action checks to Bouncer OnTileEdit.
So let me know!

PR #2043, opened 2020-06-11
Bouncer OnTileEdit - Code cleanup

Patrikkk:
I moved the block that checks for max tile and wall types to the start of the method where it checks for the editData (type) being smaller than 0.
I merged these checks, so it gets caught under the same hood.

We had a block that was a bit chaotic and hard to understand.
I've split it up so its a bit clearer. It checks if the tile data that is being placed, comes from the item they are selecting. There were additional checks merged in it because terraria does not set the createTile of some items, so without the check it would have gotten caught as invalid placement.
I added a valid placement check for one of these, which is the Ice Rod/Iceblock placement.
Handle action if the player is using icerod but not placing ice block.
There is no need for a check here on Dirtbomb because the player only sends the projectile, the tiles are being placed by the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 	* Adding staff projectiles to the directionalProjectiles Dictionary to include staffs in the valid projectile creation check.
 	* Adding GolfBallItemIDs list in Handlers.LandGolfBallInCupHandler.cs
 * Fixed an issue in the SendTileSquare handler that was rejecting valid tile objects (@QuiCM)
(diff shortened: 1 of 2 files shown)


Patrikkk:
Do I need to document this in changelog?
This is rather a backend tidy up, which should not affect any functionality.

Patrikkk:
This would also cover #2041 
I would prefer if we merge this PR and nerf  #2041.

hakusaro:
> Do I need to document this in changelog?

Let me be as unambiguous as possible: if you ask, I will literally always say yes.

Patrikkk:
There, :) Hope that's fine. I wasn't sure how to word this.

QuiCM (approved):
LGTM

QuiCM:
Fix up conflicts and should be good to merge

PR #2044, opened 2020-06-11
Bouncer OnTileEdit - Remove person storage tile check for SSC

Patrikkk:
We do not allow piggybank and safe to be placed if SSC is enabled.
The DefendersFroge and VoidVault have the same functionality as far as I know. So adding these two missing tiles.


kevzhao2:
Why do we disallow these tiles? Now that piggy bank/safe are synced with the server, it should be okay to actually delete the entire check.

QuiCM:
We no longer need to disable these tiles

QuiCM (changes requested):
We no longer need to disable these tiles

Patrikkk:
Thanks for the info guys! 
Want another commit to remove that check? Should I kill the List of personalStorage as well?

QuiCM:
> 
> 
> Thanks for the info guys!
> Want another commit to remove that check? Should I kill the List of personalStorage as well?

Yes please and yes please. I'll update the PR title :)

Patrikkk:
Sent:)

QuiCM (approved):

PR #2053, opened 2020-06-13
Bouncer OnTileEdit - Remove redundant Boulder placement check

Patrikkk:
Remove redundant Boulder placement check that prevented placing chests on them, as it is no longer possible to place a chest on a boulder in vanilla, so nothing crashes the server. "1.2.3: Boulders with Chests on them no longer crash the game if the boulder is hit."


Olink:
While the client might not be able to place a chest on a boulder, does the server allow the tile placement?  We should be preventing hacked clients from abusing whatever exploit this caused.

Patrikkk:
Good thought! I will give it a test.:)
While it won't hurt to keep it in, I personally wouldn't like TShock to be code piled on code that is redundant/undocumented/inefficient.

kevzhao2:
Terraria's changelog description seems to imply that if there is a chest on a boulder somehow, then the crash doesn't occur. Which is slightly different from the CHANGELOG.md interpretation, which says it's impossible to do it in the first place.

Patrikkk:
@kevzhao2  I went ahead to place tiles on the boulder. And I simply could not place a chest on top of it, but anything else. And it was not handled by TShock, but rather the client itself.

kevzhao2 (approved):
LGTM

Issue #2018, opened 2020-06-07
Add RegionProtectMannequins if possible 


moisterrific:
## Current behavior
Mannequins (and Womannequins) are in fact "protected" tile squares if they are in placed within protected regions, unauthorized users cannot remove them, but they can still open the mannequin and remove/change the stored items. 

## Suggested change
Add a new config defined option `RegionProtectMannequins` (`false` by default for consistency w/ other `RegionProtectX` options. 

Not sure if this is possible though, cuz right now I'm aware of how TShock protects chests if `RegionProtectChests = true`, but dunno if Mannequins use the same packet or a different one.

**From `GetDataHandlers.cs` line 2755:**
```csharp
			if (!args.Player.HasBuildPermission(x, y) && TShock.Config.RegionProtectChests)
			{
				TShock.Log.ConsoleDebug("GetDataHandlers / HandleChestActive rejected build permission and region check {0}", args.Player.Name);
				args.Player.SendData(PacketTypes.ChestOpen, "", -1);
```

Also I dunno how hat racks are affected, maybe they behave the same way as existing mannequins?

QuiCM:
Think this should be in latest prerelease now :)

PR #2061, opened 2020-06-18
RequestTileEntityInteraction hook and handling.

Patrikkk:
Implemented the OnRequestTileEntityInteraction even hook in GetDataHandler.
Created RequestTileEntityInteractionHandler which checks for building permissions when the player is attempting to open a display doll (Mannequin). This now prevents players from opening a mannequin if they have no building permissions at the position of the Mannequin.


Patrikkk:
Referencing issue #2018

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+	/// 
+	/// </summary>
+	public class RequestTileEntityInteractionHandler : IPacketHandler<RequestTileEntityInteractionEventArgs>
+	{
+		public void OnReceive(object sender, RequestTileEntityInteractionEventArgs args)
+		{
+			if (args.TileEntityID != -1)
+			{
+				TileEntity tileEntity;
+				if (TileEntity.ByID.TryGetValue(args.TileEntityID, out tileEntity))
```
QuiCM: Can this ever set tileEntity to `null`?

bartico6: This is a packet handler, so if the client sends an invalid ID on purpose (or sends one right as one was removed) this will return false.

Patrikkk: TileEntity.ByID is a dictionary of int, TileEntity, and the dictionary does not store any nulls to the ID.s 

So this block only executes if the TryGet returns true and it can get the entity by ID, and it will not return a null object.
In my next PR regarding DollItemSync, I am passing the tileentity object. 
Should I modify this PR and pass the tile entity object, instead of just the ID, so it gives better usability for Devs when they work with this hook?

QuiCM: > 
> Should I modify this PR and pass the tile entity object, instead of just the ID, so it gives better usability for Devs when they work with this hook?

That sounds good to me

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace TShockAPI.Handlers
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public class RequestTileEntityInteractionHandler : IPacketHandler<RequestTileEntityInteractionEventArgs>
+	{
+		public void OnReceive(object sender, RequestTileEntityInteractionEventArgs args)
+		{
+			if (args.TileEntity is TEHatRack && !args.Player.HasBuildPermission(args.TileEntity.Position.X, args.TileEntity.Position.Y, false))
```
QuiCM: Might be worth using `args.Player.HasBuildPermissionForTileObject` here, as it will check all tiles in the entity for build permission

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+			{
+				args.Player.SendErrorMessage("You do not have permission to modify a Hat Rack in a protected area!");
```
QuiCM, line 22: These if statements all perform the same logic.
Might be nice to separate it such that the if statements determine what text to send and then perform the logic outside of it.
Eg
```csharp
string type = "Tile Entity";
if (args.TileEntity is TEHatRack)
{
    type = "Hat Rack";
}
// etc

args.Player.SendErrorMessage($"You do not have permission to modify this {type}.");
TShock.Log.ConsoleDebug($"RequestTileEntityInteractionHandler: Rejected packet due to lack of building permissions - From {args.Player.Name} | Position X:{args.TileEntity.Position.X} Y:{args.TileEntity.Position.Y}, TileEntity type: {args.TileEntity.type}, Tile type: {Main.tile[args.TileEntity.Position.X, args.TileEntity.Position.Y].type}");
args.Handled = true;
return;
```

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
Patrikkk: Do you know what is the difference between the two checks? If not, I can go have a look.

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
Patrikkk: What's your reasoning for that code approach? 
I'm assuming is to have one block of logic that (the output messages) with a variable to tend to, rather than having the same code with a hardcoded variant.
It is rather only visuals/organization, than performance because it would not change much at all. (an extra string being instantiated is minuscule) 
But yeah I don't mind to change it to that at all. Just simply curious of your way of thinking:)

QuiCM: Mostly just clarity. We have many spots in TShock where we have the same logic run in multiple if/else blocks with slightly different messages and it ends up being hard to maintain/adjust/extend

Patrikkk: Yeah, I was thinking the same once you've pointed it out. Thanks!:)

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
Patrikkk: Ah, TileObject requires width and height. This data is not in a TileEntity object, would need to be cast as a TEHatRack or TEDisplayDoll.

TShockAPI/Handlers/RequestTileEntityInteractionHandler.cs
Patrikkk: If you check the other comment above, to use HasBuildPermissionForTileObject I would need to cast the object as TEHatRack or TEDisplayDoll.  This means I would need to have 

```c#
if (args.TileEntity is TEHatRack && !args.Player.HasBuildPermissionForTileObject(args.TileEntity.Position.X, args.TileEntity.Position.Y, ((TEHatRack)args.TileEntity).entityTileWidth, ((TEHatRack)args.TileEntity).entityTileHeight,  false))
```

And keep the same logic structure.

Before looking into ForTileObject perm check I made this edit

hakusaro:
FYI: I've updated your changelog entry to match renormalized changelog changes on the changelog.

Patrikkk:
I've now implemented a change that I see the best fit for allowing HasBuildPermissionForTileObject to be used for an accurate check over a Hat Rack and Mannequin. If @QuiCM is okay with this approach, feel free to merge.
I'm open for suggestions for better code handling.

QuiCM:
> 
> 
> I've now implemented a change that I see the best fit for allowing HasBuildPermissionForTileObject to be used for an accurate check over a Hat Rack and Mannequin. If @QuiCM is okay with this approach, feel free to merge.
> I'm open for suggestions for better code handling.

For the moment I think this is the best solution. Ideally I'd like to be able to condense the if/else blocks but from earlier discussion and experimentation its more trouble than its worth.
LGTM

QuiCM (approved):

PR #2062, opened 2020-06-19
Displaydollsynchook

Patrikkk:
**The event:**
An event that is called when a player modifies the slot of a DisplayDoll (Mannequin).

I was trying to think from a developer-friendly perspective here.
Instead of passing separate variables for type/stack/prefix, thought I pass an Item object.
As well as, instead of having devs who work with this hook figure out and implement how to get the Item of the DisplayDoll, I just provide it in the hook.
I can imagine this being used for creative purposes in plugins.

**The handler:**
In a previous PR I have added the tile entity request packet handler which checks for building permissions to prevent the unauthorized player to open a DisplayDoll and see its content.

This Handler is being added to prevent *Hackers* from modifying a DisplayDoll through direct/crafted packet sending, or by sending raw byte data to the server.
In a valid environment, the player couldn't even get to see the content of the doll in the first place, to then try to modify its items.

Because of this,  I do not bother with making sure the player gets their item back.


TShockAPI/GetDataHandlers.cs
```diff
@@ -3703,6 +3762,48 @@ private static bool HandleEmoji(GetDataHandlerArgs args)
...
+
+				ushort itemType = args.Data.ReadUInt16();
+				ushort stack = args.Data.ReadUInt16();
+				int prefix = args.Data.ReadByte();
+
+				newItem.SetDefaults(itemType);
+				newItem.stack = stack;
+				newItem.Prefix(prefix);
+
+				if (oldItem.type == 0 && newItem.type == 0)
```
QuiCM: Can this check occur before calling SetDefaults?

QuiCM: Maybe also check for `stack = 0` somewhere

QuiCM: Can `oldItem` be null?

Patrikkk: If stack is 0, nothing happens, no item is placed.
`oldItem` is never null because the doll entity has default Item() object set for empty slots.
Sent commit for the type check order.

QuiCM (approved):

PR #2073, opened 2020-06-25
Missing parenthesis from an if logic. Fix valid Tile placement reject.

Patrikkk:
A previous PR slipped in which missed a parenthesis in an IF logic which used both AND and OR operators, which resulted in valid tile placement being rejected.
No changelog update because this issue did not get into a (pre)release and not something that should be documented.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -463,7 +463,7 @@ namespace TShockAPI
 				}
 
 				//make sure it isnt a snake coil related edit so it doesnt spam debug logs with range check failures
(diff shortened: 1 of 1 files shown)


hakusaro:
In the future, please use present imperative tense in the commit message, e.g., the first word should be a verb in active voice denoting what you've done.

"Fix valid tile placement reject defective if statement

A previous PR slipped in which missed a parenthesis in an IF logic which used both AND and OR operators, which resulted in valid tile placement being rejected."

^ this would be an ideal commit message

hakusaro:
i.e., https://chris.beams.io/posts/git-commit

PR #2101, opened 2020-07-27
Fix config file access errors when starting up multiple TShock servers at once

ZakFahey:
Use case would be multi-server networks. This commit ensures that sscconfig.json is read with FileShare.Read enabled and stops writing to config.json and sscconfig.json on server load immediately after reading it if the file already exists, which is a no-op since neither of the config files have changed at that point.


ZakFahey:
Alright, fixed. Now the configs will be overwritten, but only if the file doesn't exist or if the list of fields in the json doesn't match the list of fields in the config class.

Ijwu (approved):
Looks good to me. Per 2 review rule, I'll leave this for @QuiCM or @hakusaro to merge.

PR #2170, opened 2020-11-26
New bans + some other stuff

QuiCM:
* Bans completely overhauled
   - Bans are based on single pieces of identifiable info (e.g., one of: uuid, IP, name, account name) rather than an amalgamation of the lot
     - A generic user ban (account, IP, UUID) would result in three unique bans
   - Bans expire, but remain in the system. New bans for a given identifier can only be added once an existing ban for that identifier expires. Expired bans can still be inspected, but are  not actively displayed
     - I.E., you cannot have two enforced bans for the same UUID. Note that this is enforced in code, not in database.
  - Conversion path exists from old bans -> new bans

* Console automagically detects Terraria colour codes and attempts to display 'nearby' colours (only 'nearby' because ConsoleColor is pretty sparse, and Windows Cmd prompt doesn't properly support ANSI color codes yet)

* Shutdown now disconnects rather than kicks. There's something wrong with the kick code that can cause a connection drop rather than a clean disconnect

* Sudo should now properly forward `"quoted parameters"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,10 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming Changes
-* Overhauled Bans system. Bans are now based on 'identifiers'.(@QuiCM)
-  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Use `/ban convert` to convert to the new system. 
+* Overhauled Bans system. Bans are now based on 'identifiers'. (@QuiCM)
+  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Old bans will be converted automatically to the new system.
   * All old ban routes in REST are now redirected. Please use `/v3/bans/*` for REST-based ban management.
-  * TShock recognizes and acts upon 4 main identifiers: UUID, IP, Player Name, Account name. This can be extended by plugins.
+  * TShock recognizes and acts upon 4 main identifiers: UUID, IP, Player Name, Account name. This can be extended by plugins. New identifiers can be added to the `ban help identifiers` output by registering them in `TShockAPI.DB.Identifier.Register(string, string)`
   * By default, bans are no longer removed upon expiry or 'deletion'. Instead, they remain in the system. A new ban for an indentifier can be added once an existing ban has expired.
 * Server Console now understands Terraria color codes (e.g., `[c/FF00FF:Words]`) and prints the colored text to the console. Note that console colors are limited and thus only approximations. (@QuiCM)
 * Fixed a bug in `/sudo` that prevented quoted arguments being forwarded properly. Example: `/sudo /user group "user name" "user group"` should now work correctly. (@QuiCM)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming Changes
+* Overhauled Bans system. Bans are now based on 'identifiers'.(@QuiCM)
+  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Use `/ban convert` to convert to the new system. 
+  * All old ban routes in REST are now redirected. Please use `/v3/bans/*` for REST-based ban management.
+  * TShock recognizes and acts upon 4 main identifiers: UUID, IP, Player Name, Account name. This can be extended by plugins.
+  * By default, bans are no longer removed upon expiry or 'deletion'. Instead, they remain in the system. A new ban for an indentifier can be added once an existing ban has expired.
+* Server Console now understands Terraria color codes (e.g., `[c/FF00FF:Words]`) and prints the colored text to the console. Note that console colors are limited and thus only approximations. (@QuiCM)
+* Fixed a bug in `/sudo` that prevented quoted arguments being forwarded properly. Example: `/sudo /user group "user name" "user group"` should now work correctly. (@QuiCM)
+* Shutting down the server should now correctly display the shutdown message to players rather than 'Lost connection'. (@QuiCM)
 
 ## TShock 4.4.0 (Pre-release 14)
 * Terraria v1.4.1.2 (Thanks @Patrikkk and @DeathCradle <3)
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1452,12 +1452,6 @@ namespace TShockAPI
 
 				var player = players[0];
 
-				if (player.HasPermission(Permissions.immunetoban))
-				{
-					args.Player.SendErrorMessage("That player is immune to bans.");
-					return;
-				}
-
 				string banReason = null;
 				void DoBan(string ident)
 				{
@@ -1478,23 +1472,23 @@ namespace TShockAPI
 				{
 					if (player.Account != null)
 					{
-						DoBan($"{Identifiers.Account}{player.Account.Name}");
+						DoBan($"{Identifier.Account}{player.Account.Name}");
 					}
 				}
 
 				if (banUuid)
 				{
-					DoBan($"{Identifiers.UUID}{player.UUID}");					
+					DoBan($"{Identifier.UUID}{player.UUID}");					
 				}
 
 				if (banName)
 				{
-					DoBan($"{Identifiers.Name}{player.Name}");
+					DoBan($"{Identifier.Name}{player.Name}");
 				}
 
 				if (banIp)
 				{
-					DoBan($"{Identifiers.IP}{player.IP}");
+					DoBan($"{Identifier.IP}{player.IP}");
 				}
 
 				//Using the ban reason to determine if a ban actually happened or not is messy, but it works
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1346,7 +1346,7 @@ namespace TShockAPI
 
 			void DisplayBanDetails(Ban ban)
 			{
-				args.Player.SendMessage($"{"Ban Details".Color(Utils.BoldHighlight)} - Unique ID: {ban.UniqueId.Color(Utils.GreenHighlight)}", Color.White);
+				args.Player.SendMessage($"{"Ban Details".Color(Utils.BoldHighlight)} - Ticket Number: {ban.TicketNumber.Color(Utils.GreenHighlight)}", Color.White);
 				args.Player.SendMessage($"{"Identifier:".Color(Utils.BoldHighlight)} {ban.Identifier}", Color.White);
 				args.Player.SendMessage($"{"Reason:".Color(Utils.BoldHighlight)} {ban.Reason}", Color.White);
 				args.Player.SendMessage($"{"Banned by:".Color(Utils.BoldHighlight)} {ban.BanningUser.Color(Utils.GreenHighlight)} on {ban.BanDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({ban.GetPrettyTimeSinceBanString().Color(Utils.YellowHighlight)} ago)", Color.White);
@@ -1426,7 +1426,7 @@ namespace TShockAPI
 					banResult = TShock.Bans.InsertBan(target, reason ?? "Banned", args.Player.Account.Name, DateTime.UtcNow, expiration);
 					if (banResult.Ban != null)
 					{
-						args.Player.SendSuccessMessage($"Ban ID {banResult.Ban.UniqueId} added.");
+						args.Player.SendSuccessMessage($"Ban added. Ticket Number: {banResult.Ban.TicketNumber}.");
 						DisplayBanDetails(banResult.Ban);
 					}
 					else
@@ -1458,7 +1458,7 @@ namespace TShockAPI
 					banResult = TShock.Bans.InsertBan(ident, reason, args.Player.Account.Name, DateTime.UtcNow, expiration);
 					if (banResult.Ban != null)
 					{
-						args.Player.SendSuccessMessage($"Ban ID {banResult.Ban.UniqueId} added for identifier {ident}.");
+						args.Player.SendSuccessMessage($"Ban Ticket Number {banResult.Ban.TicketNumber.Color(Utils.GreenHighlight)} created for identifier {ident}.");
 						banReason = banResult.Ban.Reason;
 					}
 					else
@@ -1587,7 +1587,7 @@ namespace TShockAPI
 				args.Player.SendWarningMessage($"If you are sure you wish to proceed, please execute {"ban-convert-confirm".Color(Utils.WhiteHighlight)} to continue.");
 				args.Player.AddResponse("ban-convert-confirm", (obj) =>
 				{
-					TShock.Bans.ConvertBans();
+					TShock.Bans.TryConvertBans();
 
 					var cmdArgs = (CommandArgs)obj;
 					cmdArgs.Player.SendSuccessMessage("Bans have been converted.");
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1271,15 +1271,15 @@ namespace TShockAPI
 			// ban add <target> [reason] [duration] [flags (default: -a -u -ip)]
 			//						Valid flags: -a (ban account name), -u (ban UUID), -n (ban character name), -ip (ban IP address), -e (exact, ban the identifier provided as 'target')
 			//						Unless -e is passed to the command, <target> is assumed to be a player or player index.
-			// ban del <target>
-			//						Target is expected to be an identifier in the format 'identifier_prefix:identifier'. Eg acc:MyAccountName
+			// ban del <ban ID>
+			//						Target is expected to be a ban Unique ID
 			// ban list [page]
 			//						Displays a paginated list of bans
-			// ban details <target>
-			//						Target is expected to be an identifier in the format 'identifier_prefix:identifier'. Eg acc:MyAccountName
-			//						Output: Banned Identifier - expiration
-			//								Reason: text
-			//								Banned by: name
+			// ban details <ban ID>
+			//						Target is expected to be a ban Unique ID
+			// ban convert
+			// ban-convert-confirm
+			//						Converts all old bans to new ban system.
 
 			void Help()
 			{
@@ -1291,11 +1291,11 @@ namespace TShockAPI
 
 				args.Player.SendMessage("TShock Ban Help", Color.White);
 				args.Player.SendMessage("Available Ban commands:", Color.White);
-				args.Player.SendMessage("ban [c/FFAAAA:add] <target> [flags]", Color.White);
-				args.Player.SendMessage("ban [c/FFAAAA:del] <target>", Color.White);
-				args.Player.SendMessage("ban [c/FFAAAA:list]", Color.White);
-				args.Player.SendMessage("ban [c/FFAAAA:details] <target>", Color.White);
-				args.Player.SendMessage("For more info, use [c/AAAAFF:ban help] [c/FFAAAA:command]", Color.White);
+				args.Player.SendMessage($"ban {"add".Color(Utils.RedHighlight)} <Target> [Flags]", Color.White);
+				args.Player.SendMessage($"ban {"del".Color(Utils.RedHighlight)} <Ban ID>", Color.White);
+				args.Player.SendMessage($"ban {"list".Color(Utils.RedHighlight)}", Color.White);
+				args.Player.SendMessage($"ban {"details".Color(Utils.RedHighlight)} <Ban ID>", Color.White);
+				args.Player.SendMessage($"For more info, use {"ban help".Color(Utils.BoldHighlight)} {"command".Color(Utils.RedHighlight)}", Color.White);
 			}
 
 			void MoreHelp(string cmd)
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming Changes
+* Overhauled Bans system. Bans are now based on 'identifiers'.(@QuiCM)
+  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Use `/ban convert` to convert to the new system. 
+  * All old ban routes in REST are now redirected. Please use `/v3/bans/*` for REST-based ban management.
+  * TShock recognizes and acts upon 4 main identifiers: UUID, IP, Player Name, Account name. This can be extended by plugins.
+  * By default, bans are no longer removed upon expiry or 'deletion'. Instead, they remain in the system. A new ban for an indentifier can be added once an existing ban has expired.
+* Server Console now understands Terraria color codes (e.g., `[c/FF00FF:Words]`) and prints the colored text to the console. Note that console colors are limited and thus only approximations. (@QuiCM)
```
hakusaro, line 21: Commit level: troll

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming Changes
+* Overhauled Bans system. Bans are now based on 'identifiers'.(@QuiCM)
```
```suggestion
* Overhauled Bans system. Bans are now based on 'identifiers'. (@QuiCM)
```

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
...
+  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Use `/ban convert` to convert to the new system.
```
hakusaro: If it's non-destructive, you should just auto-upgrade users.

Given the importance of the system (and given that I'm reading the changelog first), you should also consider doing something like https://github.com/github/scientist where you raise at least some flag somewhere if a user who's joining would be banned using the new system and not the old system, and vice versa. Not at all required though.

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming Changes
+* Overhauled Bans system. Bans are now based on 'identifiers'.(@QuiCM)
+  * The old Bans table (`Bans`) has been deprecated. New bans will go in `PlayerBans`. Use `/ban convert` to convert to the new system. 
+  * All old ban routes in REST are now redirected. Please use `/v3/bans/*` for REST-based ban management.
```
hakusaro, line 18: Seems good.

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
...
+  * TShock recognizes and acts upon 4 main identifiers: UUID, IP, Player Name, Account name. This can be extended by plugins.
+  * By default, bans are no longer removed upon expiry or 'deletion'. Instead, they remain in the system. A new ban for an indentifier can be added once an existing ban has expired.
```
hakusaro, line 20: Perfect.

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
...
+* Server Console now understands Terraria color codes (e.g., `[c/FF00FF:Words]`) and prints the colored text to the console. Note that console colors are limited and thus only approximations. (@QuiCM)
+* Fixed a bug in `/sudo` that prevented quoted arguments being forwarded properly. Example: `/sudo /user group "user name" "user group"` should now work correctly. (@QuiCM)
```
hakusaro, line 22: Awesome.

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
...
+* Shutting down the server should now correctly display the shutdown message to players rather than 'Lost connection'. (@QuiCM)
```
hakusaro, line 23: Superb.

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+				if (args.Parameters.Count > 1)
+				{
+					MoreHelp(args.Parameters[1].ToLower());
+					return;
+				}
 
-						// Effective ban target assignment
-						List<TSPlayer> players = TSPlayer.FindByNameOrID(args.Parameters[1]);
+				args.Player.SendMessage("TShock Ban Help", Color.White);
+				args.Player.SendMessage("Available Ban commands:", Color.White);
```
hakusaro, line 1293: I suggest adding at least one example of the most common type of ban here. When users are reading this, there's a good chance they're trying to ban someone _right now_, and now is not the ideal time for them to read more help pages. A simple example, e.g., `/ban add hackerperson hacking` or similar would suffice.

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
 
 		private static void Ban(CommandArgs args)
 		{
-			string subcmd = args.Parameters.Count == 0 ? "help" : args.Parameters[0].ToLower();
-			switch (subcmd)
+			//Ban syntax:
+			// ban add <target> [reason] [duration] [flags (default: -a -u -ip)]
```
hakusaro, line 1271: What're the units for duration?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-							return;
-						}
+			void MoreHelp(string cmd)
+			{
+				switch (cmd)
+				{
+					case "add":
+						args.Player.SendMessage("", Color.White);
+						args.Player.SendMessage("Ban Add Syntax", Color.White);
+						args.Player.SendMessage($"{"ban add".Color(Utils.BoldHighlight)} <{"Target".Color(Utils.RedHighlight)}> [{"Reason".Color(Utils.BoldHighlight)}] [{"Duration".Color(Utils.PinkHighlight)}] [{"Flags".Color(Utils.GreenHighlight)}]", Color.White);
```
hakusaro, line 1309: I really wish like, idk, I hate horizontal scrolling

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-							//Fail fast
-							args.Player.SendMultipleMatchError(players.Select(p => p.Name));
-							return;
-						}
+			void MoreHelp(string cmd)
+			{
+				switch (cmd)
+				{
+					case "add":
+						args.Player.SendMessage("", Color.White);
```
hakusaro, line 1307: Interesting -- why?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+						args.Player.SendMessage("Ban Add Syntax", Color.White);
+						args.Player.SendMessage($"{"ban add".Color(Utils.BoldHighlight)} <{"Target".Color(Utils.RedHighlight)}> [{"Reason".Color(Utils.BoldHighlight)}] [{"Duration".Color(Utils.PinkHighlight)}] [{"Flags".Color(Utils.GreenHighlight)}]", Color.White);
+						args.Player.SendMessage($"- {"Duration".Color(Utils.PinkHighlight)}: uses the format {"0d0m0s".Color(Utils.PinkHighlight)} to determine the length of the ban.", Color.White);
+						args.Player.SendMessage($"   Eg a value of {"10d30m0s".Color(Utils.PinkHighlight)} would represent 10 days, 30 minutes, 0 seconds.", Color.White);
```
hakusaro, line 1311: What about a permaban? How long is that?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-							}
+			void DisplayBanDetails(Ban ban)
+			{
+				args.Player.SendMessage($"{"Ban Details".Color(Utils.BoldHighlight)} - Unique ID: {ban.UniqueId.Color(Utils.GreenHighlight)}", Color.White);
+				args.Player.SendMessage($"{"Identifier:".Color(Utils.BoldHighlight)} {ban.Identifier}", Color.White);
+				args.Player.SendMessage($"{"Reason:".Color(Utils.BoldHighlight)} {ban.Reason}", Color.White);
+				args.Player.SendMessage($"{"Banned by:".Color(Utils.BoldHighlight)} {ban.BanningUser.Color(Utils.GreenHighlight)} on {ban.BanDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({ban.GetPrettyTimeSinceBanString().Color(Utils.YellowHighlight)} ago)", Color.White);
+				if (ban.ExpirationDateTime < DateTime.UtcNow)
+				{
+					args.Player.SendMessage($"{"Banned expired:".Color(Utils.BoldHighlight)} {ban.ExpirationDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({ban.GetPrettyExpirationString().Color(Utils.YellowHighlight)} ago)", Color.White);
```
hakusaro:
```suggestion
					args.Player.SendMessage($"{"Ban expired:".Color(Utils.BoldHighlight)} {ban.ExpirationDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({ban.GetPrettyExpirationString().Color(Utils.YellowHighlight)} ago)", Color.White);
```

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+					}
 
-							if (!parsedOkay)
-							{
-								args.Player.SendErrorMessage("Invalid time format. Example: 10d 5h 3m 2s.");
-								args.Player.SendErrorMessage("Use 0 (zero) for a permanent ban.");
-								return;
-							}
-						}
+					args.Player.SendMessage($"{"Banned expires:".Color(Utils.BoldHighlight)} {ban.ExpirationDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({remaining})", Color.White);
```
hakusaro:
```suggestion
					args.Player.SendMessage($"{"Ban expires:".Color(Utils.BoldHighlight)} {ban.ExpirationDateTime.ToString("yyyy/MM/dd").Color(Utils.RedHighlight)} ({remaining})", Color.White);
```

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
 
-							targetGeneralizedName = target.Name;
-							success = TShock.Bans.AddBan(target.IP, target.Name, target.UUID, target.Account?.Name ?? "", banReason, false, args.Player.Account.Name,
-								banLengthInSeconds == 0 ? "" : DateTime.UtcNow.AddSeconds(banLengthInSeconds).ToString("s"));
+				//This is hacky, but because the flag values can go at any parameter, this is forcing the ordering of 'reason' and 'duration'
+				//while still allowing them to be arbitrarily placed in the parameter list
+				for (int i = 2; i < args.Parameters.Count; i++)
+				{
+					var param = args.Parameters[i];
+					if (param != "-e" && param != "-a" && param != "-u" && param != "-n" && param != "-ip")
```
hakusaro: Could store this in an array so you can just check to see if param is in the array or not, rather than copypasta twice.

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-							}
+				string reason = "Banned.";
+				string duration = null;
+				DateTime expiration = DateTime.MaxValue;
+				AddBanResult banResult;
 
-							targetGeneralizedName = target.Name;
-							success = TShock.Bans.AddBan(target.IP, target.Name, target.UUID, target.Account?.Name ?? "", banReason, false, args.Player.Account.Name,
-								banLengthInSeconds == 0 ? "" : DateTime.UtcNow.AddSeconds(banLengthInSeconds).ToString("s"));
+				//This is hacky, but because the flag values can go at any parameter, this is forcing the ordering of 'reason' and 'duration'
```
hakusaro: So the purpose here is to allow the flags to go in any order, and then make duration and reason order dependent? If so, you should like, reword to something that emphasizes the _goal_.

```suggestion
				//This is hacky, because we want the flag values to be at any position, but this forces us to order reason and duration consecutively
```

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+				//while still allowing them to be arbitrarily placed in the parameter list
```
hakusaro:
```suggestion
				
```

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-								}
-							}
+				//If no flags were specified, default to account, uuid, and IP
+				if (!exactTarget && !banAccount && !banUuid && !banName && !banIp)
+				{
+					banAccount = banUuid = banIp = true;
+				}
 
-						}
+				if (exactTarget)
```
hakusaro, line 1472: I believe at this stage, if we have an exact target, we can be in a situation where no other flags were provided, in which case the defaults weren't set.

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
 
 		private static void Ban(CommandArgs args)
 		{
-			string subcmd = args.Parameters.Count == 0 ? "help" : args.Parameters[0].ToLower();
-			switch (subcmd)
+			//Ban syntax:
+			// ban add <target> [reason] [duration] [flags (default: -a -u -ip)]
+			//						Valid flags: -a (ban account name), -u (ban UUID), -n (ban character name), -ip (ban IP address), -e (exact, ban the identifier provided as 'target')
+			//						Unless -e is passed to the command, <target> is assumed to be a player or player index.
```
hakusaro, line 1274: If -e is passed, how do you specify a target?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-						}
+				if (player.HasPermission(Permissions.immunetoban))
+				{
+					args.Player.SendErrorMessage("That player is immune to bans.");
+					return;
+				}
 
-						return;
+				string banReason = null;
+				void DoBan(string ident)
```
hakusaro: `Banish`? lol

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-				case "del":
-					#region Delete ban
+					else
 					{
-						if (args.Parameters.Count != 2)
-						{
-							args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}ban del <player>", Specifier);
-							return;
-						}
+						args.Player.SendWarningMessage($"Ban skipped for identifier: {ident}");
```
hakusaro:
```suggestion
						args.Player.SendWarningMessage($"Ban failed for identifier: {ident}");
```

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+				//If no flags were specified, default to account, uuid, and IP
+				if (!exactTarget && !banAccount && !banUuid && !banName && !banIp)
+				{
+					banAccount = banUuid = banIp = true;
+				}
 
-						}
+				if (exactTarget)
+				{
+					banResult = TShock.Bans.InsertBan(target, reason ?? "Banned", args.Player.Account.Name, DateTime.UtcNow, expiration);
```
hakusaro: I haven't yet gotten to `InsertBan`, but am I correct in assuming that if a player is banned using an `exact` ban while online, they won't be booted?

hakusaro: Does this duplicate `DoBan`?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+				{
+					DoBan($"{Identifiers.Name}{player.Name}");
+				}
+
+				if (banIp)
+				{
+					DoBan($"{Identifiers.IP}{player.IP}");
+				}
+
+				//Using the ban reason to determine if a ban actually happened or not is messy, but it works
```
hakusaro: o.o

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+				if (banReason != null)
+				{
+					player.Disconnect($"You have been banned: {banReason}.");
```
hakusaro: ForceKick? Does that even do anything different?

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
-						{
-							"add <target> <time> [reason] - Bans a player or user account if the player is not online.",
-							"del <player> - Unbans a player.",
-							"delip <ip> - Unbans an IP.",
-							"list [page] - Lists all player bans.",
-							"listip [page] - Lists all IP bans."
-						};
+				if (TShock.Bans.RemoveBan(banId))
+				{
+					args.Player.SendSuccessMessage($"Ban {banId} has now been marked as expired.");
```
hakusaro: Could be something we'd want to log somewhere (e.g., "who removed this ban?")

TShockAPI/Commands.cs
```diff
@@ -1267,328 +1267,368 @@ private static void Kick(CommandArgs args)
...
+						break;
 
-						// Storage variable to determine if the command executor is the server console
-						// If it is, we assume they have full control and let them override permission checks
-						bool callerIsServerConsole = args.Player is TSServerPlayer;
+					case "list":
+						args.Player.SendMessage("", Color.White);
+						args.Player.SendMessage("Ban List Syntax", Color.White);
+						args.Player.SendMessage($"{"ban list".Color(Utils.BoldHighlight)} [{"Page".Color(Utils.PinkHighlight)}]", Color.White);
+						args.Player.SendMessage("- Lists active bans. Color trends towards green as the ban approaches expiration", Color.White);
```
hakusaro, line 1331: What about the inactive bans?

TShockAPI/DB/BanManager.cs
```diff
@@ -39,15 +70,13 @@ public BanManager(IDbConnection db)
 		{
 			database = db;
 
-			var table = new SqlTable("Bans",
-									new SqlColumn("IP", MySqlDbType.String, 16) { Primary = true },
-									new SqlColumn("Name", MySqlDbType.Text),
-									new SqlColumn("UUID", MySqlDbType.Text),
+			var table = new SqlTable("PlayerBans",
+									new SqlColumn("Id", MySqlDbType.Int32) { Primary = true, AutoIncrement = true },
+									new SqlColumn("Identifier", MySqlDbType.Text),
```
hakusaro, line 75: Perhaps call this something other than `Identifier` because `Id` can be thought of as shortened "identifier." On `	args.Player.SendErrorMessage("No ban found matching the given identifier");`, `identifier` was used to refer to the `Id` column, too.

TShockAPI/DB/BanManager.cs
```diff
@@ -62,231 +91,319 @@ public BanManager(IDbConnection db)
...
+
+		internal void BanValidateCheck(object sender, BanEventArgs args)
+		{
+			//Only perform validation if the event has not been cancelled before we got here
+			if (args.Valid)
 			{
-				TShock.Log.Error(ex.ToString());
-				Console.WriteLine(ex.StackTrace);
+				//We consider a ban to be valid if the start time is before now and the end time is after now, and the player is not immune to bans
+				args.Valid = (DateTime.UtcNow > args.Ban.BanDateTime && DateTime.UtcNow < args.Ban.ExpirationDateTime) && !args.Player.HasPermission(Permissions.immunetoban);
```
hakusaro: Can you use an exact target to ban an immune player?

TShockAPI/DB/BanManager.cs
```diff
@@ -62,231 +91,319 @@ public BanManager(IDbConnection db)
...
+				BanningUser = banningUser,
+				BanDateTime = fromDate,
+				ExpirationDateTime = toDate
+			};
+
+			OnBanPreAdd?.Invoke(this, args);
+
+			if (!args.Valid)
+			{
+				string message = $"Ban was invalidated: {(args.Message ?? "no further information provided.")}";
```
hakusaro:
```suggestion
				string message = $"Ban was not valid: {(args.Message ?? "no further information provided.")}";
```

TShockAPI/DB/BanManager.cs
```diff
@@ -62,231 +91,319 @@ public BanManager(IDbConnection db)
...
-		/// <param name="uuid">UUID.</param>
-		/// <param name="reason">Reason.</param>
-		/// <param name="exceptions">If set to <c>true</c> enable throwing exceptions.</param>
-		/// <param name="banner">Banner.</param>
-		/// <param name="expiration">Expiration date.</param>
-		public bool AddBan(string ip, string name = "", string uuid = "", string accountName = "", string reason = "", bool exceptions = false, string banner = "", string expiration = "")
+		/// <param name="identifier">Identifier to search with</param>
+		/// <param name="currentOnly">Whether or not to exclude expired bans</param>
+		/// <returns></returns>
+		public IEnumerable<Ban> RetrieveBansByIdentifier(string identifier, bool currentOnly = true)
```
hakusaro, line 348: What if I pass `Id`? That screams that it's a valid identifier to me...

TShockAPI/DB/BanManager.cs
```diff
@@ -62,231 +91,319 @@ public BanManager(IDbConnection db)
...
+			using (var reader = database.QueryReader(query, identifier))
 			{
-				if (exceptions)
-					throw ex;
-				TShock.Log.Error(ex.ToString());
+				while (reader.Read())
+				{
+					var uniqueId = reader.Get<int>("Id");
+					var ident = reader.Get<string>("Identifier");
+					var id = reader.Get<string>("Identifier");
```
hakusaro: Both `id` and `ident` are set to the same thing here.

TShockAPI/DB/BanManager.cs
```diff
@@ -330,122 +441,129 @@ public enum BanSortMethod
...
-				if (y == null)
-				{
-					//If both bans have no BanDateTime they're considered equal
-					return 0;
-				}
-				//If we're sorting by a newest to oldest method, a null value will come after the valid value.
-				return _method == BanSortMethod.AddedNewestToOldest || _method == BanSortMethod.ExpirationSoonestToLatest ? 1 : -1;
-			}
+	/// <summary>
+	/// Event args used for formalized bans
```
hakusaro: Formalized?

TShockAPI/DB/BanManager.cs
```diff
@@ -330,122 +441,129 @@ public enum BanSortMethod
...
+	public static class Identifiers
 	{
 		/// <summary>
-		/// Gets or sets the IP address of the ban entry.
+		/// IP identifier prefix constant
 		/// </summary>
-		/// <value>The IP Address</value>
-		public string IP { get; set; }
-
+		public const string IP = "ip:";
```
hakusaro: Okay, I officially think these should be in the help menu somewhere.

TShockAPI/DB/BanManager.cs
```diff
@@ -330,122 +441,129 @@ public enum BanSortMethod
...
+	/// </summary>
+	public class Ban
+	{
 		/// <summary>
-		/// Gets or sets the Client UUID of the ban
+		/// A unique ID assigned to this ban
 		/// </summary>
-		/// <value>The UUID</value>
-		public string UUID { get; set; }
+		public int UniqueId { get; set; }
```
hakusaro: Maybe call this a ticket number?

TShockAPI/Extensions/DbExt.cs
```diff
@@ -156,10 +185,10 @@ public static SqlType GetSqlType(this IDbConnection conn)
 				typeof (Int32?),
 				(s, i) => s.IsDBNull(i) ? null : (object)s.GetInt32(i)
 			},
-			{
+			/*{
```
hakusaro: ?

TShockAPI/GetDataHandlers.cs
```diff
@@ -3592,7 +3592,7 @@ private static bool HandleSyncExtraValue(GetDataHandlerArgs args)
 				return true;
 			}
 
-			return false;
+			return true;
```
hakusaro: Wait, why?

hakusaro (commented):
This is a really really good improvement. I left comments asking questions here & there where I got confused. You don't need to address things really -- I don't think I see any show stoppers. I do see some potential confusion between using `identifier` and a `database ID that auto increments and is the primary key` though. I think it's worth renaming the DB column to a `ticket number` or something, to be less confusing to people. Users can specify `identifiers` which are criteria to identify someone who is banned, but the primary user interface requires users to pass around `database IDs`, and I can see this easily confusing someone.

I think we should consider just throwing away the concept of being "immune to ban." In practice, you can ban anyone with an IP. It's probably better to advise users not to give ban permissions to untrusted people, or restrict who can ban by what (maybe only some people can ban using an exact target, and that's a separate permission?). Finally, we don't consider other fancy things like wildcard banning or CIDR range banning, which might also be worth considering at some point.

Overall though, I'd say that any love given to this system is good. Thank you so much for working on it!!

TShockAPI/GetDataHandlers.cs
QuiCM: Oops, that wasn't meant to be in this particular PR

CHANGELOG.md
QuiCM: > 
> 
> This is a really really good improvement. I left comments asking questions here & there where I got confused. You don't need to address things really -- I don't think I see any show stoppers. I do see some potential confusion between using `identifier` and a `database ID that auto increments and is the primary key` though. I think it's worth renaming the DB column to a `ticket number` or something, to be less confusing to people. Users can specify `identifiers` which are criteria to identify someone who is banned, but the primary user interface requires users to pass around `database IDs`, and I can see this easily confusing someone.

This is a good point. The original plan was not to have an auto-inc column and for everything to be done around the identifiers. However this encountered an issue when I found that GUIDs are 128-bit integers in .NET, and I couldn't reasonably expect people to type out `ban del uuid:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...etc`.
I like the idea of 'ticket number' as opposed to an arbitrary 'unique ID'. Will adjust the naming with that in mind

> 
> I think we should consider just throwing away the concept of being "immune to ban." In practice, you can ban anyone with an IP. It's probably better to advise users not to give ban permissions to untrusted people, or restrict who can ban by what (maybe only some people can ban using an exact target, and that's a separate permission?). 

This is fair enough. I had this thought too when I re-added the ban immunity, because it technically only works if you ban on a player target. You could very easily do `ban add acc:hakusaro -e` and it wouldn't check if account 'hakusaro' was immune to bans.

>
>Finally, we don't consider other fancy things like wildcard banning or CIDR range banning, which might also be worth considering at some point.

I agree - I didn't go too in depth with expanding on identifiers we already exist. However, because of the flexibility of arbitrary identifiers, plugins should easily be able to add, validate, and enforce their own identifiers - all using the same database & structure as the vanilla TShock system. Theoretically it shouldn't be too difficult to add something like `cidr:127.0.0.1/24` as a valid identifier and check it on user join. I also attempted to design around TShock not enforcing anything if other plugins get in first. I may implement an example extension plugin just to see how simple or complex it is to extend the system

TShockAPI/Extensions/DbExt.cs
QuiCM: This is an interesting one. Our defined override for Int64 was throwing exceptions. After extending `Get<T>` to hypothetically work on anything as long as you pass the right type, problem went away. I may need to retest this

TShockAPI/Commands.cs
QuiCM: Was to split up messages a bit. Despite the snazzy colours, if you use a bunch of text-heavy commands in a row (eg `ban help`, `ban help add`) you end up with a big wall of text

TShockAPI/Commands.cs
QuiCM: If `-e` is passed, the `target` parameter is treated directly as an identifier rather than a player.
E.g., `ban add ip:127.0,0,1 -e` would add a ban for the identifier `ip:127.0.0.1`.
If we are not using `-e`, the `target` parameter is used to attempt to identify an online player

TShockAPI/Commands.cs
QuiCM: Kick causes the 'Lost Connection' screen with worrying frequency at the moment, while Disconnect triggers a clean disconnect displaying the disconnection reason

TShockAPI/Commands.cs
QuiCM: >
>I haven't yet gotten to InsertBan, but am I correct in assuming that if a player is banned using an exact ban while online, they won't be booted?

Yep

>
>Does this duplicate `DoBan`?

Also yep

TShockAPI/Commands.cs
QuiCM: Added a 'quick usage' line: `Quick usage: ban add {player's name} "Griefing"`

TShockAPI/Commands.cs
QuiCM: Added a comment about durations

TShockAPI/Commands.cs
QuiCM: Me too. It's vaguely easier to read if I use raw colour codes (i.e., `"[c/FFAAAA:ban add] stuff"` but far more difficult to reuse. Splitting lines doesn't add much readability either :(

TShockAPI/Commands.cs
QuiCM: Added a line about permanent banning

TShockAPI/Commands.cs
QuiCM: Inactive bans are currently invisible unless you know the ticket number. Could add a parameter to include those in the list

TShockAPI/DB/BanManager.cs
QuiCM: Yep

TShockAPI/DB/BanManager.cs
QuiCM: Changed to 'completed'. Ban objects used in these eventargs are complete and have a valid ticket number

CHANGELOG.md
hakusaro: Yeah -- in general, immunity is something tba. If anything, we should warn people that it's not foolproof.

hakusaro (approved):
lgtm_6

In general, I think the changes you've made will help a lot! Everything looks good to me, though I didn't read it with a super fine tooth comb. I do really think the help for identifiers is top tier though, and shows that it's a good design. I honestly think this is a wicked improvement over what we've had before.

I totally understand how you got here, with respect to whether or not to use an autoincrementing ID. It makes total sense now. But yeah, all good now 👍

Tysm for working on this!! 💚

PR #2210, opened 2021-01-31
Add initial support for sentry raven

hakusaro:
This adds initial support for Sentry Raven, the goal being to both:

1. Find random exceptions.
2. Patch up Bouncer.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -880,6 +880,7 @@ namespace TShockAPI
 				else
 				{
 					TShock.Log.ConsoleDebug("Bouncer / OnNewProjectile please report to tshock about this! normally this is a reject from {0} {1}", args.Player.Name, type);
+					TShock.Utils.SentryLog("Bouncer / OnNewProjectile normal reject from {0}", type);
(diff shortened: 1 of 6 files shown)


QuiCM (commented):
Suggestion on implementation:
Rather than adding more statics to the `TShock` and `Utils` classes, could we instead create a `Sentry` class, singleton it, and keep implementation in there?

QuiCM:
> 
> 
> Suggestion on implementation:
> Rather than adding more statics to the `TShock` and `Utils` classes, could we instead create a `Sentry` class, singleton it, and keep implementation in there?

Something along the lines of the following:

```csharp
public static class Sentry
{
	internal static RavenClient client;


	static Sentry()
	{
		if (TShock.Config.Settings.SendExceptionsAndOtherDebugInformationToPryaxisFromSentry)
		{
			client = new RavenClient(...);
		}
	}

	static void Log(Exception ex)
	{
		client?.CaptureAsync(new SentryEvent(ex));
	}

	static void Log(string format, params object[] args)
	{
		client?.CaptureAsync(new SentryEvent(string.Format(format, args)));
	}
}
```

PR #2214, opened 2021-02-02
Fix ∞ velocity part 2 (thanks @tlworks, @bartico6)

hakusaro:
This may or may not solve some weird client crash issue? Not quite sure
because it seems to not be possible to reproduce the original report.
However, this *does* kick users who attempt to crash people with
infinite velocity, so there's no harm in shipping it.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -777,6 +777,30 @@ namespace TShockAPI
 			short type = args.Type;
 			int index = args.Index;
 
+			if (Single.IsInfinity(vel.X) || Single.IsInfinity(vel.Y))
+			{
+				TShock.Log.ConsoleInfo("Bouncer / OnNewProjectile force kicked (attempted to set velocity to infinity) from {0}", args.Player.Name);
+				args.Player.Kick("This Steam account has been banned from playing on secure servers due to a cheating infraction.", true, true);
+				args.Handled = true;
+				return;
+			}
+
+			if (Single.IsNaN(vel.X) || Single.IsNaN(vel.Y))
+			{
+				TShock.Log.ConsoleInfo("Bouncer / OnNewProjectile force kicked (attempted to set velocity to NaN) from {0}", args.Player.Name);
+				args.Player.Kick("This Steam account has been banned from playing on secure servers due to a cheating infraction.", true, true);
+				args.Handled = true;
+				return;
+			}
+
+			if (vel.X > 50000 || vel.Y > 50000 || vel.X < -50000 || vel.Y < -50000)
+			{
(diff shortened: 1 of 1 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -777,6 +777,30 @@ internal void OnNewProjectile(object sender, GetDataHandlers.NewProjectileEventA
...
+
+			if (Single.IsNaN(vel.X) || Single.IsNaN(vel.Y))
+			{
+				TShock.Log.ConsoleInfo("Bouncer / OnNewProjectile force kicked (attempted to set velocity to NaN) from {0}", args.Player.Name);
+				args.Player.Kick("This Steam account has been banned from playing on secure servers due to a cheating infraction.", true, true);
+				args.Handled = true;
+				return;
+			}
+
+			if (vel.X > 50000 || vel.Y > 50000 || vel.X < -50000 || vel.Y < -50000)
```
kevzhao2, line 796: Not sure if you care too much about this, but the conditionals can be simplified as:
```csharp
if (!(-50000 < vel.X && vel.X < 50000 && -50000 < vel.Y && vel.Y < 50000))
{
    // ...
}
```

This is because if X or Y are infinite, then one of the comparisons fails, and if X or Y are NaN, then all of the relevant comparisons fail.

bartico6: I was unsure how NaN interacts with comparisons but if that's the case then it makes sense to change it. I have to admit though that it's not very readable as a result.

kevzhao2: You could make an extension method like `IsInBounds`, and then now it's like this:
```csharp
if (!vel.X.IsInBounds(-5000, 5000) || !vel.Y.IsInBounds(-5000, 5000))
{
    // ...
}
```

bartico6: Well, if NaN and +/- infinity behave like one would expect (positive infinities are greater than any non-inf number and negative infinities smaller than any non-inf number, and all comparisons with NaN fail) then I'm all for solving it this way 🎉

bartico6: For what it's worth I applied something like this in my server engine, where I have extensions for asserting "reality" and bounds of vectors and floats. This 100% makes sense to do from code clarity perspective.

hakusaro: By marking this discussion as resolved, I hope to free you all from it. You’ve been chained to this for far too long!

Now that you’re free, you can read about the [traveling wave reactor](https://en.wikipedia.org/wiki/Traveling_wave_reactor)!

hakusaro:
I’m closing this because so far, I’ve seen no evidence or reproduction steps. The only thing that’s consistent is that @bartico6 can crash himself. While amusing, we can’t patch that in TShock.

PR #2219, opened 2021-02-08
Permission added to REST Server Broadcast .

TheBambino:
Nub trying this again.

**Feature:**
If a REST token is breached, there could be potential spam issues. Where each token perm could pose a potential risk, not having perms on ServerBroadcast makes it an issue if any token is breached. It also gives more control over those that have access to the token.

### DIFF
--- a/TShockAPI/Rest/RestPermissions.cs
+++ b/TShockAPI/Rest/RestPermissions.cs
@@ -67,6 +67,9 @@ namespace Rests
 		[Description("REST user can reload configurations, save the world and set auto save settings.")]
 		public const string restcfg = "tshock.rest.cfg";
 
+		[Description("REST user can turn off / restart the server.")]
+		public const string restbroadcast = "tshock.rest.broadcast";
+
 		[Description("REST user can kick players.")]
 		public const string restkick = "tshock.rest.kick";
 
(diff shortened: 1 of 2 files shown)


TShockAPI/Rest/RestPermissions.cs
```diff
@@ -67,6 +67,9 @@ public static class RestPermissions
 		[Description("REST user can reload configurations, save the world and set auto save settings.")]
 		public const string restcfg = "tshock.rest.cfg";
 
+		[Description("REST user can turn off / restart the server.")]
```
bartico6: Wrong description? :p

bartico6 (changes requested):
LGTM once the comments are addressed.

Also, should determine if this was intended behaviour or not.

TShockAPI/Rest/RestPermissions.cs
TheBambino: Yes, I caught that earlier. Changing it now

TheBambino:
> LGTM once the comments are addressed.

Done.

bartico6 (approved):
While I can't accept PRs (nor do my reviews mark it as ready for merge), I'll say that everything looks good now

hakusaro (dismissed):
Can you update the changelog?

QuiCM (approved):

PR #2249, opened 2021-03-21
Rewritten /grow; Adding /wallow

csmir:


QuiCM (approved):
This looks good to me. I may rework the /wallow command a little, but I think its acceptable in this state to be merged

PR #2266, opened 2021-04-05
Update to 1.4.2.1

Stealownz:


hakusaro (approved):

SignatureBeef (approved):
These changes look fine to me, lines up with the decompiled OTAPI packet and projectile source arg.
I have also been able to confirm i can run the artifacts and join as a super admin and projectiles etc seem to still be working.

PR #2269, opened 2021-04-11
Fix ban system conversion issue with MySQL

hakusaro:
DeathCradle spotted a typo in the ban converter and identified that the
issue is likely because we used table_name in some, but not all, of the
conversion SQL.


SignatureBeef (dismissed):
Agreed, this is exactly what i would have committed.

SignatureBeef:
IMO This should not be merged until PR #2271 is reviewed/merged.

PR #2275, opened 2021-04-16
Replace /godmode with Journey mode's godmode

### LINKED ISSUE #2209: Replace current Godmode (Healmode) with Journey mode Godmode
**Is your feature request related to a problem? Please describe.**
It's not related to a problem.
**Describe the solution you'd like**
Instead of continue using the inconsistent annoying healmode, implement actual journey mode godmode.
**Describe alternatives you've considered**
No alternative solution.
**Additional context**
No, it's not impossible because Quake already managed to do but refuses to share it.

Stealownz:
Closes #2209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Fixed server crash from `/v2/players/list` & other parameterised REST endpoints. (@QuiCM, reported by @ATFGK)
 * Added handling to the PlayerChat hook event. (@QuiCM - Thanks for the suggestion @Arthri)
 * Changed the spawnboss command to support silent command specifiers. (@QuiCM, suggested by @nojomyth-dev)
+* Changed `/godmode` to use Journey mode godmode. (@Stealownz, suggested by @tlworks, implementation hints from @0x3fcf1bbd)
 
 ## TShock 4.5.0.1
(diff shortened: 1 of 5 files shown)


Stealownz:
Important side-effect changes:
If a player has Journey godmode set enabled in local save and **does not** have permission `tshock.journey.godmode`,
when player joins the server will reject and broadcast send to disable Journey godmode.
They will also see the message "You are not allowed to toggle godmode"

However, if a player with Journey godmode set enabled in local save and **has** the permission `tshock.journey.godmode`,
then player will keep the Journey godmode state on join.
Note that this behavior is changed from TShock's godhealmode, which is always default false on player join

QuiCM:
Based on discussion in Discord, implementation LGTM. Has this been tested for any gotchas?

Stealownz:
Yes, tested two cases:

Tested for player with Journey godmode enabled in local save on join, see previous comment.

Tested for player joining into the index slot of a previous player with godmode enabled.
Issue: if a player in index slot 2 enables godmode then disconnects, another player who connects into index slot 2 is still considered godmode to the server and other players.
Expected behavior: the other player should not be godmode since it did not request to be godmode, and was not godmode enabled in the first place.
Solution: send disable godmode when player leaves if they were godmode,
and also check if player does not have permission, then reject and send a disable godmode

Stealownz:
Closing PR, found other creative powers also require handling on leave/join due to client/server mismatch on enabled states.
Will test and fix first

PR #2276, opened 2021-04-16
Update /godmode

bartico6:
- Previous Bouncer checks for GodMode (namely, Hurt) were removed.
- The command now uses the GodmodePower from core Terraria
- The toggle powers (which this command will now make use of) are now
  reset on disconnect to prevent accidentally "gifting" godmode to an
  unsuspecting player.

I'm the original copyright owner for the code in this commit (it is a backport from Frostspark.Server, which I am the sole author of, as mentioned in #2209)

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -37,6 +37,8 @@ using OTAPI.Tile;
 using TShockAPI.Localization;
 using System.Text.RegularExpressions;
 using Terraria.DataStructures;
+using Terraria.GameContent.Creative;
+using static Terraria.GameContent.Creative.CreativePowers;
 
 namespace TShockAPI
 {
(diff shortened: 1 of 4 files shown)


TShockAPI/Commands.cs
```diff
@@ -6445,6 +6447,10 @@ private static void ToggleGodMode(CommandArgs args)
 
 			playerToGod.GodMode = !playerToGod.GodMode;
 
+			var god_power = CreativePowerManager.Instance.GetPower<GodmodePower>();
```
QuiCM: Minor nitpick, could this be adjusted to `godPower` to keep style consistency?

bartico6: Yes, done & marking as resolved :)

QuiCM (dismissed):
LGTM

hakusaro (changes requested):
Please warn players that they need to disable god mode before disconnecting as part of the grant message when a user enables god mode. Also, please revise the message when a user joins without permission for god mode, and tell them that they had it on, and we turned it off because they didn't have permission.

hakusaro (dismissed):
lgtm_trick

hakusaro (approved):
lgtm_trick

QuiCM (approved):

PR #2300, opened 2021-05-12
Change max buff limit from short to int

moisterrific:
trying something pls ignore this for now

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5966,7 +5966,7 @@ namespace TShockAPI
 			}
 			int id = 0;
 			int time = 60;
-			var timeLimit = (int.MaxValue / 60) - 1
+			var timeLimit = (int.MaxValue / 60) - 1;
 			if (!int.TryParse(args.Parameters[0], out id))
 			{
 				var found = TShock.Utils.GetBuffByName(args.Parameters[0]);
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5985,7 +5985,7 @@ namespace TShockAPI
 				int.TryParse(args.Parameters[1], out time);
 			if (id > 0 && id < Main.maxBuffTypes)
 			{
-				if (time < 0 || time > short.MaxValue)
+				if (time < 0 || time > int.MaxValue)
 					time = 60;
 				args.Player.SetBuff(id, time * 60);
 				args.Player.SendSuccessMessage(string.Format("You have buffed yourself with {0}({1}) for {2} seconds!",
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -539,7 +539,7 @@ namespace TShockAPI
 			});
 			add(new Command(Permissions.buffplayer, GBuff, "gbuff", "buffplayer")
 			{
-				HelpText = "Gives another player a buff for an amount of time."
+				HelpText = "Gives another player a buff or debuff for an amount of time. Putting -1 for time will set it to 415 days."
 			});
 			add(new Command(Permissions.godmode, ToggleGodMode, "godmode")
 			{
@@ -6001,11 +6001,12 @@ namespace TShockAPI
 		{
 			if (args.Parameters.Count < 2 || args.Parameters.Count > 3)
 			{
-				args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}gbuff <player> <buff id/name> [time(seconds)]", Specifier);
+				args.Player.SendErrorMessage("Invalid syntax! Proper syntax: {0}gbuff <player> <buff name or ID> [time in seconds]", Specifier);
 				return;
 			}
 			int id = 0;
 			int time = 60;
+			var timeLimit = (int.MaxValue / 60) - 1;
 			var foundplr = TSPlayer.FindByNameOrID(args.Parameters[0]);
 			if (foundplr.Count == 0)
 			{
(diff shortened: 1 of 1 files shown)


bartico6:
This is technically possible as player buffs are sent as int32s now, not int16.
I'm nearly confident they were int16s at one point, though.

moisterrific:
> This is technically possible as player buffs are sent as int32s now, not int16.
> I'm nearly confident they were int16s at one point, though.

I just tested in-game and it doesn't ~work lol~ work the way i thought it would

moisterrific:
So after extensive trial and error testing, the max possible duration you can set to a buff is exactly `35791393` seconds. The text under the icon says that's approx. 415 days.

Terraria_2021_05_12_14_15_42_176 (2)

bartico6:
If you want to cap this at maximum, instead of hardcoding the value use the formula derived for it:

`var max_time = (int.MaxValue / 60) - 1`

Also keep default time (the one given if you give a bad time parameter) = 60, don't extend it into 415 days.

Other than that, LGTM

TShockAPI/Commands.cs
```diff
@@ -5985,10 +5986,11 @@ private static void Buff(CommandArgs args)
 				int.TryParse(args.Parameters[1], out time);
 			if (id > 0 && id < Main.maxBuffTypes)
 			{
-				if (time < 0 || time > short.MaxValue)
-					time = 60;
+				// Max possible buff duration as of 1.4.2.2 is 35791393 seconds (415 days).
+				if (time < 0 || time > timeLimit)
+					time = timeLimit;
```
QuiCM, line 5991: Agreed with Quake on this one, keep default time as 60 (approx 1 second)

QuiCM (changes requested):

hakusaro:
I guess I'm kinda confused. To me it looks like the intent is to set it as long as possible by @moisterrific. I don't think it's bad to let it go for 415 days if the game supports it.

TShockAPI/Commands.cs
moisterrific: Yes I've done that in the most recent commit. If the player didn't input any duration, the buff will default to 60 seconds. I wanted to make it easier to get the highest possible time so if you enter something like `-1` for duration it will give you 415 days. Also updated the HelpText to reflect this. 
`/buff 1`
`/buff 1 -1`
`/help buff`

Terraria_2021_05_13_08_54_53_977

Though I would argue it's not a big deal if you accidentally gave yourself a debuff for 415 days since you can just go to the Nurse to remove that or use `/kill`.

moisterrific:
> I guess I'm kinda confused. To me it looks like the intent is to set it as long as possible by @moisterrific. I don't think it's bad to let it go for 415 days if the game supports it.

yup that's what I intended at first, never understood why it defaulted to 60 sec

bartico6:
While talking to existing users I was told this is an existing use situation that they rely on the 60sec default while testing buffs, as to not remain stuck with the wrong buff and have to suicide, resetting all their buffs, not just the one they mistyped.
Maybe extending to 415d is the way but I'm not sure.

hakusaro:
I think this calls for...an option or a new command or a switch!

QuiCM (dismissed):

QuiCM:
Given there's no way to obtain 415 days of buff time legitimately in the game (as far as I know), I feel its a little over the top to *by default* provide 415 days of buff time upon using the command without arguments. Especially if its requires the player to die or visit an NPC to remove said buff.

Having a sane default and providing an option for the max time is a fair and reasonable solution in my opinion (and keeps the existing usage the same, with the addition of the max time option).
The `-1` value acts as a switch, so I don't think a new command is necessary.

Happy to merge this as-is, or continue to discuss

hakusaro:
Yeah my POV and probably @moisterrific’s is that the command implies that you’re turning it on, not applying the default game behavior. I’m more or less ambivalent though I think that supporting both works if that’s what @moisterrific wants

Killia0:
I'm not really a contributor to TShock but I second what @QuiCM said, getting the maximum time possible on accident feels like unexpected behavior, and if you need to add a new command or switch that is outside of scope of just allowing the values to be set to higher values. (at that point, if having a buff that lasts forever with less typing is desired why not build in a permabuff command/system to the tshock plugin where you can easily add and clear them)

bartico6:
For reference: In my earlier response, @Killia0 is one of the existing users mentioned.

I also personally strongly believe that not specifying a valid time (aka, /buff "glowing" or /buff 15) should give the buff for a fixed short interval, preferably the time a potion of that type would give you - if you do specify the time (especially type -1) the 415d variant should instead be used.

moisterrific:
> For reference: In my earlier response, @Killia0 is one of the existing users mentioned.
> 
> 
> 
> I also personally strongly believe that not specifying a valid time (aka, /buff "glowing" or /buff 15) should give the buff for a fixed short interval, preferably the time a potion of that type would give you - if you do specify the time (especially type -1) the 415d variant should instead be used.

I like this idea, but what about buffs that are not necessarily time based? Fallback to 60s?

Also do you think it makes more sense to change the max duration option from -1 to just 0?

bartico6:
> 
> 
> > For reference: In my earlier response, @Killia0 is one of the existing users mentioned.
> > I also personally strongly believe that not specifying a valid time (aka, /buff "glowing" or /buff 15) should give the buff for a fixed short interval, preferably the time a potion of that type would give you - if you do specify the time (especially type -1) the 415d variant should instead be used.
> 
> I like this idea, but what about buffs that are not necessarily time based? Fallback to 60s?
> 
> Also do you think it makes more sense to change the max duration option from -1 to just 0?

`-1` is commonly known amongst programmers as a "magic value" for `undefined` (and in terms of time periods - intuitive as infinity) - buffing myself for `0` seconds makes less sense to me than it would to buff myself for `-1` - but that's because of my programming background and familiarity with using -1 as such a value. Whether it makes sense to your average user - I don't know. An alternative time quasi-parameter named "max" or something would be a lot friendlier for sure.

Buffs that aren't time based? You mean.... ones unobtainable with potions, yes? Fall back to a sensible default, 60 would suffice.

**This all is my opinion, pls wait for others' inputs, especially Chris and shank since they're marked in this PR**

hakusaro:
Yeah I would go with -1 or 0. -1 is probably more intuitive to be infinity.

QuiCM (approved):

hakusaro (approved):

Issue #2305, opened 2021-05-14
Add a folder for logs by default


moisterrific:
**Is your feature request related to a problem? Please describe.**
It's annoying when log files fill up the tshock folder, esp when that folder is full of other important files users can write to, while logs are mostly for reading. 

**Describe the solution you'd like**
change `LogPath` in config from `tshock` to `logs` and generate a folder named logs (idk how that would work bc on my end I manually made a folder named logs)

**Describe alternatives you've considered**
no idea

**Additional context**
should look something like this

![image](https://user-images.githubusercontent.com/57187883/118267393-ffcbd600-b489-11eb-921d-76c409ba5648.png)

steviegt6:
Should this not be marked as closed now?

PR #2306, opened 2021-05-14
Change logpath to `./tshock/logs` by default

QuiCM:
Implements #2305, but uses `tshock/logs` as the log path, to keep  TShock stuff grouped still


moisterrific:
does it automatically generate the folder? I thought there would be more work required for that lol

QuiCM:
It auto-generates it :)

hakusaro (approved):

PR #2307, opened 2021-05-14
minor improvement to stock MOTD behavior

moisterrific:
Just in case users change the specifier, motd should also reflect any changes if they kept the `/help` line.

such a minor change so I didn't bother

### DIFF
--- a/TShockAPI/FileTools.cs
+++ b/TShockAPI/FileTools.cs
@@ -29,7 +29,7 @@ namespace TShockAPI
 	public class FileTools
 	{
 		private const string MotdFormat =
-			"This is [c/FF0000:%map%] on [c/00FFFF:TShock for Terraria].\n[c/00FF00:Current players:] [c/FFFF00:%players%]\nType [c/FF0000:%specifier%help] for a list of commands.\n";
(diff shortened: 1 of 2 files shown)


QuiCM (dismissed):

moisterrific:
Also kinda want to change a bit on how motd looks out of the box.

proposed 
vs.
current

Terraria_2021_05_14_13_14_04_016 (2)

I want to get rid of using the color red since that's usually associated with warnings or error messages throughout TShock. 
I think using the same yellow for `%map%` as `%players%` bc many info messages and the periodic world saving msg broadcast also use this color. 
Also added some subtle gradients.

QuiCM:
I don't see an image for 'proposed'

bartico6:
> 
> 
> I don't see an image for 'proposed'

I think it's the first and second motd on the same image that is the proposed and current

QuiCM:
> > I don't see an image for 'proposed'
> 
> I think it's the first and second motd on the same image that is the proposed and current

Oh yeah, you right. Thanks

QuiCM (approved):

hakusaro (approved):

PR #2308, opened 2021-05-14
Minor improvements to /renameNpc

moisterrific:
Improved user feedback clarity and added silent specifier support.

yes

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5813,7 +5813,7 @@ namespace TShockAPI
 			{
 				if (args.Silent)
 				{
-					TSPlayer.All.SendInfoMessage("Renamed the {0}.", npcType);
+					args.Player.SendSuccessMessage("You renamed the {0}.", npcType);
 				}
 				else
 				{
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5787,7 +5787,7 @@ namespace TShockAPI
 					args.Player.SendMultipleMatchError(npcs.Select(n => $"{n.FullName}({n.type})"));
 					return;
 				}
-				else if (args.Parameters[1].Length > 200)
+				else if (args.Parameters[1].Length >= 200)
 				{
 					args.Player.SendErrorMessage("New name is too long! Please keep it under 200 characters.");
 					return;
@@ -5822,7 +5822,7 @@ namespace TShockAPI
 			}
 			else
 			{
-				args.Player.SendErrorMessage("Cannot rename that NPC!");
+				args.Player.SendErrorMessage("Unable to rename that NPC!");
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


moisterrific:
Is it necessary to make the message more specific? 
Instead of `Player renamed the Guide.`, should I make it like `Player renamed Andrew the Guide to John the Guide.`?
Or `Player renamed the Guide to John.`?

Also is this supposed to only work on town NPCs? I was kind of confused at the conditions here:
https://github.com/Pryaxis/TShock/blob/356cdaa7ad710974791bdb9dd5fee9a2dfd477ab/TShockAPI/Commands.cs#L5803

TShockAPI/Commands.cs
```diff
@@ -5798,22 +5798,31 @@ private static void RenameNPC(CommandArgs args)
...
 					done++;
+					npcType = Main.npc[i].TypeName;
 				}
 			}
 			if (done > 0)
 			{
-				TSPlayer.All.SendInfoMessage("{0} renamed the {1}.", args.Player.Name, args.Parameters[0]);
+				if (args.Silent)
+				{
+					TSPlayer.All.SendInfoMessage("Renamed the {0}.", npcType);
```
hakusaro: Can you not send a message if the command is silent?

hakusaro (commented):
I think we should at least more or less try to be consistent: silent should mean silent imho.

moisterrific:
> I think we should at least more or less try to be consistent: silent should mean silent imho.

Right, I'll change it soon.

Got confused at the time what silent was really supposed to do.

TShockAPI/Commands.cs
moisterrific: just to make sure, silent = only the command user sees a success message?

hakusaro: Ya.

TShockAPI/Commands.cs
```diff
@@ -5789,7 +5789,7 @@ private static void RenameNPC(CommandArgs args)
 				}
 				else if (args.Parameters[1].Length > 200)
```
steviegt6: Wouldn't this make the limit <= 200, not < 200?

bartico6: Correct, if it must be below 200, then the length test should be >= 200

moisterrific: yeah that makes sense, i somehow overlooked that lol

steviegt6: Also, is there any actual reason for the limit, or is it arbitrary?

moisterrific: I think it's to prevent crashing?

moisterrific: gonna comment out that check on my local build and see what happens

bartico6 (commented):
Sincere apologies for late response - LGTM. I'm only iffy about the 200 character limit, as I'm not aware of any such limit in the game code, but to be honest, 200 characters sounds like a reasonable cap.

**And I misclicked, this is supposed to be an approving review, not a comment xD**

moisterrific:
kinda confused on one thing, is this command only intended to be used on townNPCs?

PR #2329, opened 2021-05-21
change codename to Blood Moon edition

moisterrific:
image

We'll be seeing this on [Wednesday, May 26](https://en.wikipedia.org/wiki/May_2021_lunar_eclipse)


hakusaro:
Thanks for this! I made the change, unfortunately, before you sent this, because we were originally planning on publishing the security advisory in one-go. Unfortunately, the documentation for the security advisory feature isn't great so this didn't happen until after we merged. Thank you for your suggestion!! See: https://github.com/Pryaxis/TShock/commit/9ff3036469b373beb009a1f4891b72e874985a75

PR #2337, opened 2021-05-23
Fix torchflags

Quinci135:
Migrated from https://github.com/Quinci135/TShock/commit/8ac4d2e68518d5fd337f1fe5a8a1859420701c22
`UsingBiomeTorches`: Whether or not the player has the torchgod biometorches ability enabled
`HappyFunTorchTime`: Whether or not the player has fought the torchgod before (for logic that checks for torchgod spawning)
`unlockedBiomeTorches`: Whether or not the player has the torchgod biome torches ability unlocked

### DIFF
--- a/TShockAPI/PlayerData.cs
+++ b/TShockAPI/PlayerData.cs
@@ -49,6 +49,8 @@ namespace TShockAPI
 		public Color? eyeColor;
 		public bool[] hideVisuals;
 		public int questsCompleted;
+		public int usingBiomeTorches;
+		public int happyFunTorchTime = 1;
 		public int unlockedBiomeTorches;
 
 		public PlayerData(TSPlayer player)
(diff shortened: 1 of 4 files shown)


TShockAPI/PlayerData.cs
```diff
@@ -49,6 +49,8 @@ public class PlayerData
 		public Color? eyeColor;
 		public bool[] hideVisuals;
 		public int questsCompleted;
+		public int usingBiomeTorches;
+		public int happyFunTorchTime = 1;
```
bartico6: I'm not fully aware how this value usually looks - is this intended to be initialised to 1 by default?
Your comment would suggest otherwise:  
`Whether or not the player has fought the torchgod before (for logic that checks for torchgod spawning)`

bartico6 (commented):
Pending answer to my comment I'd say it preemptively looks good :)

TShockAPI/PlayerData.cs
Quinci135: Yeah, should be false. Think I got confused when reading the code and thought of a context of: happyFunTorchTime indicates relogic is going to give you a happy fun time with torches assaulting you underground, so we are marking you as a potential victim. In fact, it's the other way around: you get happyFunTorchTime set to true once you've obtained the torch god's favor item and then you can use the magic biome torches ability.

QuiCM (approved):

bartico6 (approved):
Looks good now 👍

PR #2346, opened 2021-05-26
Change TSPlayer.FindByNameOrID to keep searching

hakusaro:
Currently, the TSPlayer FindbyNameOrID method aborts if it finds an
"exact match" based on this criteria:

1. If the player ID is on the server, it must be the thing we're looking
   for. Therefore, return that.

2. If the case sensitive "exact match" is on the server that isn't an
   ID, that must be what we're looking for. Therefore, return that.

3. Just yolo and downcase everything and return any number of matching
   players next.

This commit changes the behavior because some players have been joining
servers with ambiguous names, like `1`. In the current system, this
player is difficult to query because they're an "ID" and therefore an
exact match will be returned even if a player name exists that matches
the criteria.

This also alleviates the issue of a case exact match falling down the
same trap. It's ambiguous enough in all of these situations that an
admin should just be using a player ID instead.`


hakusaro:
The only downside I can see here is that we're changing the existing API and some people might be depending on this weird functionality. BUT I just don't think that they should be expecting the list of found players to always return one player. It's a list for a reason.

Ijwu (approved):
LGTM

PR #2351, opened 2021-05-27
Move emoji player index check into IllegalPerSe

hakusaro:
This is the first commit in a pattern that I'd like to follow. The
concept is that we specifically create handlers for things that are
"illegal per se." That is, there are no possible situations (in the
current protocol) where a packet of this type is received from a client.
In this case, I moved the emoji handler out of the Handler just for
emoji, since it seemed like an obvious case.

The rule of thumb is simple: if something is illegal per se, there
should be no possible way in the vanilla client to achieve this result.
If a player sends this combination of packets they *must* be hacking.
Not that there is a 99.9% chance they're hacking, but that there is a
100% unambiguous chance that they're hacking.

Something is illegal per se if it can only be created by a hacked
client. If there's a crashing bug that a normal player can do with a
complex series of vanilla events, that is not illegal per se.

The goal of this namespace and class of handlers is to handle exactly
one type of protocol violation, and remove the packet accordingly. If it
is ever reported that the packet can be sent from a vanilla client, the
check must be removed as it is no longer a per se violation of the
protocol.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -36,14 +36,14 @@ namespace TShockAPI
 	/// <summary>Bouncer is the TShock anti-hack and anti-cheat system.</summary>
 	internal sealed class Bouncer
 	{
-		internal Handlers.SendTileRectHandler STSHandler { get; set; }
-		internal Handlers.NetModules.NetModulePacketHandler NetModuleHandler { get; set; }
-		internal Handlers.EmojiHandler EmojiHandler { get; set; }
-		internal Handlers.IllegalPerSe.EmojiPlayerMismatch EmojiPlayerMismatch { get; set; }
-		internal Handlers.DisplayDollItemSyncHandler DisplayDollItemSyncHandler { get; set; }
-		internal Handlers.RequestTileEntityInteractionHandler RequestTileEntityInteractionHandler { get; set; }
-		internal Handlers.LandGolfBallInCupHandler LandGolfBallInCupHandler { get; set; }
-		internal Handlers.SyncTilePickingHandler SyncTilePickingHandler { get; set; }
+		internal Handlers.SendTileRectHandler STSHandler { get; private set; }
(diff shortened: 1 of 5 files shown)


TShockAPI/Handlers/IllegalPerSe/EmojiPlayerMismatch.cs
```diff
@@ -0,0 +1,27 @@
+using static TShockAPI.GetDataHandlers;
+
+namespace TShockAPI.Handlers.IllegalPerSe
+{
+	/// <summary>
+	/// Rejects emoji packets with mismatched identifiers
+	/// </summary>
+	/// <param name="sender"></param>
```
kevzhao2: spurious?

TShockAPI/Bouncer.cs
```diff
@@ -39,6 +39,7 @@ internal sealed class Bouncer
 		internal Handlers.SendTileRectHandler STSHandler { get; set; }
 		internal Handlers.NetModules.NetModulePacketHandler NetModuleHandler { get; set; }
 		internal Handlers.EmojiHandler EmojiHandler { get; set; }
+		internal Handlers.IllegalPerSe.EmojiPlayerMismatch EmojiPlayerMismatch { get; set; }
```
kevzhao2: can we make these all `private set` or something?

kevzhao2 (dismissed):
lgtm

TShockAPI/Bouncer.cs
hakusaro: Ya

TShockAPI/Handlers/IllegalPerSe/EmojiPlayerMismatch.cs
hakusaro: Smh

QuiCM (dismissed):

QuiCM:
UH OH! There's **CONFLICTS**

kevzhao2 (dismissed):

PR #2363, opened 2021-06-04
Improvements to Heal command

moisterrific:
yes

Based on [this suggestion](https://github.com/Pryaxis/TShock/discussions/2348)


bartico6 (commented):
The new syntax implies that you must specify how much health to restore, if healing another player.

The way to "give them full health like it always was" is not clear, because the heal amount parameter is first, rather than second, and it only heals the full health if the "amount" parameter is less than one, of above their max hp. This is not telegraphed to the user at all, meaning people who have previously used /heal will probably not know how to heal someone else for full health, which was previously trivial.

TL;DR - I think the syntax should be /heal [player] [amount] - healing for a fixed amount is less frequent than healing other players, and the argument order should reflect this in my opinion.
Healing yourself for a given amount should be /heal yourname amount - again, in my opinion. Not speaking on behalf of @Pryaxis/tshock, just myself.

TShockAPI/Commands.cs
```diff
@@ -5933,28 +6013,54 @@ private static void Heal(CommandArgs args)
...
+						int.TryParse(args.Parameters[0], out amount);
+						if (amount < 1 || amount > targetMaxHp)
+						{
+							amount = targetMaxHp - targetCurrentHp;
+							playerToHeal.Heal(amount);
+						}
+						else
+						{
+							playerToHeal.Heal(amount);
+						}
```
rautamiekka, line 6140: Why not do this:
```cs
						int.TryParse(args.Parameters[0], out amount);
						if (amount < 1 || amount > targetMaxHp)
						{
							amount = targetMaxHp - targetCurrentHp;
						}
						playerToHeal.Heal(amount);
```
instead of running the same line twice with the diff of 1 missing var assignment ?

moisterrific:
> The new syntax implies that you must specify how much health to restore, if healing another player.
> 
> The way to "give them full health like it always was" is not clear, because the heal amount parameter is first, rather than second, and it only heals the full health if the "amount" parameter is less than one, of above their max hp. This is not telegraphed to the user at all, meaning people who have previously used /heal will probably not know how to heal someone else for full health, which was previously trivial.
> 
> TL;DR - I think the syntax should be /heal [player] [amount] - healing for a fixed amount is less frequent than healing other players, and the argument order should reflect this in my opinion.
> Healing yourself for a given amount should be /heal yourname amount - again, in my opinion. Not speaking on behalf of @Pryaxis/tshock, just myself.

yeah you're right this makes more sense and would be more consistent with how slap works

hakusaro:
> yeah you're right this makes more sense and would be more consistent with how slap works

@moisterrific are you planning on getting to this? I'd love to merge this

x

moisterrific:
> 
> 
> > yeah you're right this makes more sense and would be more consistent with how slap works
> 
> @moisterrific are you planning on getting to this? I'd love to merge this
> 
> x

soon™

PR #2367, opened 2021-06-06
Add Obsidian vault for docs rewrite

hakusaro:
Because Readme.io has been kinda scummy in terms of "giving us the
features that we want" and because most wikis don't work well, I'm
proposing a hybrid approach: an Obsidian vault (which is plain markdown)
and having the contents published to Obsidian publish. The end result
will be nice a nice and easily navigable website that can be used to
reference TShock things without needing a separate interface to edit the
docs for.

Plus we can integrate with CI, etc.

Major things to do:

- [ ] Undo the changes to the dumper utilities so that plain markdown tables are generated
- [ ] Ideally supply things like CI to alert if things need to be updated (changed permissions file? maybe you should possibly have re-run the dumper and updated the permissions docs?)
- [ ] Get Obsidian publish and set it up
- [ ] Manually transfer over pages worth keeping
- [ ] Re-verify the protocol docs


drunderscore:
Looks like you added `docs/.obsidian/*` to the git ignore, but added and committed `docs/.obsidian/config` before then.

hakusaro:
@drunderscore  yes, intentionally

PR #2390, opened 2021-07-05
Add permission check for summoning Empress of Light + ForceTime check for using Enchanted Sundial

moisterrific:
Players without `summonboss` permission should not be able to kill Prismatic Lacewing because this summon the Empress of Light. (Also we won't have to worry about other ways to get around this since EoL can ONLY be summoned if the critter was killed by players)

As for **Enchanted Sundial**, without this check, the game would try to fast forward time and TShock will keep trying to set it back to day (or night), so just use your imagination on how janky this would look lol

yes

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2813,12 +2813,21 @@ namespace TShockAPI
 				return true;
 			}
 			
-			if (Main.npc[id].netID == NPCID.EmpressButterfly && !args.Player.HasPermission(Permissions.summonboss))
+			if (Main.npc[id].netID == NPCID.EmpressButterfly)
 			{
-				args.Player.SendErrorMessage("You do not have permission to summon the Empress of Light.");
-				args.Player.SendData(PacketTypes.NpcUpdate, "", id);
-				TShock.Log.ConsoleDebug($"GetDataHandlers / HandleNpcStrike rejected EoL summon from {args.Player.Name}");
-				return true;
+				if (!args.Player.HasPermission(Permissions.summonboss))
+				{
+					args.Player.SendErrorMessage("You do not have permission to summon the Empress of Light.");
+					args.Player.SendData(PacketTypes.NpcUpdate, "", id);
+					TShock.Log.ConsoleDebug($"GetDataHandlers / HandleNpcStrike rejected EoL summon from {args.Player.Name}");
+					return true;
+				}
+				else if (!TShock.Config.Settings.AnonymousBossInvasions)
+				{
+					TShock.Utils.Broadcast(string.Format($"{args.Player.Name} summoned the Empress of Light!"), 175, 75, 255);
+				}
+				else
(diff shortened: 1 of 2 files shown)


QuiCM (dismissed):
LGTM

TShockAPI/GetDataHandlers.cs
```diff
@@ -3201,11 +3208,20 @@ private static bool HandleSpecial(GetDataHandlerArgs args)
...
+				if (!args.Player.HasPermission(Permissions.usesundial))
+				{
+					TShock.Log.ConsoleDebug($"GetDataHandlers / HandleSpecial rejected enchanted sundial permission {args.Player.Name}");
+					args.Player.SendErrorMessage("You do not have permission to use the Enchanted Sundial.");
+					return true;
+				}
+				else if (TShock.Config.Settings.ForceTime != "normal")
+				{
+					TShock.Log.ConsoleDebug($"GetDataHandlers / HandleSpecial rejected enchanted sundial permission (ForceTime) { args.Player.Name}");
+					args.Player.SendErrorMessage($"You must set ForceTime to normal via config to use the Enchanted Sundial.");
```
bartico6: I think this could be made more user-friendly, as this message will appear for all players, not just admins. Maybe a "Time is currently frozen on this server, and the sundial cannot be used" type message with optional information how to disable it? idk

bartico6 (dismissed):
One small nitpick if you feel like addressing it (only if @QuiCM agrees), everything else LGTM.

TShockAPI/GetDataHandlers.cs
moisterrific: yea ur right that would be better for the avg player

would it make more sense to check if the player has `tshock cfg` permission and send the current msg to them only? 
its not like that info is useful for players who can use sundial but can't change the config

bartico6: Hm. This would be easier in the realm of flag-based admin tools - for instance, a SourceMod server would tell this configuration tidbit to any player with the `b` flag ("admin generic"), but for tshock,... I'm not sure what should determine adminship for the purpose of this message. Maybe access to /reload? cc @QuiCM @hakusaro

moisterrific: I could change it to something like this:
```csharp
if (player has config editing permission) 
{
   "ForceTime must be set to normal to use sundial"
}
else
{
   "time is frozen so u can't use sundial"
}
```
?

moisterrific: > Hm. This would be easier in the realm of flag-based admin tools - for instance, a SourceMod server would tell this configuration tidbit to any player with the `b` flag ("admin generic"), but for tshock,... I'm not sure what should determine adminship for the purpose of this message. Maybe access to /reload? cc @QuiCM @hakusaro

yea reload sounds good, sorry I thought there was some other perm related to the config lol

bartico6 (dismissed):
Looks great now, no complaints :)

QuiCM (approved):

PR #2397, opened 2021-07-13
Fix invalid group problems & error reporting on startup

### LINKED ISSUE #2396: Players with invalid group names are unable to join the server
* TShock version: Latest, diagnosed from source code.

**For registered users:**
- Delete the account's group from the storage provider without changing the account's group

**For unregistered users:**
- Rename the guest group in the config, and delete it (or rename to a non-existent group).

#### What should happen?

Graceful failure

#### What actually happens?

Handler for packet 8 throws an exception and the connection becomes stuck.

bartico6:
Server will no longer start up without the two essential groups (default and guest).

Additionally, players whose groups are removed in the database will no longer be stuck on connecting phase, and are instead disconnected with a descriptive message.

The server now also correctly reports startup errors in the console. The previous call to 'Error' instead of 'ConsoleError' caused the error to appear in the log, but not in the console, leading to a confusing situation for both users and developers where the server would randomly terminate mid-startup.

Fixes #2396


TShockAPI/Utils.cs
```diff
@@ -212,6 +212,27 @@ public int GetActivePlayerCount()
...
 
+		/// <summary>
+		/// Asserts that the group reference can be safely assigned to the player object.
+		/// <para>If this assertion fails, and <paramref name="kick"/> is true, the player is disconnected. If <paramref name="kick"/> is false, the player will receive an error message.</para>
+		/// </summary>
+		/// <param name="player"></param>
+		/// <param name="group"></param>
+		/// <returns></returns>
+		public bool AssertGroupValid(TSPlayer player, Group group, bool kick)
+        {
```
hakusaro: Can you indent this with tabs?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Correct rejection message in LandGolfBallInCupHandler to output the proper expected player id. (@drunderscore)
 * Clarified the error mesage that the console is presented if a rate-limit is reached over REST to indicate that "tokens" actually refers to rate-limit tokens, and not auth tokens, and added a hint as to what config setting determines this. (@hakusaro, @patsore)
 * Fixed an issue where, when the console was redirected, input was disabled and commands didn't work, in TSAPI. You can now pass `-disable-commands` to disable the input thread, but by default, it will be enabled. Fixes [#1450](https://github.com/Pryaxis/TShock/issues/1450). (@DeathCradle, @QuiCM)
+* Fixed missing group problems & error reporting on startup. (@bartico6)
```
hakusaro: Can you be as specific and descriptive as possible with how you did this?

TShockAPI/Utils.cs
```diff
@@ -212,6 +212,27 @@ public int GetActivePlayerCount()
...
 		}
 
+		/// <summary>
+		/// Asserts that the group reference can be safely assigned to the player object.
+		/// <para>If this assertion fails, and <paramref name="kick"/> is true, the player is disconnected. If <paramref name="kick"/> is false, the player will receive an error message.</para>
+		/// </summary>
+		/// <param name="player"></param>
+		/// <param name="group"></param>
+		/// <returns></returns>
+		public bool AssertGroupValid(TSPlayer player, Group group, bool kick)
```
hakusaro: Can you move this into the group manager place/thing? We're trying not to add anything else into utils, because it's kinda just the trash dumping ground. I don't want you to just chuck something in, because then it has to live with the other `Dump` methods, which really make this look kinda...redheaded.

TShockAPI/TShock.cs
```diff
@@ -322,6 +322,7 @@ public override void Initialize()
 				Regions = new RegionManager(DB);
 				UserAccounts = new UserAccountManager(DB);
 				Groups = new GroupManager(DB);
+				Groups.EnsureCoreGroupsPresent();
```
hakusaro: Couldn't this happen on the constructor for the GroupManager?

TShockAPI/DB/GroupManager.cs
```diff
@@ -202,6 +202,21 @@ public GroupManager(IDbConnection db)
...
+			{
+				TShock.Log.ConsoleError("The guest group could not be found. This may indicate a typo in the configuration file, or that the group was renamed or deleted.");
+				throw new Exception("The guest group could not be found.");
+			}
+
+			if(!GroupExists(TShock.Config.Settings.DefaultRegistrationGroupName))
+            {
+				TShock.Log.ConsoleError("The default usergroup could not be found. This may indicate a typo in the configuration file, or that the group was renamed or deleted.");
+				throw new Exception("The default usergroup could not be found.");
+            }
```
hakusaro: Can you indent using tabs here?

TShockAPI/DB/GroupManager.cs
```diff
@@ -202,6 +202,21 @@ public GroupManager(IDbConnection db)
...
+		internal void EnsureCoreGroupsPresent()
+		{
+			if (!GroupExists(TShock.Config.Settings.DefaultGuestGroupName))
+			{
+				TShock.Log.ConsoleError("The guest group could not be found. This may indicate a typo in the configuration file, or that the group was renamed or deleted.");
+				throw new Exception("The guest group could not be found.");
+			}
+
+			if(!GroupExists(TShock.Config.Settings.DefaultRegistrationGroupName))
+            {
```
hakusaro: Can you indent using tabs here?

hakusaro (changes requested):
I think this is a really good step to helping our support issues. I think, though, I'm a bit unsure if we should just have `Assert` and `Ensure` methods. `Ensure` seems to me like something that would create it if it didn't exist. I would prefer that `Ensure` did that, actually, else it should just be something like `Assert` just like the other method.

Fundamentally, though, I think if a player tries to join and their group is gone, you should also output an exception and tell someone how to resolve the issue, too.

Basically, I would say that this is good, but you'll probably want to convey more to the server operator. The `Ensure` method is more of a subset of `Assert` and they both try to accomplish similar things, so I'd suggest just combining them so that there's always an exception going to a server console. If the login fails because the group doesn't exist, that can be a separate thing.

Also, because this will probably cause someone some heart ache, I would suggest noting in the changelog exactly what will happen if groups are missing in which stages of the connection and startup process.

bartico6:
My sincere apologies for the delay, I'll get around to it now.

bartico6:
This should be good now, I made the changes you requested & resolved the merge conflict that came up.

QuiCM (dismissed):

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * `OnPlaceTileEntity`: The check was newly added.
   * `OnPlaceItemFrame`: The check was newly added.
   * `OnFoodPlatterTryPlacing`: The check was newly added.
+* Fixed errors on startup not being reported to console. (@bartico6)
+* The server now correctly disconnects players with missing groups instead of throwing an exception, stalling the connection (@bartico6)
```
hakusaro, line 54:
```suggestion
  * The server now correctly disconnects players with missing groups instead of throwing an exception, stalling the connection (@bartico6)
```

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
...
+* The server now rejects login attempts from players who would end up with a missing group. (@bartico6)
```
hakusaro, line 55:
```suggestion
  * The server now rejects login attempts from players who would end up with a missing group. (@bartico6)
```

TShockAPI/Commands.cs
```diff
@@ -823,10 +823,16 @@ private static void AttemptLogin(CommandArgs args)
...
 						!String.IsNullOrWhiteSpace(args.Player.UUID)))
 				{
-					args.Player.PlayerData = TShock.CharacterDB.GetPlayerData(args.Player, account.ID);
-
 					var group = TShock.Groups.GetGroupByName(account.Group);
 
+					if (!TShock.Utils.AssertGroupValid(args.Player, group, false))
+					{
+						args.Player.SendErrorMessage("Login attempt failed - see the message above.");
+						return;
```
hakusaro, line 840: At this return path, wouldn't the user be left in an inconsistent state since they've been connected but everything else is broken?

hakusaro (changes requested):
Hmm, this still adds a method to TShock.Utils, but you marked my last comment as resolved. So I unresolved it.

I'm not quite sure what the status is if a user joins in a bad state. I guess it sucks but it's better than kicking them?

Why did the submodule change?

TShockAPI/Commands.cs
bartico6: No, this is the login command. If they try to log in into an account that would break their groups this rejects the login attempt. But now I'm doubting my own code lol

bartico6:
I merged gendev because of a merge conflict so maybe that's why?
Also forgot to remove the method in utils after moving it to group manager it seems.

hakusaro:
> I merged gendev because of a merge conflict so maybe that's why?

photo_2021-08-02 22 50 39

hakusaro:
@bartico6 you still hacking on this or nah?

bartico6:
Oh my, I forgot to update it again. I'll fix it in an hour, sincere apologies :c

bartico6:
This time this *really* should be good. Please check :-)

hakusaro (changes requested):
This is almost ready for merge, but the submodule change will revert TSAPI back by 1 commit to the tree from DeathCradle's fork, not the tree post merge. Can you undo that change? Everything else lgtm!

bartico6:
I'm not sure what I'm supposed to do here, I didn't make any changes to TSAPI myself, only did a merge from general-devel... should I merge gendev again hoping it evens out? should I forcibly checkout the previous submodule version?

bartico6:
This explains what I did. As a sidenote it appears I.. checked out `disable-commands` in the work tree? I guess I was checking something and forgot to undo it, it has been a month. Sorry for the idiotic commits, we might want to squash it maybe?
obraz

hakusaro (approved):
lgtm_7

QuiCM (approved):

PR #2405, opened 2021-07-17
Add notify msg for online player group change

moisterrific:
Minor issue: changing a player's group permanently does not send a success msg to the target player, this is inconsistent considering `tempgroup` sends a success msg.

This makes it more consistent with how `/tempgroup` works, also added silent command support.

yes


TShockAPI/Commands.cs
```diff
@@ -1116,6 +1116,12 @@ private static void ManageUsers(CommandArgs args)
 					TShock.UserAccounts.SetUserGroup(account, args.Parameters[2]);
 					TShock.Log.ConsoleInfo(args.Player.Name + " changed account " + account.Name + " to group " + args.Parameters[2] + ".");
 					args.Player.SendSuccessMessage("Account " + account.Name + " has been changed to group " + args.Parameters[2] + "!");
+					
+					var players = TSPlayer.FindByNameOrID(account.Name);
+					if (players.Count == 0)
+						return;
+					else if (!args.Silent)
+						players[0].SendSuccessMessage($"{args.Player.Name} has changed your group to {args.Parameters[2]}");
```
QuiCM:
```suggestion
					//send message to player with matching account name
					var player = TShock.Players.FirstOrDefault(p => p != null && p.Account?.Name == account.Name)
					if (player != null && !args.Silent)
						player.SendSuccessMessage($"{args.Player.Name} has changed your group to {args.Parameters[2]}");
```

QuiCM (changes requested):

TShockAPI/Commands.cs
QuiCM: Pleaase note this suggestion was written by hand & untested - will need to be sanity checked

moisterrific: I can test it and let u know

moisterrific: tested it and it works, just missing a `;`

moisterrific (commented):
?

hakusaro (changes requested):
This is pretty good but why does the PR not have any description of what changed?

moisterrific:
> 
> 
> This is pretty good but why does the PR not have any description of what changed?

fixed

hakusaro (dismissed):

QuiCM (dismissed):

QuiCM (approved):

PR #2410, opened 2021-07-22
New features in TSPlayer.

AgaSpace:
The ability to set the pvp mode.
Ability to remove buffs.

no :(

### DIFF
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1562,6 +1562,16 @@ namespace TShockAPI
 			Main.player[Index].team = team;
 			NetMessage.SendData((int)PacketTypes.PlayerTeam, -1, -1, NetworkText.Empty, Index);
 		}
+		
+		/// <summary>
+		/// Sets the player's pvp.
+		/// </summary>
+		/// <param name="mode">The state of the pvp mode.</param>
+		public virtual void SetPvP(bool mode)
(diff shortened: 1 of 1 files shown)


moisterrific:
Does this mean buffs can also be banned as long as the server is in SSC?

AgaSpace:
> Does this mean buffs can also be banned as long as the server is in SSC?

You can remove buff from the player, so I think you are right.

bartico6:
The removal of buffs won't work, after you send the `PlayerBuff` packet, the player will lose all buffs.

AgaSpace:
> The removal of buffs won't work, after you send the `PlayerBuff` packet, the player will lose all buffs.

I apologize for my inattention.
Terraria Server does not record buffs.

Kojirremer:
> 
> 
> Does this mean buffs can also be banned as long as the server is in SSC?

The BanBuff plugin gives you the ability to ban (not remove) them in the same way as Itemban, TileBan and ProjBan. It's strict though; it will immediately kick offenders rather than disable them.

imagen

bartico6:
> 
> 
> > The removal of buffs won't work, after you send the `PlayerBuff` packet, the player will lose all buffs.
> 
> I apologize for my inattention.
> Terraria Server does not record buffs.

You can keep the PvP status part, though - that one looks good and should work :-)

PR #2415, opened 2021-07-31
Change RespawnSeconds & RespawnBossSeconds from 10 to 0

moisterrific:
This way it will use the game's timers. I tested on a master world, setting them both to 0 will just make the respawn default to 15s normally and 30s for bosses. So I would assume it will also use whatever the default time is on lower difficulties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 ## Upcoming changes
 * Fixed SendTileRectHandler not sending tile rect updates like Pylons/Mannequins to other clients. (@Stealownz)
 * Fix some typos that have been in the repository for over a lustrum. (@Killia0)
+* Changed `RespawnSeconds` and `RespawnBossSeconds` from `10` to `0` to respect the game's default respawn timers. (@moisterrific)
 
 ## TShock 4.5.5
 * Changed the world autosave message so that it no longer warns of a "potential lag spike." (@hakusaro)
--- a/TShockAPI/Configuration/TShockConfig.cs
+++ b/TShockAPI/Configuration/TShockConfig.cs
@@ -241,12 +241,12 @@ namespace TShockAPI.Configuration
 		[Description("Allows groups on the banned item allowed list to spawn banned items even if PreventBannedItemSpawn is set to true.")]
 		public bool AllowAllowedGroupsToSpawnBannedItems = false;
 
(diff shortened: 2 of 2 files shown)


iamyou (approved):

iamyou (dismissed):
I found players will respawn faster than expected in expert mode

moisterrific:
> I found players will respawn faster than expected in expert mode

is this with the existing 10s values or with my proposed 0s?

QuiCM:
So to summarize
* Less than vanilla time works
* More than vanilla time doesn't work (not sure what happens though)
* 0 makes it vanilla time?

moisterrific:
> Less than vanilla time works

Yes

> More than vanilla time doesn't work (not sure what happens though)

I think it just defaults to the max possible timer in vanilla? (15s normal, 30s bosses, master difficulty) Don't remember if TShock will force respawn you again when it hits 40s though if you set the time to that.

> 0 makes it vanilla time?

That's what I found out from setting it to 0, I was just curious what would happen and played around w/ the setting. Turns out 1s is the lowest possible time you can have, 0 somehow defaults to vanilla timers.

hakusaro (dismissed):
I raised it to 10s to stop desync. I don't see any evidence to support changing this to 0 by default. Can you provide videos that demonstrate the exact behavior when set to 0 and why this is better than not?

moisterrific:
> 
> 
> I raised it to 10s to stop desync. I don't see any evidence to support changing this to 0 by default. Can you provide videos that demonstrate the exact behavior when set to 0 and why this is better than not?

You can test it out yourself if you have any doubts, 0s somehow "just works".  Using the game default respawn countdown is better because it's adaptive to the world difficulty instead of a static 10s. The countdown varies depending on difficulty, so 10s only works if you're playing on journey/classic, meanwhile expert/master is 15 and 30 for bosses.

(also the 0s way does not have any desync issue)

hakusaro:
I'm changing my review to _abstain._

hakusaro (approved):
lgtm_7

QuiCM (approved):
This is probably worthwhile. Keeping vanilla things where possible & augmenting where desired is a good path to take

QuiCM:
@moisterrific merge conflicts :(

moisterrific:
i can't easily resolve the merge conflict on my end like i could before, im guessing it's because i deleted a prev fork of tshock i forked from chi-rei-den/axeelander since on top here it shows the patch no longer exists

image

i think what i'll do is just close this one and make a new commit with the same changes since it's such a small change

moisterrific:
See new: https://github.com/Pryaxis/TShock/pull/2450

PR #2420, opened 2021-08-08
Return loopback IP if RealPlayer is false

csmir:
Quick line replacement to return the loopback IP if the RealPlayer bool is false on requesting the player IP.


QuiCM (approved):
Approved assuming this doesn't impact any other systems (eg REST)

hakusaro (approved):

hakusaro:
lgtm_7

PR #2421, opened 2021-08-08
Fix ban add example visual bug

moisterrific:
Fixes this problem if you run `/ban help examples` from the console

image


moisterrific:
nvm

PR #2437, opened 2021-08-23
Fix return for sundial

csmir:
Fixed a bug in the handler that handles sundial usage.
This change adjusts the (unintentional) logic from:
* Sundials can never be used

To:
* Sundials can only be used when:
  - The user has permission
  - Time is not frozen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added a new `/respawn` command that lets you respawn yourself or another player. Respawning yourself requires the `tshock.respawn` permission and respawning others requires the `tshock.respawn.other` permission. The full command syntax is `/respawn [player]`. (@moisterrific)
 * Added a notification message and silent command support for permanently changing a target player's user group. Now players who received a group change will be notified of their new group if they are currently online. (@moisterrific, @QuiCM)
 * Changed the TSPlayer IP method to return the loopback IP if RealPlayer is false. (@Rozen4334)
-* Changed return result in ` HandleSpecial ` to fix problem where using sundial would be ignored and return true even if succesfully passing checks. (@Rozen4334)
+* Fixed a bug that caused sundials to be ignored all the time, instead of only when the player has no permission or time is frozen. (@Rozen4334)
 
 ## TShock 4.5.5
 * Changed the world autosave message so that it no longer warns of a "potential lag spike." (@hakusaro)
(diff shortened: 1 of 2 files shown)


QuiCM (dismissed):
I think this looks ok but GitHub mobile melted my brain a little

QuiCM:
Logic flow could be adjusted with new positioning of the returns, but purely semantic. Happy for this to be merged if no other queries/complaints

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added a new `/respawn` command that lets you respawn yourself or another player. Respawning yourself requires the `tshock.respawn` permission and respawning others requires the `tshock.respawn.other` permission. The full command syntax is `/respawn [player]`. (@moisterrific)
 * Added a notification message and silent command support for permanently changing a target player's user group. Now players who received a group change will be notified of their new group if they are currently online. (@moisterrific, @QuiCM)
 * Changed the TSPlayer IP method to return the loopback IP if RealPlayer is false. (@Rozen4334)
+* Changed return in ` HandleSpecial ` (GetDataHandlers) to fix Sundial problem. (@Rozen4334)
```
hakusaro: What is the sundial problem? Please elaborate more on the sundial problem in this changelog entry.

hakusaro (changes requested):
Only reading this PR alone, there is no description and the changelog doesn't elaborate on what the sundial problem is. I do apologize but without this PR being self-descriptive, particularly the changelog (which we have been skewered over in the past), it is difficult for me to approve this.

To remedy the situation, please update the changelog and pull-request description such that you've:

1. Described the sundial problem.
2. Described what you've changed to fix the sundial problem.

Please note that the changelog is read by both developers and normal people. Unless people are already familiar with the sundial problem, it is extremely difficult to discern how this rectifies the sundial problem.

Thank you so much for your pull request! I hope we can merge it soon.

CHANGELOG.md
csmir: Hoping I have clarified enough in latest commit.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added a new `/respawn` command that lets you respawn yourself or another player. Respawning yourself requires the `tshock.respawn` permission and respawning others requires the `tshock.respawn.other` permission. The full command syntax is `/respawn [player]`. (@moisterrific)
 * Added a notification message and silent command support for permanently changing a target player's user group. Now players who received a group change will be notified of their new group if they are currently online. (@moisterrific, @QuiCM)
 * Changed the TSPlayer IP method to return the loopback IP if RealPlayer is false. (@Rozen4334)
+* Changed return result in ` HandleSpecial ` to fix problem where using sundial would be ignored and return true even if succesfully passing checks. (@Rozen4334)
```
QuiCM: May I suggest
```suggestion
* Fixed a bug that caused sundials to be ignored all the time, instead of only when the player has no permission or time is frozen. (@Rozen4334)
```

csmir: >:D, done

QuiCM (approved):

hakusaro (approved):
lgtm_8

PR #2448, opened 2021-09-05
The ability to change the SSC for the player.

AgaSpace:


moisterrific:
> No description provided.

Can you add a description for this pls?

also you need to update the changelog with a brief description of your changes

bartico6:
I'm personally against making this change. Messing with SSC status while the player is logged on is already a "*forbidden technique*" used for admin slowhacking, and instead  should be implemented by its users *IF* they really need it.

I personally believe TShock shouldn't offer this functionality because your average plugin developer may not, at least in my opinion, be entrusted with such a tool.

PR #2453, opened 2021-09-20
Fix player buffs applied to other players through `PlayerAddBuff` packet

drunderscore:
Fixed `HandlePlayerAddBuff` data handler always being marked as `Handled`, and therefore never allowing the `PlayerAddBuff` to be sent to anyone.

Fixed `OnPlayerBuff` logic inside of Bouncer when checking `WhitelistBuffMaxTime` to determine if a buff being applied longer than allowed. This also marks the handler as `Handled` if the check fails, and resyncs the buffs with the `PlayerAddBuff` packet to the sender. (@drunderscore)

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -45,6 +45,27 @@ namespace TShockAPI
 		internal Handlers.SyncTilePickingHandler SyncTilePickingHandler { get; private set; }
 
 		/// <summary>
+		/// A class that represents the limits for a particular buff when a client applies it with PlayerAddBuff.
+		/// </summary>
+		internal class BuffLimit
+		{
+			/// <summary>
+			/// How many ticks at the maximum a player can apply this to another player for.
+			/// </summary>
+			public int MaxTicks { get; set; }
+			/// <summary>
+			/// Can this buff be added without the receiver being hostile (PvP)
+			/// </summary>
+			public bool CanBeAddedWithoutHostile { get; set; }
+			/// <summary>
+			/// Can this buff only be applied to the sender?
+			/// </summary>
+			public bool CanOnlyBeAppliedToSender { get; set; }
+		}
+
+		internal static BuffLimit[] PlayerAddBuffWhitelist;
+
+		/// <summary>
 		/// Represents a place style corrector.
 		/// </summary>
 		/// <param name="player">The player placing the tile.</param>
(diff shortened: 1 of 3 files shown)


drunderscore:
The original bug is that when shooting another player with a Cursed Arrow (to apply Cursed Inferno), the check against `WhitelistBuffMaxTime` was using `<=` instead of `>`, which prevented the buff from being applied -- but never actually did because it was marked as **not** handled. Only a second bug within the `HandlePlayerAddBuff` data handler allowed this bug to continue.

Ichor Arrows (to apply Ichor) will fail simply because of the second bug.

bartico6 (changes requested):
This is a dangerous change. 
The current buff system says "drop all buffs *UNLESS* whitelisted", whereas your system says "allow all buffs *UNLESS* blacklisted".
The better choice would be to determine the parameters of buffs you can apply in vanilla and add them to the whitelist with the correct times, since that was the goal of this rubberband check in the first place. Doing it the way you just did could reopen the doors for old exploits that this currently stops.

drunderscore:
@bartico6 Yeah I see what you mean, definitely a good idea to only allow buffs that happen in vanilla to others (which should be a short list regardless)

I'll go through and update `WhitelistBuffMaxTime` with all buffs and their max time, and remove the `> 0` check.

bartico6:
Just commit a fix with the buffs you're currently seeing problems with.

hakusaro:
@bartico6 fyi, this was updated, can you check again?

drunderscore:
Rebased, whitelisted more buffs, and added `CanOnlyBeAppliedToSender`, to indicate that the sender may only apply this buff to themself. This is used by `BuffIDs.Campfire` (aka Cozy Fire), `BuffID.WaterCandle`, and some others.

hakusaro:
@drunderscore still has merge conflicts on GitHub on web, can you fix with a rebase so that it runs CI?

drunderscore:
@hakusaro Fixed! Sorry about that, I actually didn't rebase even though I thought I did...

QuiCM (dismissed):
This looks reasonable to me. Would appreciate if @bartico6 would re-review as well

TShockAPI/Bouncer.cs
```diff
@@ -1629,59 +1629,86 @@ internal void OnPlayerBuff(object sender, GetDataHandlers.PlayerBuffEventArgs ar
...
 				return;
 			}
+
+			if (time > buffLimit.MaxTicks)
+		    {
+                TShock.Log.ConsoleDebug("Bouncer / OnPlayerBuff rejected time too long from {0}", args.Player.Name);
+                args.Player.SendData(PacketTypes.PlayerBuff, "", id);
+                args.Handled = true;
+                return;
+		    }
```
Arthri: The indentation switched to spaces here

TShockAPI/Bouncer.cs
```diff
@@ -1629,59 +1629,86 @@ internal void OnPlayerBuff(object sender, GetDataHandlers.PlayerBuffEventArgs ar
...
 			{
-				TShock.Log.ConsoleDebug("Bouncer / OnPlayerBuff rejected buff time whitelists from {0}", args.Player.Name);
-				args.Handled = false;
+				TShock.Log.ConsoleDebug("Bouncer / OnPlayerBuff rejected hostile/pvp from {0}", args.Player.Name);
+				args.Player.SendData(PacketTypes.PlayerBuff, "", id);
+				args.Handled = true;
 				return;
 			}
+
+			if (time > buffLimit.MaxTicks)
```
Arthri: I'm not sure what it could be used for, but the time could be negative and it would pass this check

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+		{
+			// How many ticks at the maximum a player can apply this to another player for.
+			public int MaxTicks { get; set; }
+			// Can this buff be added without the receiver being hostile (PvP)
+			public bool CanBeAddedWithoutHostile { get; set; }
+			// Can this buff only be applied to the sender?
+			public bool CanOnlyBeAppliedToSender { get; set; }
+		}
+
+		public static BuffLimit[] AddPlayerBuffWhitelist;
```
Arthri: I think this should be in Bouncer, it feels a bit out of place in GetDataHandlers

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
 	{
 		private static Dictionary<PacketTypes, GetDataHandlerDelegate> GetDataHandlerDelegates;
 
-		public static int[] WhitelistBuffMaxTime;
+		public class BuffLimit
+		{
+			// How many ticks at the maximum a player can apply this to another player for.
```
Arthri: I think this would be better as an XML summary/comment, the one with triple slash

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+			public int MaxTicks { get; set; }
+			// Can this buff be added without the receiver being hostile (PvP)
```
Arthri: I think this would be better as an XML summary/comment

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+			public bool CanBeAddedWithoutHostile { get; set; }
+			// Can this buff only be applied to the sender?
```
Arthri: I think this would be better as an XML summary/comment

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
-			WhitelistBuffMaxTime[20] = 600;
-			WhitelistBuffMaxTime[0x18] = 1200;
-			WhitelistBuffMaxTime[0x1f] = 120;
-			WhitelistBuffMaxTime[0x27] = 420;
+			#region Whitelist
+
+			AddPlayerBuffWhitelist = new BuffLimit[Main.maxBuffTypes];
+			AddPlayerBuffWhitelist[BuffID.Poisoned] = new BuffLimit
+			{
+				MaxTicks = 60 * 10
```
Arthri: Poisoned can be inflicted for up to 60 seconds by Poison Dart(according to the wiki)

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+			{
+				MaxTicks = 60 * 20
+			};
+			AddPlayerBuffWhitelist[BuffID.Confused] = new BuffLimit
+			{
+				MaxTicks = 60 * 4
+			};
+			AddPlayerBuffWhitelist[BuffID.CursedInferno] = new BuffLimit
+			{
+				MaxTicks = 7 * 60
```
Arthri: Minor but, the format is inconsistent here, it's 60 * x everywhere else

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+			};
+			AddPlayerBuffWhitelist[BuffID.Wet] = new BuffLimit
+			{
+				MaxTicks = 60 * 30,
+				// The Water Gun can be shot at other players and inflict Wet while not in PvP
+				CanBeAddedWithoutHostile = true
+			};
+			AddPlayerBuffWhitelist[BuffID.Ichor] = new BuffLimit
+			{
+				MaxTicks = 60 * 20
```
Arthri: I believe the max is 19, from Flask of Ichor's effect

TShockAPI/GetDataHandlers.cs
```diff
@@ -80,19 +80,140 @@ public static class GetDataHandlers
...
+			};
+			AddPlayerBuffWhitelist[BuffID.GelBalloonBuff] = new BuffLimit
+			{
+				MaxTicks = 60 * 30,
+				// The Sparkle Slime Balloon inflicts this while not in PvP
+				CanBeAddedWithoutHostile = true
+			};
+			AddPlayerBuffWhitelist[BuffID.Frostburn] = new BuffLimit
+			{
+				MaxTicks = 60 * 8
```
Arthri: The max is 14(from Frost Armor) according to the wiki

Arthri (changes requested):

TShockAPI/GetDataHandlers.cs
drunderscore: ~~This does not apply to PvP in practice~~

drunderscore: I lied -- it does apply to PvP, will change it.

TShockAPI/GetDataHandlers.cs
drunderscore: Congratulations, you've discovered a vanilla desync bug -- see https://github.com/Pryaxis/Terraria/issues/9... I will leave it for the future.

TShockAPI/GetDataHandlers.cs
drunderscore: Another vanilla desync bug, the server is never informed of this buff... I might make one issue to list all of these over on [Pryaxis/Terraria](https://github.com/Pryaxis/Terraria)... I'll talk about it in Discord

TShockAPI/GetDataHandlers.cs
Arthri: Ah I see, in that case the max would be 14 from Ichor Dart. I don't remember stuff that inflicts up to 20s ichor

Arthri:
https://github.com/Pryaxis/TShock/compare/c93cc7bb09d34c19fc908ca201061f5b9fc52f7d..b2214685ca3d49ea009112cb97a1d21e9cf9bd66

drunderscore:
Added buff limits for `MonsterBanner`, `HeartLamp`, `PeaceCandle`, `StarInBottle`, `CatBast`

Arthri:
https://github.com/Pryaxis/TShock/compare/b2214685ca3d49ea009112cb97a1d21e9cf9bd66..f53367399fd1baa7aad55ca77f827fa3b2e2f8f6

hakusaro:
Anyone know the status on this?

drunderscore:
The Venom buff in `Main.pvpBuff` change will be removed, as it has been corrected by Re-Logic in 1.4.4. All of the existing buff limits will be checked to ensure they are still correct.

There are certainly still some missing buff limits, but I think it would be beneficial to review and merge this, as we would only gain more features.

I'll take another look at this today, and hopefully persuade some others to review.

drunderscore:
I've rebased this, and the only thing I've removed is the `Main.pvpBuff[BuffID.Venom] = true` patch as described earlier.

@hakusaro or @Arthri if you are able to review, I'd appreciate it -- I believe this to be ready to merge, other missing buff limits can be added in due time.

drunderscore:
Changed buff limits for Solar Armor (`SolarShield1`, `SolarShield2`, `SolarShield3`), to allow them to be applied by non-senders. Added limits for Hellfire (`OnFire3`), Hearty Meal, and Frostbite (`Frostburn2`).

hakusaro (approved):

PR #2454, opened 2021-09-20
Fixed a bug that allowed projectiles out of bounds to be created

drunderscore:
Because the out of bounds check did not check equality (only greater than-ness), a projectile index of `1000` (the current `Main.maxProjectiles` value) would be allowed, which is out of bounds of `Main.projectiles`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Added a notification message and silent command support for permanently changing a target player's user group. Now players who received a group change will be notified of their new group if they are currently online. (@moisterrific, @QuiCM)
 * Changed the TSPlayer IP method to return the loopback IP if RealPlayer is false. (@Rozen4334)
 * Fixed a bug that caused sundials to be ignored all the time, instead of only when the player has no permission or time is frozen. (@Rozen4334)
+* Fixed a bug in Bouncer's `OnNewProjectile` handler that allowed projectile indexes equal to `Main.maxProjectiles` (which would be out of bounds of `Main.projectiles`) to be created. (@drunderscore)
 
 ## TShock 4.5.5
 * Changed the world autosave message so that it no longer warns of a "potential lag spike." (@hakusaro)
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -790,7 +790,7 @@ namespace TShockAPI
 			short type = args.Type;
 			int index = args.Index;
 
-			if (index > Main.maxProjectiles)
+			if (index >= Main.maxProjectiles)
 			{
 				TShock.Log.ConsoleDebug("Bouncer / OnNewProjectile rejected from above projectile limit from {0}", args.Player.Name);
 				args.Player.RemoveProjectile(ident, owner);


punchready:
**TL;DR**: Index 1000 means the projectile is going to be created newly by the server and is therefore valid in this place.


**Explanation**:
(I hope i didn't mess anything up)
Packet 27 is used for projectile creation and updates in one ("create if not exists").
Projectiles are identified by the client-side index and owner player index, which can be seen in action both in the vanilla server source:
```cs
int num167 = 1000;
for (int num168 = 0; num168 < 1000; num168++)
{
	if (Main.projectile[num168].owner == num163 && Main.projectile[num168].identity == num162 && Main.projectile[num168].active)
	{
		num167 = num168;
		break;
	}
}
```
and in tShock's handling (called inside the packet handling by `var index = TShock.Utils.SearchProjectile(ident, owner);`):
```cs
{
	for (int i = 0; i < Main.maxProjectiles; i++)
	{
		if (Main.projectile[i].identity == identity && Main.projectile[i].owner == owner)
			return i;
	}
	return 1000;
}
```
In both cases, if no projectile with the given identifiers exists, The number 1000 is returned to signify that the projectile needs to be created instead of updated. Ergo, getting this number in the `NewProjectile` event is valid.

**More explanation on index 1000**:
To create projectiles if nothing was found to update, the vanilla server tries finding an empty projectile slot or an old projectile that can be replaced:
```cs
if (num167 == 1000)
{
	for (int num169 = 0; num169 < 1000; num169++)
	{
		if (!Main.projectile[num169].active)
		{
			num167 = num169;
			break;
		}
	}
}
if (num167 == 1000)
{
	num167 = Projectile.FindOldestProjectile();
}
```
If both of these checks fail (`FindOldestProjectile` may also return 1000), the code will write the projectile data into `Main.projectile[1000]` which is an existing slot specifically made for this case (the same thing exists for `Main.npc`, `Main.item` and a few others):
```cs
public static Projectile[] projectile = new Projectile[1001];
```

drunderscore:
Where does the client send a packet 27 with an ID of 1000? I've never seen that.

drunderscore:
It does not help that `Identity` and `Index` are named very unhelpfully, and are not properly documented either -- often with confusing question marks (in `NewProjectileEventArgs`), or unsure ellipses (in `ProjectileKillEventArgs`)

With this patch
```diff
diff --git a/TShockAPI/Bouncer.cs b/TShockAPI/Bouncer.cs
index a625190b..f33ec46c 100644
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -790,6 +790,8 @@ namespace TShockAPI
 			short type = args.Type;
 			int index = args.Index;
 
+			TShock.Log.ConsoleDebug($"Bouncer::OnNewProjectile for Identity {args.Identity}, Index {args.Index}");
+
 			if (index > Main.maxProjectiles)
 			{
 				TShock.Log.ConsoleDebug("Bouncer / OnNewProjectile rejected from above projectile limit from {0}", args.Player.Name);
```

I have seen `Index` with 1000 on occasion:
`Debug: Bouncer::OnNewProjectile for Identity 3, Index 1000`
I'm wondering why it's feel so arbitrary when it is 1000? To my knowledge, `Index` represents the ID the client sent in the packet.

In my own server software, I've never seen a projectile with ID 1000. It certainly could be a bug in my own implementation, but I'd be surprised to learn that as I've had no issues thus far with it.

drunderscore:
Unfortunately I was far too impatient and not vigilant to see that `Identity` actually represents what the client has sent. So the client never does send packet 27 with an ID of 1000, but I do see why now this code is the way it is... which is unfortunate cause I can't seem to understand why it needs to be this way...

Why does `SearchProjectile` not simply check if the projectile `Main.projectile[Identity]` is active (and has the proper owner) to know if it is new or not? Why is it that we must iterate all of the projectiles like this? I must be either missing something obvious, or this is some weird implementation that I'm glad I've never gone down.

punchready:
In the first code snippet i sent, `num163` is the "owner" and `num162` is the "identity" of the projectile:
```cs
int num162 = reader.ReadInt16();
Vector2 position3 = reader.ReadVector2();
Vector2 velocity4 = reader.ReadVector2();
int num163 = reader.ReadByte();
```
Both of which are distinct from the "index" of the projectile inside the server-side projectile array, which is why there is a for loop looking for a matching projectile with the given owner and identity which i showed earlier.
This can also be seen in the code sending packet 27:
```cs
Projectile projectile = Main.projectile[number];
writer.Write((short)projectile.identity);
writer.WriteVector2(projectile.position);
writer.WriteVector2(projectile.velocity);
writer.Write((byte)projectile.owner);
writer.Write((short)projectile.type);
```
It is called with the first number being the index in the local projectile array, but it sends the identity and owner of the projectile, not an index.

About identity vs index:
```cs
public static int NewProjectile(IProjectileSource spawnSource, float X, float Y, float SpeedX, float SpeedY, int Type, int Damage, float KnockBack, int Owner = 255, float ai0 = 0f, float ai1 = 0f)
{
	int num = 1000;
	for (int i = 0; i < 1000; i++)
	{
		if (!Main.projectile[i].active)
		{
			num = i;
			break;
		}
	}
	if (num == 1000)
	{
		num = FindOldestProjectile();
	}
	Projectile projectile = Main.projectile[num];
	projectile.SetDefaults(Type);
	//...
	projectile.owner = Owner;
	//...
	projectile.identity = num;
	//...
}
```
On projectile creation, the projectile identity is set to the index the projectile has in the local projectile array (may it be on the client or server). However, to identify projectiles between client and server only the identity is used for search, never a direct index, this is probably done to prevent desync problems and allow clients to create projectiles in a slot that is already full on the server side. Who knows why the same system doesn't exist for items or NPCs...

And don't worry, we are all struggling to understand the redcode sometimes :)

drunderscore:
ah I understand, the `Identity` and `Index` can be distinct from each other, and the `1000` is a magic value that is used to specify a new projectile.

I see it as an unusual implementation but you're certainly right the `Index` being `1000` in ~~many~~all cases is correct.

bartico6:
-snip- actually unnecessary everything was hashed out

PR #2504, opened 2021-11-22
Remove DIY password hashing crypto

hakusaro:
The old system for hashing passwords and permitting users to select
their algorithm has been deprecated and phased out since 2015. This
removes the remaining functions for hashing passwords to clear the way
for .NET5/6 and for OTAPI 3.

In 211b70ca3771fb42c5284d5a8565e7a2bad1ac3a, I allowed blank passwords
to upgrade to bcrypt hashes. However, the minimum password length has
been 4 historically for a long time. So I don't actually assume a lot of
users have blank passwords, so I think there are very few, if any of the
old hashes laying around.

So therefore, I think this is pretty much safe to merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming changes
 * Fixed the `/respawn` command to permit respawning players from the console. (@hakusaro, @Kojirremer)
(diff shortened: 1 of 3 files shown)


SignatureBeef (approved):
Lgtm. The code when reviewing last night shown that bcrypt should be used anyway and that these were essentially a legacy fallback. These changes also removes my concern about md5cng which is not on the new .net platform due to being recommended to move to better algorithms.

PR #2513, opened 2021-11-26
Update new buffs for NPC

sgkoishi:
The new TentacleSpike buff is allowed

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2383,6 +2383,7 @@ namespace TShockAPI
 		}
 
 		// These time values are references from Projectile.cs, at npc.AddBuff() calls.
+		// Moved to Projectile.StatusNPC(int i).
 		private static Dictionary<int, short> NPCAddBuffTimeMax = new Dictionary<int, short>()
 		{
 			{ BuffID.Poisoned, 3600 },              // BuffID: 20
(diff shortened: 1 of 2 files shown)


moisterrific:
Just one question, how did you find out what the max buff time is? Other than that LGTM

sgkoishi:
> Just one question, how did you find out what the max buff time is? Other than that LGTM

As mentioned in the comment, those times were moved to `Projectile.StatusNPC`. ~~Maybe I should not post decompiled code here?~~

fix #2514

TShockAPI/Bouncer.cs
```diff
@@ -2414,9 +2415,10 @@ internal static int GetMaxPlaceStyle(int tileID)
 			{ BuffID.RainbowWhipNPCDebuff, 240  },  // BuffID: 316
 			{ BuffID.MaceWhipNPCDebuff, 240  },     // BuffID: 319
 			{ BuffID.GelBalloonBuff, 1800  },       // BuffID: 320
-			{ BuffID.OnFire3, 360 },		// BuffID: 323
-			{ BuffID.Frostburn2, 900 },		// BuffID: 324
-			{ BuffID.BoneWhipNPCDebuff, 240 }	// BuffID: 326
+			{ BuffID.OnFire3, 360 },                // BuffID: 323
```
hakusaro, line 2418: Why did this line change?

hakusaro (approved):
lgtm_7

hakusaro:
@moisterrific are you okay with this being merged?

TShockAPI/Bouncer.cs
sgkoishi: It was a mixture of space and tabs and I replace them with spaces xD

hakusaro: Those are tabs, not spaces. Also I think the convention is only tabs. You sure you intended to convert to spaces?

sgkoishi: All the previous lines of this dictionary are spaces except these three lines (`BuffID 323, 324, 326`). I can replace them with tabs, but GitHub uses 8 spaces for a tab so it might be not aligned if someone reads it from the website.
image

hakusaro: Nope, that’s okay! You’ve done well!

hakusaro (approved):

PR #2521, opened 2021-11-30
Fixed exploit with cheaters ability to create custom death message

AgaSpace:


TShockAPI/Bouncer.cs
```diff
@@ -2242,6 +2243,17 @@ internal void OnPlayerDamage(object sender, GetDataHandlers.PlayerDamageEventArg
 				return;
 			}
 
+			/*
+			 * A player must take damage from another player in any case
+			 * I didn't add sourceProjectile because the player can't always die from a projectile.
+			*/
+			if (TShock.Config.Settings.DisableCustomDeathMessages &&
+				(reason._sourcePlayerIndex == -1 || reason._sourceNPCIndex != -1 || reason._sourceOtherIndex != -1 || reason._sourceCustomReason != null))
```
hakusaro, line 2254: Can you explain why these are all checking for == -1?

TShockAPI/Configuration/TShockConfig.cs
```diff
@@ -460,6 +460,10 @@ public class TShockSettings
 		/// <summary>Prohibit the use of Zenith projectile with different objects instead of weapons.</summary>
 		[Description("Prohibit the use of Zenith projectile with different objects instead of weapons.")]
 		public bool DisableModifiedZenith = false;
+
+		/// <summary>Allows you to disable or enable protection against creating custom messages with death. Created for developers who came up with a more original solution to this problem.</summary>
```
hakusaro, line 464: Is there any way to do this ingame? If not, should we just fix it and move on and not add a config option for it?

hakusaro (changes requested):
I'm not quite sure on the first check why you're checking to see if these are -1, because it seems like we should just reject this outright all the time. Can you explain the -1 checks?

TShockAPI/Configuration/TShockConfig.cs
Arthri: No, afaict there is no way to do this in the vanilla client. However I think it would be friendlier to leave the option in, since some servers may permit/encourage the use of modified clients

TShockAPI/Bouncer.cs
AgaSpace: The constructor assigns values of -1.
When we create a "custom cause of death", we must specify the cause. Basically it is an Empty.
image

AgaSpace: When we kill another player, death must always be from either the player or the projectile
image
image

TShockAPI/Configuration/TShockConfig.cs
AgaSpace: I've encountered a server that uses a custom cause of death to make the death messages prettier.
I think similar servers will always appear, so I added this to the config.

TShockAPI/Bouncer.cs
hakusaro: @AgaSpace that makes sense. Can you add some or all of this as code comments above the changes?

AgaSpace: Done

hakusaro (approved):
lgtm! requesting another pair of eyes next.

lgtm_trick

QuiCM (approved):

PR #2528, opened 2021-12-02
Corrected customDeathReason in OnPlayerDamage

AgaSpace:
Previously it was called for absolutely any damage from the player.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2249,6 +2249,8 @@ namespace TShockAPI
 			 * 
 			 * If the player was not specified, that is, the player index is -1, then it is definitely a custom cause, as you can only deal damage with a projectile or another player.
 			 * This is how everything else works. If an NPC is specified, its value is not -1, which is a custom cause.
+			 * 
+			 * Checking whether this damage came from the player is necessary, because the damage from the player can come even when it is hit by a NPC
 			*/
 			if (TShock.Config.Settings.DisableCustomDeathMessages && id != args.Player.Index && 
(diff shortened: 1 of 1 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -2250,7 +2250,7 @@ internal void OnPlayerDamage(object sender, GetDataHandlers.PlayerDamageEventArg
 			 * If the player was not specified, that is, the player index is -1, then it is definitely a custom cause, as you can only deal damage with a projectile or another player.
 			 * This is how everything else works. If an NPC is specified, its value is not -1, which is a custom cause.
 			*/
-			if (TShock.Config.Settings.DisableCustomDeathMessages &&
+			if (TShock.Config.Settings.DisableCustomDeathMessages && id != args.Player.Index &&
```
hakusaro, line 2255: Can you add this to a comment explaining why this check is necessary?

Can you update the changelog?

hakusaro (changes requested):

TShockAPI/Bouncer.cs
AgaSpace: >Can you add this to a comment explaining why this check is necessary?

it turns out that the player sends a packet about his damage when he is hit by a npcs

AgaSpace: >Can you update the changelog?

It seems to me that this is unnecessary

hakusaro (approved):
lgtm_7

QuiCM (approved):

hakusaro:
I merged this only because the server is virtually broken in a way without the patch. To be clear, the changelog is not optional.

PR #2535, opened 2021-12-08
Added hook for ReleaseNPC packet and anticheat for both catch and release npc

tru321:
This code should stop unregistered and logged out players to catch critter and/or releasing critter by using item. Also has additional filter for ReleaseNPC to check critter being release from right item.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 ## Upcoming changes
-* Added hook `GetDataHandlers.OnReleaseNpc` to handling ReleaseNPC packet and a bouncer to stops unregistered and logged out players from releasing critters NPC. The bouncer has additional filter to stops players who tried to release different critter using crafted packet, e.g. using bunny item to release golden bunny. (@tru321)
-* Added filter in `GetDataHandlers.HandleCatchNpc` that stops unregistered and logged out players to catch critters. (@tru321)
+* Added hook `GetDataHandlers.OnReleaseNpc` to handling ReleaseNPC packet and a bouncer to stops unregistered and logged out players on SSC servers from releasing critters NPC. The bouncer has additional filter to stops players who tried to release different critter using crafted packet, e.g. using bunny item to release golden bunny. (@tru321)
+* Added filter in `GetDataHandlers.HandleCatchNpc` that stops unregistered and logged out players on SSC servers to catch critters. (@tru321)
 
 ## TShock 4.5.12
 * Fixed the ability to spawn Zenith projectile with non-original items. (@AgaSpace)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not insert tabs into this file, under any circumstances, ever.
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
+## Upcoming changes
+* Added hook `GetDataHandlers.OnReleaseNpc` to handling ReleaseNPC packet and a bouncer to stops unregistered and logged out players from releasing critters NPC. The bouncer has additional filter to stops players who tried to release different critter using crafted packet, e.g. using bunny item to release golden bunny. (@tru321)
+* Added filter in `GetDataHandlers.HandleCatchNpc` that stops unregistered and logged out players to catch critters. (@tru321)
 
 ## TShock 4.5.12
 * Fixed the ability to spawn Zenith projectile with non-original items. (@AgaSpace)
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -105,6 +105,7 @@ namespace TShockAPI
 			GetDataHandlers.NPCAddBuff += OnNPCAddBuff;
 			GetDataHandlers.NPCHome += OnUpdateNPCHome;
 			GetDataHandlers.HealOtherPlayer += OnHealOtherPlayer;
+			GetDataHandlers.ReleaseNPC += OnReleaseNPC;
 			GetDataHandlers.PlaceObject += OnPlaceObject;
 			GetDataHandlers.PlaceTileEntity += OnPlaceTileEntity;
 			GetDataHandlers.PlaceItemFrame += OnPlaceItemFrame;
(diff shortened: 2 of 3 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -1829,6 +1830,52 @@ internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEv
...
+		/// <param name="args">The packet arguments that the event has.</param>
+		internal void OnReleaseNPC(object sender, GetDataHandlers.ReleaseNpcEventArgs args)
+		{
+			int x = args.X;
+			int y = args.Y;
+			short type = args.Type;
+			byte style = args.Style;
+			
+			// if npc released outside allowed tile
+			if ( x >= Main.maxTilesX * 16 - 16 || x < 0 || y >= Main.maxTilesY * 16 - 16 || y < 0)
```
hakusaro:
```suggestion
			if (x >= Main.maxTilesX * 16 - 16 || x < 0 || y >= Main.maxTilesY * 16 - 16 || y < 0)
```

TShockAPI/Bouncer.cs
```diff
@@ -1829,6 +1830,52 @@ internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEv
...
+			if (args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected npc release from {0}", args.Player.Name);
+				args.Handled = true;
+				return;
+			}
+
+			// if released npc not from its item (from crafted packet)
+			// e.g. using bunny item to release golden bunny 
+			if(args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type && args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
```
hakusaro:
```suggestion
			if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type && args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
```

TShockAPI/Bouncer.cs
```diff
@@ -1829,6 +1830,52 @@ internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEv
...
+			{
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected npc release from {0}", args.Player.Name);
+				args.Player.Kick("Trying to release different critter exploit!", true);
```
hakusaro: Bouncer doesn't currently kick players that perform exploits and even in your own internal checks, you don't do this all the time. I'm not opposed to kicking players that do this, but it seems inconsistent with other anticheat we have. Do you really think we should kick players, but only in this circumstance?

TShockAPI/GetDataHandlers.cs
```diff
@@ -3658,10 +3709,33 @@ private static bool HandleCatchNpc(GetDataHandlerArgs args)
 				NetMessage.SendData((int)PacketTypes.NpcUpdate, -1, -1, NetworkText.Empty, npcID);
 				return true;
 			}
+			
+			if(args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug("GetDataHandlers / HandleCatchNpc rejected catch npc {0}", args.Player.Name);
+				Main.npc[npcID].active = true;
```
hakusaro: Why does this change the NPC active state if the player's changes are being rejected?

TShockAPI/GetDataHandlers.cs
```diff
@@ -3658,10 +3709,33 @@ private static bool HandleCatchNpc(GetDataHandlerArgs args)
 				NetMessage.SendData((int)PacketTypes.NpcUpdate, -1, -1, NetworkText.Empty, npcID);
 				return true;
 			}
```
hakusaro, line 3711: What even is this indentation? Why are our source files like this? :(

TShockAPI/GetDataHandlers.cs
```diff
@@ -1638,6 +1639,56 @@ private static bool OnHealOtherPlayer(TSPlayer player, MemoryStream data, byte t
...
+			public short Type { get; set; }
+
+			/// <summary>
+			/// The NPC release style
+			/// </summary>
+			public byte Style { get; set; }
+		}
+
+		/// <summary>
+		/// Called when player release a NPC,  for checking critter released from item.
```
hakusaro:
```suggestion
		/// Called when player release a NPC, for checking critter released from item.
```

hakusaro (changes requested):
I have minor suggestions, but nothing major. Everything else looks okay to me, but I haven't tested this in terms of the exploit or not.

tru321:
thank you for reviewing. :)
i'll check the issue one by one.

TShockAPI/Bouncer.cs
tru321: i wasn't sure of what to do for players in case they doing this. my option was:
1. give warn and disabling them like using banned item,
2. kicking them out of the game, or
3. ban the player.

TShockAPI/GetDataHandlers.cs
tru321: because i saw the CatchNPC code they make it inactive before sending the data. 

but after further test, it doesnt needed. :)

TShockAPI/GetDataHandlers.cs
tru321: ?

TShockAPI/Bouncer.cs
tru321: i'm starting to thinking that option 1 is suited for this case. what about you?

TShockAPI/GetDataHandlers.cs
hakusaro: @tru321 that's not your problem but it is a problem for us

TShockAPI/GetDataHandlers.cs
hakusaro: kk, I think we can remove that unless you need it

TShockAPI/Bouncer.cs
hakusaro: I'm honestly fine with all options. You could argue that we should just make the decision but idk. @Pryaxis/tshock might have a better idea if we should kick/ban players that break the game logic or just warn them?

tru321: My reference for doing this was from GetDataHandlers.HandleSyncCavernMonsterType. It kicked players who using crafted packed.

TShockAPI/GetDataHandlers.cs
tru321: at least for now, its better remove it.

TShockAPI/Bouncer.cs
hakusaro: Fair

hakusaro (dismissed):
This just needs one more approver. lgtm!
lgtm_trick

TShockAPI/Bouncer.cs
```diff
@@ -1829,6 +1830,52 @@ internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEv
...
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected npc release from {0}", args.Player.Name);
+				args.Handled = true;
+				return;
+			}
+
+			// if released npc not from its item (from crafted packet)
+			// e.g. using bunny item to release golden bunny 
+			if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type && args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
+			{
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected npc release from {0}", args.Player.Name);
```
Arthri: I think it can be clarified here the reject was from a item <-> released npc mismatch

TShockAPI/Bouncer.cs
```diff
@@ -1829,6 +1830,52 @@ internal void OnHealOtherPlayer(object sender, GetDataHandlers.HealOtherPlayerEv
...
+			{
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected out of bounds from {0}", args.Player.Name);
+				args.Handled = true;
+				return;
+			}
+
+			// if player disabled
+			if (args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC rejected npc release from {0}", args.Player.Name);
```
Arthri, line 1856: I think it can be clarified here the reject was because the player is disabled

TShockAPI/Bouncer.cs
Arthri: I think this kick message is a bit inappropriate

Arthri (changes requested):

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not insert tabs into this file, under any circumstances, ever.
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
+## Upcoming changes
+* Added hook `GetDataHandlers.OnReleaseNpc` to handling ReleaseNPC packet and a bouncer to stops unregistered and logged out players from releasing critters NPC. The bouncer has additional filter to stops players who tried to release different critter using crafted packet, e.g. using bunny item to release golden bunny. (@tru321)
+* Added filter in `GetDataHandlers.HandleCatchNpc` that stops unregistered and logged out players to catch critters. (@tru321)
```
Arthri: This is only true if the server requires login

tru321:
I removed some line on GetDataHandlers.HandleCatchNPC. It wasn't need to executed.

And changing the Kick message on Bouncer.OnReleaseNPC

Arthri:
Everything looks good :+1:. The only thing left now is the changelog entry about logged-out users. Afaict it doesn't prevent all logged-out users from placing critters. It only prevents it if the server requires login(via SSC or something else)

tru321:
how about that one? i am bad at word choices.

Arthri (approved):

hakusaro (approved):

hakusaro:
lgtm_7

PR #2536, opened 2021-12-10
Remove monomod dev build shims

hakusaro:
This removes shims in-place to fetch monomod from outside of nuget,
since monomod has pushed the applicable versions to their primary nuget
repositories, and because the dev builds add unnecessary complexity to
the CI pipeline if kept in place.

### DIFF
--- a/.github/workflows/ci-otapi3.yml
+++ b/.github/workflows/ci-otapi3.yml
@@ -15,9 +15,6 @@ jobs:
         with:
           dotnet-version: '6.0.100'
 
-      - name: MonoMod dev build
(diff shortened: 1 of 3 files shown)


QuiCM:
Is this a thing that still needs to happen?

hakusaro:
In theory but if we want development build updates maybe not? I dunno tbh

SignatureBeef:
i think these can be removed, up to you guys either way though.
imo though its not often we would need to do this - i rather use the nuget build. only needed this temporarily while there were bleeding edge builds that worked on a new platform, e.g. when i shifted net5 => net6 but net6 wasnt working on the nuget release and monomod dev forgot to push the dev build to nuget.

PR #2538, opened 2021-12-13
Turn packet list into table

Arthri:
Turn packet list into table. i.e., this:
- Connect Request
- Disconnect

into

| ID  | Name |
|----|--------|
| 1  | Connect Request |
| 2 | Disconnect |

### DIFF
--- a/docs/Developer/Protocol/Index.md
+++ b/docs/Developer/Protocol/Index.md
@@ -71,11 +71,11 @@ TODO: Document the packets used and process that goes into connecting to a serve
 | 53  | [[Npc Add Buff]]                       |
 | 54  | [[Npc Update Buff]]                    |
 | 55  | [[Player Add Buff]]                    |
-| 56  | [[Update N P C Name]]                  |
+| 56  | [[Update NPC Name]]                    |
 | 57  | [[Update Good Evil]]                   |
 | 58  | [[Play Harp]]                          |
 | 59  | [[Hit Switch]]                         |
-| 60  | [[Update N P C Home]]                  |
+| 60  | [[Update NPC Home]]                    |
 | 61  | [[Spawn Bossor Invasion]]              |
 | 62  | [[Player Dodge]]                       |
 | 63  | [[Paint Tile]]                         |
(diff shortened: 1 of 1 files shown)


hakusaro:
Did you test/write these in Obsidian and verify that changing the links from paths won't break them? I don't think they will, but just so we're on the same page, you did, right?

hakusaro:
Also why is `N P C` spaced?

Arthri:
> Also why is `N P C` spaced?

oops, my bad. I used regex to add spaces before caps

Arthri:
> Did you test/write these in Obsidian and verify that changing the links from paths won't break them? I don't think they will, but just so we're on the same page, you did, right?

yeah, they worked

hakusaro:
> oops, my bad. I used regex to add spaces before caps

bruh

docs/preset permission group mappings.md
```diff
@@ -1,150 +1,150 @@
-|Permission|[[guest]]|[[default]]|[[vip]]|[[newadmin]]|[[admin]]|[[trustedadmin]]|[[owner]]|[[superadmin]]|
```
hakusaro: This file is auto generated by TShock. Please don't change it.

hakusaro (changes requested):
There's at least one file that's autogenerated by TShock that's changed here, so please don't change that.

For other changes: I'll take a look in Obsidian tomorrow. I'm not opposed to the changes related to table spacing, but I feel like I need more justification as to why we should add a ton of whitespace around everything so that the table looks nice in fixed width fonts when it's rendered in non-fixed-width fonts. Markdown uses the 2nd line with the `|----|` stuff to determine the relative placement of the elements in the table. So adding space is unnecessary on other parts of the table, at least in my understanding of the language. This means that it's aesthetic changes that require more work to comply with in the long term while the docs are still being chiseled away at, imho. If you can elaborate more on why you decided to do this, I'd be less ambivalent I think.

Please try to avoid grab bag PRs like this that are size `XXL`. If you want to update the docs, please aim for size/M or lower unless you're performing a giant content update in one burst. You have multiple disparate changes in this including reformatting, content, headers, etc. This makes it extremely difficult to review. Yes, the bar is significantly lower for a wip branch, but it's still a diff that I have to read.

hakusaro:
Also, there's no PR title

hakusaro (approved):

PR #2552, opened 2022-01-26
Fix `TSPlayer.SendTileSquare` not being centered.

0x3fcf1bbd:
STS has been used while expecting it to be a square whose center is the passed (x,y) coordinates, while said coordinates where used as the top left corner of the square; and this should fix it.
Hopefully I have checked enough times not to have any other oversights.

### DIFF
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1311,7 +1311,7 @@ namespace TShockAPI
 		/// <returns>true if the tile square was sent successfully, else false</returns>
 		public virtual bool SendTileSquare(int x, int y, int size = 10)
 		{
-			return SendTileRect((short)x, (short)y, (byte)size, (byte)size);
+			return SendTileRect((short)(x-(size/2)), (short)(y-(size/2)), (byte)size, (byte)size);
 		}
 
 		/// <summary>
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1345,8 +1345,8 @@ namespace TShockAPI
 			{
 				x = TShock.Utils.Clamp(x, Main.maxTilesX, 0);
 				y = TShock.Utils.Clamp(y, Main.maxTilesY, 0);
-				width = TShock.Utils.Clamp(x + width, Main.maxTilesX, 1) - x;
-				length = TShock.Utils.Clamp(y + length, Main.maxTilesY, 1) - x;
+				width = TShock.Utils.Clamp(x + width, Main.maxTilesX, 0) - x;
+				length = TShock.Utils.Clamp(y + length, Main.maxTilesY, 0) - x;
 				NetMessage.SendTileSquare(Index, x, y, width, length, changeType);
 				return true;
 			}
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1343,10 +1343,6 @@ namespace TShockAPI
 		{
 			try
 			{
-				x = TShock.Utils.Clamp(x, Main.maxTilesX, 0);
-				y = TShock.Utils.Clamp(y, Main.maxTilesY, 0);
-				width = TShock.Utils.Clamp(x + width, Main.maxTilesX, 0) - x;
-				length = TShock.Utils.Clamp(y + length, Main.maxTilesY, 0) - x;
 				NetMessage.SendTileSquare(Index, x, y, width, length, changeType);
 				return true;
 			}
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -1227,7 +1227,7 @@ namespace TShockAPI
 				y = 992;
 			}
 
-			SendTileSquare((int)(x / 16), (int)(y / 16), 15);
+			SendTileSquareCentered((int)(x / 16), (int)(y / 16), 15);
 			TPlayer.Teleport(new Vector2(x, y), style);
 			NetMessage.SendData((int)PacketTypes.Teleport, -1, -1, NetworkText.Empty, 0, TPlayer.whoAmI, x, y, style);
 			return true;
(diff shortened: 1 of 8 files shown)


TShockAPI/TSPlayer.cs
```diff
@@ -1311,7 +1311,7 @@ public void RemoveProjectile(int index, int owner)
 		/// <returns>true if the tile square was sent successfully, else false</returns>
 		public virtual bool SendTileSquare(int x, int y, int size = 10)
 		{
-			return SendTileRect((short)x, (short)y, (byte)size, (byte)size);
+			return SendTileRect((short)(x-(size/2)), (short)(y-(size/2)), (byte)size, (byte)size);
```
Arthri: I think it would be better to correct the specific use case, in this case. places like `TSPlayer.Teleport` rather than the method.

Fixing it at this method would mean a breaking change for plugins that expect it to not be centered

0x3fcf1bbd: Most uses I've seen inside TShockAPI is expecting it to be centered, such as in `/grow` or `/sync`, or `TSPlayer.Teleport` as you said. 
Wouldn't it be better in the long term to have this fixed, even if it causes breaking changes to some plugins, rather than keeping a somewhat faulty method whose action can be understood in two different ways? Even then the fix for those plugins would only be to switch to STR.

bartico6: the issue is you checked usages in tshock only whereas many plugin authors may have used this before, checked how it works, and adapted to it. Such a change is potentially breaking to plugins and shouldn't be snuck into main dev like this without an API tick or whatever.

Personally I'd fix the places in tshock that use it wrong and add a second overload named SendTileSquareCentered or something that will do what your current patch suggests.

0x3fcf1bbd: I guess this could work, but then should a doccomment be added to the method that states not to use it expecting a centering logic?

QuiCM: Yep. Preferably we keep backwards compat but add a way to do it properly - then add an obsolete attribute to the method that is incorrect so we can remove it in future

0x3fcf1bbd: Unable to think of a proper name for the other method, would an overload work? It would use a `byte` for size so to make it more obvious that this method cannot sync too large of tile areas. The current STS method having an obsolete attribute telling to switch to either the centered method or STR.
Also, should the method clamps the coordinates to avoid tile array OOBs?

hakusaro: Can you overload on types with the same signature in C#? I think that's not valid in Java, but I could be wrong about which language doesn't allow that.

How about `SendTileCenteredSquare` or something? It doesn't necessarily have to be perfect. I _would_ advocate changing TShock usages to the new method though.

hakusaro:
Thank you for your work on this, by the way! I really appreciate it! ✨

TShockAPI/TSPlayer.cs
0x3fcf1bbd: Well every name I could think of all sounded a quite awkward to me, so this is why I asked. 
I'm not sure if it is fully possible to do that, as assemblies already compiled referencing the current method should have the signature stored, and as each would differ it shouldn't break.
However, the method calls would be ambiguous due to the fact that numbers can easily be converted, so there would have to be quite a few funky casts to do so it works correctly. The obsolete attribute would still give off the warning so people would definitely know which method would be called. 
Not the best, so guess `SendTileCenteredSquare` will do.

0x3fcf1bbd:
Should the methods (STR, STS, & STSC) have the coordinates clamped so as to avoid OOBs that otherwise would be annoying? I think this would be nice so people wouldn't have to check for the coordinates to be in the world themselves. Ideally I think STR could have the clamping logic.

TShockAPI/TSPlayer.cs
hakusaro: I think I would change my vote to putting Centered at the end or after Send, since it’s an adjective that modifies a TileSquare, actually. Other than that it’s eh.

If you can demonstrate overloading purely on types I’m happy to say that’s a good path, but I dunno what other people think.

hakusaro:
> Should the methods (STR, STS, & STCS) have the coordinates clamped so as to avoid OOBs that otherwise would be annoying? I think this would be nice so people wouldn't have to check for the coordinates to be in the world themselves. Ideally I think STR could have the clamping logic.

I think that’s logical to support.

TShockAPI/TSPlayer.cs
0x3fcf1bbd: Got it!
It would likely mean errors on every call unless the type of the number is explicit, so maybe not the best.

bartico6: > Well every name I could think of all sounded a quite awkward to me, so this is why I asked. I'm not sure if it is fully possible to do that, as assemblies already compiled referencing the current method should have the signature stored, and as each would differ it shouldn't break. However, the method calls would be ambiguous due to the fact that numbers can easily be converted, so there would have to be quite a few funky casts to do so it works correctly. The obsolete attribute would still give off the warning so people would definitely know which method would be called. Not the best, so guess `SendTileCenteredSquare` will do.

the problem isn't physically breaking methodcalls, the problem is that the *behaviour* of the method changed. you would now be performing a different operation under the same method, this breaks expectations.

I'd expect all overloads of the same method to act the same but take different parameters - one method taking a byte and therefore performing centered networking and not cornered networking would be a TERRIBLE developer experience to debug

bartico6: all existing calls to this method use cornered networking and I'm convinced most plugins expect that, if you're going to change it to centered, suffixing the method name + obsoleting the old one would make the most sense imo

hakusaro: > the problem isn't physically breaking methodcalls, the problem is that the behaviour of the method changed. you would now be performing a different operation under the same method, this breaks expectations.

Meh, if someone has to recompile, I would rather give an old method or something anyways? E.g., keep the old function around if you want the old behavior. Ehh. Idk. It's not technically breaking ABI or API compat, but I can see how spiritually it is.

0x3fcf1bbd:
Should be fine for review now.

Arthri (approved):

hakusaro (approved):
lgtm_trick

PR #2572, opened 2022-02-25
Update for Terraria 1.4.3.4

Patrikkk:
I've decided to use the `DebugCommand` entity source for anything server related, as it makes the most sense as source. Feel free to discuss.
Running 1.4.3.4 stable on Aurora. First waiting for OTAPI to be updated on nuget, then TSAPI has to be accepted and updated to the nuget ref. Then this can be accepted and the terrariaserver submodule updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,10 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
-
 ## Upcoming changes
 * Improved the `/grow` command to reduce code duplication, use `TileID` constants for less ambiguous types. (@drunderscore)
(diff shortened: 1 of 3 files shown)


drunderscore:
Any fears it would be removed by ReLogic in the future? I doubt it, but should we create our own source?

SignatureBeef:
> First waiting for OTAPI to be updated on nuget

its up: https://www.nuget.org/packages/OTAPI/2.0.0.48

QuiCM (dismissed):
Debug command sounds fine for now, can maybe move to our own later on it if provides us anything worthwhile

hakusaro:
@QuiCM, I've updated this branch to reference the new OTAPI version (2.0.0.48) via updating to point to https://github.com/Pryaxis/TSAPI/commit/91e94388f15bd9b40786e052ef1a06173830b09d. This just included missing version numbers because the update wasn't complete in the previous PR. I've also updated general-devel on this repo to downgrade to the github actions runner for windows server 2019, which fixes the build failure, and updated this branch with those changes. So if you want to approve, you can now.

I can handle the release for the next version + increment + etc if you're good with that after.

QuiCM (approved):

PR #2576, opened 2022-03-01
Reduce status text spam (e.g. load/save world) 

### LINKED ISSUE #2560: Saving world data: X% Spam
### Discussed in https://github.com/Pryaxis/TShock/discussions/2559

<div type='discussions-op-text'>

<sup>Originally posted by **YehnBeep** February  2, 2022</sup>
To reduce console/log spam, is there a way to suppress the "Saving world data" and related messages? Or at least reduce it to something like "starting saving" and "finished saving"?

115ish lines every few minutes is rather cumbersome...</div>

SignatureBeef:
First attempt at implementing #2560

Additional changes: Potentially improve debug builds not copying the plugin to ServerPlugins

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming changes (TShock 5.0.0)
-* Reduced load/save console spam
+* Reduced load/save console spam. (@SignatureBeef, @YehnBeep)
 
 ## Upcoming changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
+## Upcoming changes (TShock 5.0.0)
+* Reduced load/save console spam
+
 ## Upcoming changes
 
 ## TShock 4.5.15
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -46,6 +46,7 @@ using TShockAPI.Localization;
 using TShockAPI.Configuration;
 using Terraria.GameContent.Creative;
 using System.Runtime.InteropServices;
+using TShockAPI.Modules;
 
 namespace TShockAPI
 {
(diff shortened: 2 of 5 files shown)


hakusaro:
@SignatureBeef protip if you put "fixes #2560" in your commit it'll close the issue when the branch is merged

TShockAPI/TShock.cs
```diff
@@ -442,6 +444,41 @@ public override void Initialize()
 			}
 		}
 
+		/// <summary>
```
hakusaro: Is there any way we can move this method + the related variables/hooks into any other file that isn't `TShock.cs`? It's a minor nit but the file is giant and I think it would be better served as a "micro-plugin" that lives in its own file somewhere, even if we have to spend some work initializing it here.

TShockLauncher/TShockLauncher.csproj
```diff
@@ -32,8 +32,10 @@
 	<Target Name="MoveTShockDebug" AfterTargets="FinalCleanup;PostBuildEvent">
 		<ItemGroup>
 			<TShockPluginFilesDebug Include="$(TargetDir)/TShockAPI*" />
+			<TShockPluginFilesDebugNonPublish Include="$(ProjectDir)../TShockAPI/bin/$(Configuration)/$(TargetFramework)/TShockAPI*" Condition="'$(PublishSingleFile)' != 'true'" />
```
hakusaro: Can this change be broke into a separate commit? I definitely initially thought this was a misplaced addition.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * Do not forget to sign every line you change with your name. (@hakusaro)
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
+## Upcoming changes (TShock 5.0.0)
+* Reduced load/save console spam
```
```suggestion
* Reduced load/save console spam. (@SignatureBeef, @YehnBeep)
```

I don't know that this is going to be clear enough for average people to understand, but with names + formatting it's okay.

hakusaro (commented):
Thank you so much for your work on this! I have minor nits/suggestions, but nothing major as a deal breaker. We can ultimately accept these changes easily, but I'd prefer a couple changes. If you don't want to make changes, lmk and I'll swap to an lgtm and merge + change stuff.

TShockLauncher/TShockLauncher.csproj
SignatureBeef: I've dropped this until i can test it further, and it's not specific for this PR.

TShockAPI/TShock.cs
SignatureBeef: i've implemented a module system via d2a060b, and then the new functionality in a single file via fb67fcc
Hopefully this is what you are meaning by micro plugin - these modules can be a single file and will be discovered automatically. Im thinking over time TShock.cs can be split up to use this same system.

hakusaro (approved):
lgtm_trick


Sorry, I was in San Francisco, but I'm back at my satellite office and I'm happy to say this is AWESOME and it sets a good precedent for the future. Cheers! I love you!

QuiCM (approved):

Issue #2530, opened 2021-12-03
SSC rollback due to connection timeout, resulting in dupe exploit


pcider:
* TShock version: v4.5.11.0
* TShock build number (if known): (not sure)

#### Reproduction steps

1. Login to SSC enabled server
2. Modify inventory (by putting a valuable item in a chest)
3. Disconnect internet on client & wait for timeout
4. Reconnect internet and join server.
5. Say hi to your evil twin 
   ![image](https://user-images.githubusercontent.com/28196972/144642667-17a4bee0-f37e-43fb-b227-f3c7b87a74a8.png)
6. Inventory is rolled back to when you first joined = dupe exploit / lost items

sgkoishi:
I thought TShock will prevent players from joining when existing players have identical names. Is that changed?

pcider:
Did you guys managed to replicate it? I also noticed that setting `-connperip 1` still allows me to join with my duplicate character still in the world. I also noticed that for some reason I couldn't kick the twin sometimes.

hakusaro:
> I thought TShock will prevent players from joining when existing players have identical names. Is that changed?

This isn't the case now, at least in my testing. I'm not sure if this is a regression or if this has always been the case, however. I'm not quite sure what to say about this. I feel like it would be appropriate to kick players if they have duplicate names from other players. But obviously, it would make sense if two players both came up with the same name and joined at the same time. Should we just reject these players outright?

I kind of don't want to add a config option to turn off a duplicate detector though. It seems like it's silly to permit duplicate names. But at the same time, I could genuinely see it being useful to allow multiple of the same name, if people indeed created the same name uniquely. Then, we could just add a shim that prevents players from logging into the same account twice. Is that bad?

hakusaro:
Actually, after a period of time, one of my two test clients disconnects with no error or any reason except "lost connection." So...maybe this is actually intended? I'm actually quite confused.

tru321:
We actually experienced this too. Especially when having poor connection to server. Then trying to keep join as fast as possible even after rejected because the character still active, we managed get into server with twins still inside.

And at one time we found not only two but three same character ingame, although only one can be controlled.

bartico6:
> > I thought TShock will prevent players from joining when existing players have identical names. Is that changed?
> 
[...]

PR #2599, opened 2022-03-17
Fix `NetHooks_NameCollision` to kick duplicate players.

pcider:
This PR fixes connections being stuck in a limbo where `PendingTermination` is `true` but `PendingTerminationApproved` is `false`. This problem occurs after a name collision.

The server doesn't fully try to reset the existing player's socket, creating the situation where the existing player is still present in the server but cannot interact with it due to `Player.PendingTermination` being `true`. SSC characters are not properly handled in this situation due to SSC not being saved unless the existing player disconnects first. I handled this by just kicking the player ensuring that `saveSSI` is set to `true`.

Fixes issue #2530.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change.
 
 ## Upcoming changes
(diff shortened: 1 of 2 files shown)


drunderscore (dismissed):
Yeah this looks like the more sane, proper way to do it. Thanks so much for the fix!

drunderscore (approved):

hakusaro (approved):

hakusaro:
@gohjoseph yo! thank you so much for this! cheers! :)
lgtm_trick

QuiCM (approved):

PR #2605, opened 2022-03-21
Only allow using Teleportation Potion and friends whilst holding them

drunderscore:
Some simple protections against using these when you shouldn't be able to!

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -3814,6 +3814,13 @@ namespace TShockAPI
 			switch (type)
 			{
 				case 0: // Teleportation Potion
+					if (args.Player.ItemInHand.type != ItemID.TeleportationPotion &&
+					    args.Player.SelectedItem.type != ItemID.TeleportationPotion)
+					{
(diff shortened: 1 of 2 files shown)


TShockAPI/GetDataHandlers.cs
```diff
@@ -3814,20 +3814,41 @@ void Fail(string tpItem)
 			switch (type)
 			{
 				case 0: // Teleportation Potion
+					if (args.Player.ItemInHand.type != ItemID.TeleportationPotion &&
+					    args.Player.SelectedItem.type != ItemID.TeleportationPotion)
```
QuiCM, line 3818: If I consume my last potion to teleport, leaving the slot empty, does this check pass or fail?

drunderscore: This check passes, surprisingly. Good edge-case thinking though!

QuiCM: That's neat. Genuinely surprised lol

QuiCM (dismissed):

TShockAPI/GetDataHandlers.cs
drunderscore: Yeah sorta lucked out there... wouldn't bet on the packet order for other consumables being action then item removal though!

hakusaro (approved):

PR #2606, opened 2022-03-27
Arm64 implementation for Raspberry PI

SignatureBeef:
This PR bumps all dependencies to their latest, and in particular MonoMod for its new Arm64 support for raspberry pi thanks to @kevzhao2. 

I have also changed the SQLite library to use Microsoft.Data.Sqlite as it works out of the box for Arm64, so there is a slight change for the QueryResult to dispose of the command instead of disposing of it preemptively.

_Note: apple silicon is not included in this, [yet](https://github.com/MonoMod/MonoMod/issues/90).
Note 2: i didn't bother adding arm64 support to the changelog, since it already has the expectation of working for the current gen-dev, but i'm happy to do so as kev's work does need some praise._

Screen Shot 2022-03-27 at 9 37 46 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming changes (TShock 5.0.0)
 * Reduced load/save console spam. (@SignatureBeef, @YehnBeep)
+* Replaced SQLite library with Microsoft.Data.Sqlite for arm64 support (@SignatureBeef)
 
 ## Upcoming changes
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 
 ## Upcoming changes (TShock 5.0.0)
 * Reduced load/save console spam. (@SignatureBeef, @YehnBeep)
+* Replaced SQLite library with Microsoft.Data.Sqlite for arm64 support (@SignatureBeef)
```
```suggestion
* Replaced SQLite library with Microsoft.Data.Sqlite for arm64 support. (@SignatureBeef)
```

nit: dot

hakusaro (approved):
lgtm_trick

lgtm. minor changelog nit. I think you could include arm64 note in the changelog just as a note that says hey, we got it working. that's about it.

hakusaro (approved):
lgtm_8

PR #2610, opened 2022-04-02
More correctly implement GPLv3 compliance

hakusaro:
The previous startup disclaimer probably didn't do a good-enough job of
explaining license terms to people. This updates it emphasizing that
there is no warranty with the software, and that the actual license is
GPL v3 or later, not GPL v3. Further, since we don't ship the license
with the software (yet), this gives people a link to the license. The
license itself says we should provide a link if we don't, so yeah.

Inspired by the fact that we had another user pop into discord asking
why all of their stuff was deleted.

I tried to keep the startup disclaimer roughly-the-same-size as the
original disclaimer (no more than 3 lines, some-ish length), because
this is so annoying but I feel like we actually need to try to be
compliant here.

I'm 60% sure we should implement a command that explains this but I
really don't want to do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 ## Upcoming changes
 * Fixed `TSPlayer.GiveItem` not working if the player is in lava. (@gohjoseph)
 * Only allow using Teleportation Potions, Magic Conch, and Demon Conch whilst holding them. (@drunderscore)
+* Updated GNU GPL disclaimer when the server starts to say that it comes with `ABSOLUTELY NO WARRANTY` and added a link to the GPL as per what the license requires. (@hakusaro)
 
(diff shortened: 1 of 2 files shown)


hakusaro:
The exact thing I'm looking at is (from https://www.gnu.org/licenses/gpl-3.0.en.html):

> If the program does terminal interaction, make it output a short notice like this when it starts in an interactive mode:

    <program>  Copyright (C) <year>  <name of author>
    This program comes with ABSOLUTELY NO WARRANTY; for details type `show w'.
    This is free software, and you are welcome to redistribute it
    under certain conditions; type `show c' for details.

> The hypothetical commands `show w' and `show c' should show the appropriate parts of the General Public License. Of course, your program's commands might be different; for a GUI interface, you would use an “about box”.

Ijwu (approved):

QuiCM (approved):

hakusaro:
I no longer care about this branch

PR #2621, opened 2022-04-28
Fixed Bouncer possibly rejecting Explosive Bunny critter releases

### LINKED ISSUE #2616: Players getting kicked when using bunny cannon with quick actions
* TShock version: 4.5.17.0

#### Reproduction steps:

1. Play with a character without any elevated permissions, but that can use the bunny cannon and explosive bunnies
2. Try to shoot the bunny cannon using the quick action button without having exploding bunnies in the hotbar
3. The cannon will shoot as normal, the proyectile will fly as usual, but...
4. If the proyectile didn't hit anything while in the air, as soon as the explosive bunny lands and turns into a creature, the player gets kicked for critter not matching item being used.

#### Any stack traces or error messages?

```
2022-04-15 01:26:59 - TSPlayer: INFO: Kicked Shredder for : 'Released critter was not from its item.'
2022-04-15 01:26:59 - Utils: INFO: Broadcast: Shredder was force kicked for 'released critter was not from its item.'
2022-04-15 01:26:59 - Utils: INFO: Broadcast: Shredder has left.
2022-04-15 01:26:59 - TextLog: INFO: Shredder disconnected.
```
My guess is that since the latest Terraria updates changed the behavior of explosive bunnies to now being critters instead of proyectiles once they land, and usually critters can't be released via the quick action extra slot that appears on the hotbar while using quick actions, they are being caught as a cheat.
When I tried using the explosive bunnies directly on the bunny cannon, releasing them, or using quick actions with them in the hotbar to shoot I did not get kicked. It seems it only happens when the bunnies are not in the hotbar.

drunderscore:
Fixes #2616

When using a Bunny Cannon, an Explosive Bunny item (which is also a
critter release item) is used to create an Explosive Bunny projectile,
which will later (in the future) release an Explosive Bunny NPC, by the
release critter packet. The existing checks required that the player be
actively selecting the item to create the critter, however this didn't
make sense for Explosive Bunnies, as they would be released in the
future, possibly when the player was no longer selecting that item.

This commit relaxes the restrictions on Explosive Bunny critter
releases, now requiring either holding the release item, or having
recently created an Explosive Bunny projectile, and that the release
coordinates are within the area of one of their Explosive Bunny
projectiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Fixed `TSPlayer.GiveItem` not working if the player is in lava. (@gohjoseph)
 * Only allow using Teleportation Potions, Magic Conch, and Demon Conch whilst holding them. (@drunderscore)
 * Updated server startup language to be more clear when encountering a fatal startup error. Now, the server gives more context as to what happened so that there's a better chance of people being able to help themselves. (@hakusaro)
+* Fixed Bouncer rejecting Explosive Bunny critter release when using the Bunny Cannon, if the player had since stopped selecting the Explosive Bunny. (@drunderscore)
 
 ## TShock 4.5.17
 * Fixed duplicate characters (twins) after repeatedly logging in as the same character due to connection not being immediately closed during `NetHooks_NameCollision`. (@gohjoseph)
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1858,14 +1858,48 @@ namespace TShockAPI
 				return;
 			}
 
-			// if released npc not from its item (from crafted packet)
-			// e.g. using bunny item to release golden bunny 
-			if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type && args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
+			void rejectForCritterNotReleasedFromItem()
 			{
 				TShock.Log.ConsoleDebug("Bouncer / OnReleaseNPC released different critter from {0}", args.Player.Name);
 				args.Player.Kick("Released critter was not from its item.", true);
 				args.Handled = true;
-				return;
+			}
+
+			// if released npc not from its item (from crafted packet)
+			// e.g. using bunny item to release golden bunny
+			if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type || args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
+			{
+				// If the critter is an Explosive Bunny, check if we've recently created an Explosive Bunny projectile.
+				// If we have, check if the critter we are trying to create is within range of the projectile
+				// If we have at least one of those, then this wasn't a crafted packet, but simply a delayed critter release from an
+				// Explosive Bunny projectile.
+				if (type == NPCID.ExplosiveBunny)
+				{
+					bool areAnyBunnyProjectilesInRange;
+
+					lock (args.Player.RecentlyCreatedProjectiles)
+					{
+						areAnyBunnyProjectilesInRange = args.Player.RecentlyCreatedProjectiles.Any(projectile =>
+						{
+							if (projectile.Type != ProjectileID.ExplosiveBunny)
+								return false;
+
+							var projectileInstance = Main.projectile[projectile.Index];
+							return projectileInstance.active && projectileInstance.WithinRange(new Vector2(args.X, args.Y), 32.0f);
+						});
+					}
+
+					if (!areAnyBunnyProjectilesInRange)
+					{
+						rejectForCritterNotReleasedFromItem();
+						return;
+					}
+				}
+				else
+				{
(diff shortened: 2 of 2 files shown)


tru321:
i guess the old code should be on after checking the type, or it will never executed (not tested yet)

if (type == NPCID.ExplosiveBunny){
//new code
}else{
//old code
}

drunderscore:
@tru321 Not exactly... your issue is valid but solution isn't correct, as we still want to check if you're holding the Explosive Bunny item, as it can also be used to release the critter. I've updated the commit to ensure the correct behavior is maintained (reject if the item isn't being held for everything other than an Explosive Bunny)

tru321:
i've test the code and notice some:

**_change the "&&" to "||" because crafting packet can be one of either "type" or "style". the old line detect only if both crafted._**
if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type || args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)

**_add style detection here because the style from explosive bunny projectile is 1 or 2. to make sure its from item or projectile._**
if (type == NPCID.ExplosiveBunny && (style == 1 || style == 2))
{						
_explosives bunny projectile detector code:
i'm stuck here because after debug the **countOfExplosiveBunnyProjectilesInRange,**
sometimes its detect the projectile and sometimes not_
}

punchready:
Basically with the current code (and with the old code, actually) one can simply hold an item with `placeStyle = 0`, send an npc release with any npc id and style set to `0`, and `if ([...].makeNPC != type && [...].placeStyle != style)` will fail because the style matches, allowing the packet through entirely.
It makes more sense to first check if the npc is an explosive bunny with valid style of `1` or `2` and a valid projectile is active and early exit in that case:
```cs
if(isExplosiveBunny && styleValid && projectileExists) {
    return;
}
if(isInvalid) {
    //reject
}
```

For performance it also does not make sense to first count all valid projectiles and then compare the count to 0, better to implement a manual loop to allow exiting early to go from O(n) to O(log n) at least.

drunderscore:
Fixed existing code to check `makeNPC` and `placeStyle` separately, and now try to find a projectile within range using `Any`, instead of doing whatever dumb thing I was doing before by `Where`ing and `Count`ing.

punchready:
It is probably still necessary to limit the style to `1 or 2` for bunny projectiles.
And perhaps there is a return missing after the last `rejectForCritterNotReleasedFromItem`? I'd still move the projectile check as an optional exemption above the other logic, but the current code gets the job done too 👍

Arthri:
> i've test the code and notice some:
> 
> **_change the "&&" to "||" because crafting packet can be one of either "type" or "style". the old line detect only if both crafted._** if (args.Player.TPlayer.lastVisualizedSelectedItem.makeNPC != type || args.Player.TPlayer.lastVisualizedSelectedItem.placeStyle != style)
> 
> **_add style detection here because the style from explosive bunny projectile is 1 or 2. to make sure its from item or projectile._** if (type == NPCID.ExplosiveBunny && (style == 1 || style == 2)) { _explosives bunny projectile detector code: i'm stuck here because after debug the **countOfExplosiveBunnyProjectilesInRange,** sometimes its detect the projectile and sometimes not_ }

I think you're able to submit a review, try that next time since code suggestions are easier to visualize

Arthri (approved):

Arthri:
This works, but if the bunny is shot far enough, you still get kicked

drunderscore:
@Arthri Yeah, I suppose I could iterate _all_ this player's projectiles looking for the a Bunny, instead of just the recent ones... I just threw this together so might redo it anyway.

Arthri:
I'm not sure how recent projectiles work but, I fired a bunny first thing and it still kicked me

punchready:
`RecentlyCreatedProjectiles` only keeps projectiles for 5 seconds, so things go 💥 if your bunny was in the air for more than that

tru321:
> @Arthri Yeah, I suppose I could iterate _all_ this player's projectiles looking for the a Bunny, instead of just the recent ones... I just threw this together so might redo it anyway.

i tried to inspect projectile directly on Main.projectile, its work like a charm.
lock (Main.projectile)
{
	areAnyBunnyProjectilesInRange = Main.projectile.Any(projectile =>
	{
		return projectile.type == ProjectileID.ExplosiveBunny && projectile.owner == args.Player.Index && projectile.active && projectile.WithinRange(new Vector2(x, y), 32.0f);
	});
}

i'm testing it with packet crafting. its work too as it should be.

punchready:
This still allows mass spawning of bunnies with only one projectile active

tru321:
I think mass spawning goes on different scope. Its hard to tell from what **bunny projectile** that **bunny** spawn.

Marking a bunny projectile on projectile update like give them tag like **AlreadySpawnedABunny=false** then set them **true** in this packet it might do, but i guess that goes too far :)

hakusaro:
@Kojirremer did you get a chance to test this fix or do you need a rebuild?

Kojirremer:
> @Kojirremer did you get a chance to test this fix or do you need a rebuild?

I didn't since at the time it still looked like a WIP, and then I eventually forgot about it.

PR #2626, opened 2022-05-08
More robust parsing of large timespans (Fixes #2625)

### LINKED ISSUE #2625: Setting a /tempgroup a time too far into the future results in command failure
* TShock version: 4.5.17
* TShock build number (if known): Unknown but it's the release here: https://github.com/Pryaxis/TShock/releases/tag/v4.5.17

1. Type `/tempgroup <username> <group> 24d` into the chat/console.
2. Type `/tempgroup <username> <group> 25d` into the chat/console.
3. An error should be throw when you execute the second command.

```
2022-05-07 18:24:12 - Command: ERROR: System.ArgumentException: Invalid value '-2134967296' for parameter 'interval'.
   at System.Timers.Timer..ctor(Double interval)
   at TShockAPI.Commands.TempGroup(CommandArgs args)
   at TShockAPI.Command.Run(String msg, Boolean silent, TSPlayer ply, List`1 parms)
```

#### Any screenshots?

image

#### Any log messages from files that end in `.log` or `.txt`? What are the last 100 log messages from the server console?

[2022-05-07_18-23-20.log](https://github.com/Pryaxis/TShock/files/8646049/2022-05-07_18-23-20.log)

#### What plugins and what versions of those plugins are you running?
No plugins other than the TShockAPI.dll itself.

Please let me know if you require any further information. I'm very sorry if I missed anything; this is my first time doing this.

punchready:
Added a robust parsing method for positive timespans using an `ulong`.
Commands that parse timespans now support up to `9999999` days.
The old method must stay unmodified for compatibility, other plugins can upgrade manually if needed.

Fixes #2625

### DIFF
--- a/TShockAPI/Utils.cs
+++ b/TShockAPI/Utils.cs
@@ -608,6 +608,82 @@ namespace TShockAPI
 		}
 
 		/// <summary>
+		/// Attempts to parse a string as a positive timespan (_d_m_h_s).
+		/// </summary>
+		/// <param name="str">The time string.</param>
+		/// <param name="seconds">The seconds.</param>
+		/// <returns>Whether the string was parsed successfully.</returns>
+		public bool TryParseTime(string str, out ulong seconds)
+		{
+			seconds = 0;
+
+			if (string.IsNullOrWhiteSpace(str))
+			{
+				return false;
+			}
+
+			var sb = new StringBuilder(3);
+			for (int i = 0; i < str.Length; i++)
+			{
+				if (char.IsDigit(str[i]) || (str[i] == '-' || str[i] == '+' || str[i] == ' '))
+					sb.Append(str[i]);
+				else
+				{
+					int num;
+					if (!int.TryParse(sb.ToString().Trim(' '), out num))
+						return false;
+
+					sb.Clear();
+
+					if (num == 0)
+					{
+						continue;
+					}
+
+					int numSeconds;
+					switch (str[i])
+					{
+						case 's':
+							numSeconds = num;
+							break;
+						case 'm':
+							numSeconds = num * 60;
+							break;
+						case 'h':
+							numSeconds = num * 60 * 60;
+							break;
+						case 'd':
+							numSeconds = num * 60 * 60 * 24;
+							break;
+						default:
+							return false;
+					}
+
+					if (numSeconds > 0)
+					{
+						if (ulong.MaxValue - seconds < (uint)numSeconds)
+						{
+							return false;
+						}
+
+						seconds += (uint)numSeconds;
+					}
+					else if (seconds >= (uint)Math.Abs(numSeconds))
+					{
+						seconds -= (uint)Math.Abs(numSeconds);
+					}
+					else
+					{
+						return false;
+					}
+				}
+			}
+			if (sb.Length != 0)
+				return false;
+			return true;
+		}
+
+		/// <summary>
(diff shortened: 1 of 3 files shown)


JordanDi123:
Hey! Thanks so much for actioning this so quickly. 👍

QuiCM (dismissed):
Could you add a changelog entry?
Otherwise LGTM. Could probably update the existing `TryParseTime(string, int)` to call `TryParseTime(string, long)` and reduce code duplication, but I'm not overly fussed :)

TShockAPI/Utils.cs
```diff
@@ -607,6 +607,82 @@ public bool TryParseTime(string str, out int seconds)
...
+
+			if (string.IsNullOrWhiteSpace(str))
+			{
+				return false;
+			}
+
+			var sb = new StringBuilder(3);
+			for (int i = 0; i < str.Length; i++)
+			{
+				if (char.IsDigit(str[i]) || (str[i] == '-' || str[i] == '+' || str[i] == ' '))
```
Arthri:
```suggestion
				if (char.IsDigit(str[i]) || str[i] == '-' || str[i] == '+' || str[i] == ' ')
```

TShockAPI/Utils.cs
```diff
@@ -607,6 +607,82 @@ public bool TryParseTime(string str, out int seconds)
...
+					sb.Append(str[i]);
+				else
+				{
+					int num;
+					if (!int.TryParse(sb.ToString().Trim(' '), out num))
```
Arthri, line 633: should this be `ulong` too?

punchready: this part needs to parse signed to allow inputs like `10d-5h` which is still a positive time span. could go for long instead of int, but at this point it doesn't matter because both are incredibly large in terms of days.

punchready:
> Could probably update the existing `TryParseTime(string, int)` to call `TryParseTime(string, long)` and reduce code duplication, but I'm not overly fussed :)

the other method should stay to allow negative time spans imo, for whomever may need it.

Arthri:
Also, I just noticed, I think a second method is unnecessary. You could fit 24855 days 3 hours 14 minutes and 7 seconds worth of seconds into the positive bounds of a 32-bit signed integer, casting the seconds to ulong before multiplying to turn it into milliseconds should be enough

punchready:
The method still also ensures that the time span is positive, which is probably wanted in most cases.
The issue in `/tempgroup` is already fixed by simply multiplying by `1000d` which causes the seconds to be casted to double anyways (if my understanding is correct), considering that the timer class wants a double in any case.

Arthri:
Looks good to me :+1:

punchready:
Although now that i think of it, i forgot about some edge cases, like `1s-10d+11d`...

Arthri (approved):

hakusaro:
@punchready

> Although now that i think of it, i forgot about some edge cases, like 1s-10d+11d...

Were you planning on fixing those edge cases?

punchready:
I don't think it's worth it since nobody is ever gonna enter time in that way (first creating a negative duration and then offsetting it to be positive again), and even if somebody does it, it will just be seen as invalid instead of breaking in weird ways.

bartico6:
if you want to have it "1h shorter than 2 days" you'll input "1d23h" and not "2d-1h" lmao

I concur that it's pointless

PR #2633, opened 2022-06-29
World Evil Startup Parameter

### LINKED ISSUE #2631: World Evil Startup Parameter
### Is your feature request related to a problem? Please describe.

It looks like by default in Terraria, you can only choose between the `Corrupt`, `Crimson` and `Random` world evil types once you've reached a certain amount of progress (though I could not find any official source on this so I'm not entirely sure)

When hosting Terraria servers via TShock, since PR #1396 added support to force this option so that you can pick a type no matter how far you've progressed: [https://github.com/Pryaxis/TShock/pull/1396](https://github.com/Pryaxis/TShock/pull/1396) — which is fantastic. 

img

Unfortunately, this option is only available when generating a new world with console commands. 

Terraria already has a few startup parameters / command-line arguments to automatically create worlds on startup to make things a little bit easier: 

- `-seed <seed>` — Set the world seed
- `-autocreate <number>` — Automatically create a world with a specific size 

(_Source: https://terraria.fandom.com/wiki/Server_)

As far as I could tell, there is no way currently to do the same with the world evil settings. Having something like `-worldevil <crimson/corrupt/random>` would be a great addition in my opinion, although I'm not entirely sure whether it's possible.

---

### Describe the solution you'd like
Add a startup parameter to the server which would automatically modify the `WorldGen.WorldGenParam_Evil` setting to a specific number depending on what's passed in. 

For example, adding `-worldevil corrupt`  would modify that to be `0` so that the corrupted world evil is used. 

I am unsure whether this is the right solution, or whether there is one at all — but poking around in the source code, this is how the game does it in `Terraria.Main.DedServ`:

```c#
if (SettingsUnlock_WorldEvil) {
    flag2 = true;
    while (flag2) {
        Console.WriteLine(Language.GetTextValue("CLI.Server", versionNumber2));
        Console.WriteLine("");
        Console.WriteLine("1\t" + Language.GetTextValue("CLI.Random"));
        Console.WriteLine("2\t" + Language.GetTextValue("CLI.Corrupt"));
        Console.WriteLine("3\t" + Language.GetTextValue("CLI.Crimson"));
        Console.WriteLine("");
        Console.Write(Language.GetTextValue("CLI.ChooseEvil"));
        string value = Console.ReadLine();
        try {
            switch (Convert.ToInt32(value)) {
            case 1:
                WorldGen.WorldGenParam_Evil = -1;
                flag2 = false;
                break;
            case 2:
                WorldGen.WorldGenParam_Evil = 0;
                flag2 = false;
                break;
            case 3:
                WorldGen.WorldGenParam_Evil = 1;
                flag2 = false;
                break;
            }
        } catch {}
        try {
            Console.Clear();
        } catch {}
    }
}
```

---

### Describe alternatives you've considered
I'm afraid there really aren't any alternative options, aside from using the standard input to pipe console commands directly. This is not an ideal solution, in my opinion. Of course, if this feature is not currently possible, that is still a route I can consider. 

---

### Additional context

_None_

NotGeri:
# Story

Since [#1396](https://github.com/Pryaxis/TShock/pull/1396), people could pick their world evil type when a new world is created without having the necessary in game progression. 

We automatically generate worlds using `-autogenerate` as we use an interface for the size, seed and other properties of worlds. This meant we'd have either needed to use some hacky standard input way of piping in the desired world evil type or create a feature request here.

The prior is less than ideal for several reasons so I decided to give this a shot. 

With this change, adding a simple startup parameter, `-worldevil <crimson|corruption|random>` should achieve what we wanted and hopefully, other people having this issue can use this as well!

Closes [#2631](https://github.com/Pryaxis/TShock/issues/2631).

---

# Testing
I wasn't too confident using Terraria's built in features like this so I did a few tests to see if my changes worked and it seems like they do.

Once I managed to finally build the project locally, I simply did:

```bash
./TerrariaServer.exe -autocreate 1 -worldname <name> -worldpath ./worlds/ -world ./worlds/ -worldevil <setting>
```

Replacing `<name>` with something like `corruption_test_2` incrementing it each time and replacing `<setting>` with `crimson`, `corrupt` or `random` depending on what I wanted to achieve. 

**Results**: Everything seemed to generate perfectly. In-game venturing out for a few moments always showed the correct biome and copying over the world file to single-player showed the correct icons as well: https://i.geri.dev/vuoSLNPUO42X.gif

---

Please do let me know if I should change anything!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Fixed `TSPlayer.GiveItem` not working if the player is in lava. (@gohjoseph)
 * Only allow using Teleportation Potions, Magic Conch, and Demon Conch whilst holding them. (@drunderscore)
 * Updated server startup language to be more clear when encountering a fatal startup error. Now, the server gives more context as to what happened so that there's a better chance of people being able to help themselves. (@hakusaro)
+* Added `-worldevil <type>` command line argument (@NotGeri)
 
 ## TShock 4.5.17
 * Fixed duplicate characters (twins) after repeatedly logging in as the same character due to connection not being immediately closed during `NetHooks_NameCollision`. (@gohjoseph)
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -768,6 +768,28 @@ namespace TShockAPI
 						}
 					})
 
+				.AddFlag("-worldevil", (value) =>
+				{
+
+					int worldEvil;
+					switch (value.ToLower())
+					{
+						case "random":
+							worldEvil = -1;
+							break;
+						case "corrupt":
+							worldEvil = 0;
+							break;
+						case "crimson":
+							worldEvil = 1;
+							break;
+						default:
+							throw new InvalidOperationException("Invalid value given for command line argument \"-worldevil\".");
+					}
+
+					ServerApi.LogWriter.PluginWriteLine(this, String.Format("New worlds will be generated with the {0} world evil type!", value), TraceLevel.Verbose);
+					WorldGen.WorldGenParam_Evil = worldEvil;
+				})
 
 				//Flags without arguments
 				.AddFlag("-logclear", () => LogClear = true)


QuiCM:
I love the thorough write up! 

Code looks reasonable to me, I'm happy with this unless anyone has any strong feelings otherwise :)

QuiCM (approved):

bartico6 (approved):
A quick look reveals this is indeed good to merge, setting the correct fields etc.

PR #2655, opened 2022-08-14
Prevent server broadcast to accept empty message

PackmanDude:
If used without a parameter, the /broadcast, /bc or /say command will display `(Server Broadcast) ` only. This PR fixes that and prevents Super Admins from spamming into chat with an empty sentences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * Updated server startup language to be more clear when encountering a fatal startup error. Now, the server gives more context as to what happened so that there's a better chance of people being able to help themselves. (@hakusaro)
 * Added `-worldevil <type>` command line argument (@NotGeri)
 * Added PlayerHasBuildPermission hook to PlayerHooks. (@AnzhelikaO, @Killia0)
+* Prevent Server Broadcast to display prefix without a message. (@PackmanDude)
 
 ## TShock 4.5.17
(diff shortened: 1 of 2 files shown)


drunderscore:
This doesn't make sense... `string.Join` cannot return null, so this would never work -- you would rather want to compare it against `string.Empty`.

However, what you _should_ be doing is checking the argument count, instead of the result of the join.

PackmanDude:
Oh, so sorry. I'm not well-experienced in C# and doesn't knew how this method works and what returns. Thanks, for pointing out about arguments, I will try it now.

PackmanDude:
Huh, I need to use something called `method group`, hmm. Looks over complicated, I will leave it for someone smarter than that.

punchready:
image
What you are seeing is probably this error. This is the compiler telling you that you are trying to compare a (in this case nonexistent) method with a number, which doesn't make any sense.
The actual issue is simply the fact that the `CommandArgs` class has no `Count` field or property. You probably meant `args.Parameters.Count` instead of `args.Count`.
After fixing that this PR is perfectly fine, don't be discouraged :)

PackmanDude:
Ah yes. I looked at Microsoft Docs about Lists and there was Count, so I used this, without `Parameters`. Also, I was in lack of some wiki about CommandArgs, so I decided to look in the source code itself and it give me less than nothing of understanding👍, so yeah.

PackmanDude:
Uhh, what a shame, original fork was deleted by me by accident, so the only way to make a change is to duplicate a PR ;<

PackmanDude:
#2656

PR #2656, opened 2022-08-14
[Dub] Prevent server broadcast to accept an empty message

PackmanDude:
If used without a parameter, the /broadcast (/bc or /say too) command will display (Server Broadcast) only. This PR fixes that.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -1955,12 +1955,15 @@ namespace TShockAPI
 
 		private static void Broadcast(CommandArgs args)
 		{
-			string message = string.Join(" ", args.Parameters);
+			if (args.Parameters.Count > 0)
+			{
+				string message = string.Join(" ", args.Parameters);
 
-			TShock.Utils.Broadcast(
-				"(Server Broadcast) " + message,
-				Convert.ToByte(TShock.Config.Settings.BroadcastRGB[0]), Convert.ToByte(TShock.Config.Settings.BroadcastRGB[1]),
-				Convert.ToByte(TShock.Config.Settings.BroadcastRGB[2]));
+				TShock.Utils.Broadcast(
+					"(Server Broadcast) " + message,
+					Convert.ToByte(TShock.Config.Settings.BroadcastRGB[0]), Convert.ToByte(TShock.Config.Settings.BroadcastRGB[1]),
+					Convert.ToByte(TShock.Config.Settings.BroadcastRGB[2]));
+			}
 		}
 
 		private static void Off(CommandArgs args)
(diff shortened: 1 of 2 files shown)


Arthri (approved):

QuiCM (approved):

TShockAPI/Commands.cs
```diff
@@ -1955,12 +1955,15 @@ private static void SubstituteUserDo(CommandArgs args)
...
 
-			TShock.Utils.Broadcast(
-				"(Server Broadcast) " + message,
-				Convert.ToByte(TShock.Config.Settings.BroadcastRGB[0]), Convert.ToByte(TShock.Config.Settings.BroadcastRGB[1]),
-				Convert.ToByte(TShock.Config.Settings.BroadcastRGB[2]));
+				TShock.Utils.Broadcast(
+					"(Server Broadcast) " + message,
+					Convert.ToByte(TShock.Config.Settings.BroadcastRGB[0]), Convert.ToByte(TShock.Config.Settings.BroadcastRGB[1]),
+					Convert.ToByte(TShock.Config.Settings.BroadcastRGB[2]));
+			}
```
QuiCM, line 1966: Looks good, only suggestion being we could add a response message if someone does do a `/bc` without any parameters:
```suggestion
			}
			else
			{
				args.Player.SendErrorMessage("Please include a message to be broadcasted.");
				args.Player.SendMessage($"Syntax: {"broadcast".Color(Utils.BoldHighlight)} <message>", Color.White);
			}
```

Non-essential though, happy for @Pryaxis/tshock to merge without the above

PackmanDude: Idk, I think, that's how user will expect the behavior: if you give nothing, you get nothing in reply. Could be done with displaying "{command-which-was-used}: Invalid syntax. Try looking up into help page via '/help {command-which-was-used}'". It would be a mess, if every command had few more lines of code for this message specially, method called DisplayHelpMsg() for example, would be useful to include in if-statement directly, with ternary operator, if possible (depends on code style used). Makes code not so big, but behavior is more friendly.

punchready: The normal behavior would be to give the command syntax help message if parameters are used wrongly.

bartico6: > Idk, I think, that's how user will expect the behavior: if you give nothing, you get nothing in reply. Could be done with displaying "{command-which-was-used}: Invalid syntax. Try looking up into help page via '/help {command-which-was-used}'". It would be a mess, if every command had few more lines of code for this message specially, method called DisplayHelpMsg() for example, would be useful to include in if-statement directly, with ternary operator, if possible (depends on code style used). Makes code not so big, but behavior is more friendly.

While you make a valid point (commands shouldn't reinvent the wheel and display their own syntax message), truth of the matter is most if not all commands kind of do this already - and no feedback whatsoever would be breaking the current convention of "some form of error feedback" being provided. The syntax tree for this command is shallow enough that it can easily be added without worrying about code clutter. Rewiring how command parameters, syntaxes and errors work is an entirely different task that has to be done, but nobody is currently interested in doing that, so things remain as they are for the time being :)

hakusaro: +1 to quake + QuiCM + punch's comments

hakusaro:
@PackmanDude due to a security patch going out tonight, this might be bumped to the release after next.

PR #2675, opened 2022-10-05
Introduce loadout support (with SSC)

drunderscore:


drunderscore:
Merge #2677 _BEFORE_ this PR.

hakusaro:
@drunderscore can you rebase or merge changes from gen-dev?

drunderscore:
@hakusaro Rebased on `general-devel`

hakusaro:
@drunderscore 3277e865ad96ec1b777519812160f3bb835bc4bf on TSAPI has been merged into tshock as of d55f3840e943926ba04115a21d8850c4440bae9c, so if you rebase now this should compile correctly _in theory_

drunderscore:
@hakusaro Rebased, looks like it properly compiled as well on CI.

TShockAPI/TSPlayer.cs
```diff
@@ -594,7 +600,126 @@ public bool HasHackedItemStacks(bool shouldWarnPlayer = false)
...
+						item.netDefaults(loadout1Armor[index].netID);
+						item.Prefix(loadout1Armor[index].prefix);
+						item.AffixName();
+
+						if (loadout1Armor[index].stack > item.maxStack || loadout1Armor[index].stack < 0)
+						{
+							check = true;
+							if (shouldWarnPlayer)
+							{
+								SendErrorMessage("Stack cheat detected. Remove Loadout 1 item {0} ({1}) and then rejoin.", item.Name, loadout1Armor[index].stack);
```
hakusaro: Blessed be the day we just ban people on sight instead of telling them exactly what's wrong

hakusaro (dismissed):
For the most part this lgtm, but I'd really love another set of eyes on this besides mine.
lgtm_8

drunderscore:
@hakusaro Sorry to bother, but hoping you could re-review/approve this -- found an issue with UUID login, because we can't broadcast packets to a connecting client that soon.

For changes between, see (ignoring the rebased files):
https://github.com/Pryaxis/TShock/compare/1f499769f70f3d06d7c6d03e88a52a35c81927bb..5c7c612ff019d6ad283617f769ffe198ce89e015#diff-196da5adc1c1e1f703e8210074c7a7c29db59b55779ecb9521a7d06fda2d2d10L477

hakusaro:
Can you press the button?

5C5450DD-A2E8-43F9-9156-B0F23B1197EF

hakusaro (dismissed):
lgtm?

I think someone else should review this still yo

punchready:
If I'm not mistaken, this will not properly update loadouts in `TSPlayer.PlayerData` when a player changes loadouts, because the game does not send slot updates for the loadouts and instead relies on the server swapping the items on its own.
`CloneLoadouts(Main.clientPlayer);` is called after swapping loadouts, which updates the local client clone to not send changed slots for loadouts and the inventory on the next game tick.
Basically, the server has to update `TSPlayer.PlayerData` on its own when a loadout switch is received.

drunderscore:
I've addressed the issue mentioned above by emulating what `Player.TrySwitchingLoadout` does onto `PlayerData.inventory`.

I've also removed the different slot count constants for each loadout index in `NetItem`, and now only count the number of armor slots (`LoadoutArmorSlots`) and the number of dye slots (`LoadoutDyeSlots`) in a loadout, for all loadouts (all loadouts have the same number of armor and dye slots).

I've also added a check to ensure that loadout indexes are in-bounds.

hakusaro (dismissed):

hakusaro:
@punchready can you review this?

punchready:
There is still an issue, which arises from the fact that the game will straight up ignore the loadout sync packet if the player is CCed or currently using an item. This could easily happen when logging in way after joining the server when being webbed if not logged in, because those two conditions are mutually exclusive, so the player is always either able to use items and thus ignore the packet or cannot use items (CCed) and also ignores the packet. If the packet is ignored, all hell breaks loose: if the client joined a server with slot A selected in the character file, and then logs into an account with slot B selected on the server, the client will still have slot A selected but the server will send it the items for slot B.

\- Other than that, lgtm.

TShockAPI/Commands.cs
```diff
@@ -779,6 +779,29 @@ private static void AttemptLogin(CommandArgs args)
 				return;
 			}
 
+			// We need to emulate the checks done in Player.TrySwitchingLoadout, because otherwise the server is not allowed to sync the
```
punchready, line 782: This also needs to be checked in all other locations where the player is logged in with ssc enabled, thinking of automatic uuid login for example.

TShockAPI/Commands.cs
```diff
@@ -779,6 +779,29 @@ private static void AttemptLogin(CommandArgs args)
...
+			// The player must not be dead, using an item, or CC'd to switch loadouts.
+			if (args.TPlayer.dead)
+			{
+				args.Player.SendErrorMessage("You cannot login whilst dead.");
+				return;
+			}
+
+			// This shouldn't ever be reached, because Bouncer rejects PlayerAnimation packets whilst disabled, but we'll leave it here for
+			// completeness.
+			if (args.TPlayer.itemTime > 0 || args.TPlayer.itemAnimation > 0)
```
punchready, line 797: This is probably not enough, because the client could start using an item right before it receives the packet. I think that generally it is not feasible to check if the loadout packet will succeed because that just has too many edge cases.

drunderscore:
My solution to the issue above was to prevent login whilst under those effects unfortunately. However:
- Logging in whilst dead is currently broken on `general-devel` as it stands, so preventing it is not a stretch, as it would have to be fixed anyway.
- The item time/item animation checks never come into play, as Bouncer rejects the `PlayerAnimation` packet whilst disabled.

The only significant change here then the CC'd checks, which checks for frozen, webbed, or stoned. I believe this to be an acceptable breaking change.

If desired, it could perhaps be done that we sync the player's buffs without those 3, sync loadout index, and then re-sync their original buffs with CCing ones... but that seems a bit over-engineered for a problem that I cannot imagine happening all too often or easily, unless I mis-understand how TShock or other plugins normally operate.

punchready:
Being disabled while logging in will happen almost all the time on servers that use dimension switching because those oftentimes disable the player in some way during the switch (on top of the client being able to use items during the switch and cancel the packet).

TShockAPI/Commands.cs
drunderscore: It doesn't make sense anyway as I realize, because even if the server rejects the animation packet, the client will still set them and use the item regardless.

TShockAPI/Commands.cs
drunderscore: I guess you can log in dead? Would we be forced to respawn you or delay the load then?

punchready: I'm not really sure, i personally removed the disabling debuffs on server switching as a bandaid fix and just pray that the player doesn't use items...
There is also the option of using this weird "cant use items" debuff, but that one can be prevented clientside with immunity accessories...

hakusaro:
> Being disabled while logging in will happen almost all the time on servers that use dimension switching because those oftentimes disable the player in some way during the switch (on top of the client being able to use items during the switch and cancel the packet).

Honestly, what would you suggest? We can't really control their code, and without SSC support we're kinda in a holding pattern.

punchready:
With purely tShock it probably works "good enough" like this (after the dead check and stuff has been applied to the other login places), but this should still be addressed in the future.
Servers with proxies can also deal with all of the extra checks and avoiding disabling debuffs on their own at critical points like switching dimensions.
And for example for me, even without any mitigation attempts at all and having a proxy that uses disabling debuffs, it only happened a few times over many days.
Also, servers have been live like this for quite a while now and nothing has imploded yet.

drunderscore:
As discussed, this is about as good as can be done unfortunately... I've added some `FIXME`s to describe the issues we'll be merging.

hakusaro (approved):

PR #2677, opened 2022-10-05
Use a string interpolation and escape single quotes when escaping tables

drunderscore:
This should have previously caused issues, but it didn't, and now it does.
Modernize it by using a string interpolation, and properly escaping the single quotes (which you apparently have to do?)

This was noticed when implementing #2675, as the TShock database was regenerated with the new SSC columns, but occasionally failed because of the lack of escapes.


hakusaro:
@drunderscore can you rebase also you can't just ignore the changelog

drunderscore:
@hakusaro Rebased and added changelog entry.

hakusaro (approved):
lgtm_trick

seems legit

Arthri:
Doesn't this run a query that's structured like this: `CREATE TABLE \'TABLENAME\' ...`? That's not valid SQLite afaik

drunderscore:
@Arthri I just checked in `sqlite3` -- it is valid.
Notice that the query actually being created here would be `CREATE TABLE 'bruh'`, not `CREATE TABLE \'bruh\'` -- the backslashes are escapes for the single quotes in C#, which are apparently required, as I said in the PR

Arthri (approved):

PR #2688, opened 2022-10-06
Fix #2674 - Change the use of current directory to match the server a…

### LINKED ISSUE #2674: Change Environment.CurrentDirectory (follow up from #2670)
This is a follow-up item to track @Rozen4334's comment from #2670 (since we're gonna merge that for logistic reasons around CI, nuget, etc) before addressing:

Screen Shot 2022-10-05 at 02 08 42

SignatureBeef:
…ssembly location

This will allow dependencies to be resolved beside the assembly as intended when startup scripts are used outside the working directory. I am not sure why or how many people would do this, as TSAPI wont load plugins (existing issue), but this is a step in the direction of opening up to allow considerations for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This is the rolling changelog for TShock for Terraria. Use past tense when addin
 * SetDefaultsEventArgs now includes a nullable ItemVariant instance. (@SignatureBeef)
 * Use a string interpolation and escape single quotes when escaping tables (@drunderscore)
 * Removed obsolete resource files `TShockAPI/Resources.resx` and `TShockAPI/Resources.Designer.cs`. (@Arthri)
+* Plugins and ./bin dependencies are now loaded relative to the launcher, this improves the use of startup files (@SignatureBeef)
 
 ## TShock 4.5.18
 * Fixed `TSPlayer.GiveItem` not working if the player is in lava. (@gohjoseph)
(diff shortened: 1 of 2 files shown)


hakusaro (changes requested):
Since this is a change in behavior can you update the changelog?

SignatureBeef:
> Since this is a change in behavior can you update the changelog?

change of behavior would be https://github.com/Pryaxis/TSAPI/pull/220 since it will change how the existing plugins are found. prior to that PR, plugins wont load if the launch script is in another location (at least in my own testing).

this one, is essentially a update to a new file since last release, and ./bin is expected to be beside the assembly in TShock 5, so this one hasnt seen a actual release yet and thus not really a change in behavior unless youre talking about pre-release builds...?

i can instead add a note for the TSAPI one, if you like? i can see the TSAPI one slipping through with autoupdates, so happy to mix in here since it's a result from looking at this one.

hakusaro:
> change of behavior would be https://github.com/Pryaxis/TSAPI/pull/220 since it will change how the existing plugins are found. prior to that PR, plugins wont load if the launch script is in another location (at least in my own testing).

that's what I mean yeah.

all PRs require changelog updates unless they're done by robots or change things that aren't related to tshock (i.e., github actions)

hakusaro:
@SignatureBeef it's merged into tshock as of 3d7ba2b2d504f4246b95bf05025a246e72c2f039, so you can update CL here

hakusaro (approved):

hakusaro (approved):

PR #2699, opened 2022-10-07
 Allow Blood Butcherer and Shimmer buffs to be applied to NPCs by players

drunderscore:
See: https://terraria.wiki.gg/wiki/Blood_Butcherer

Edit: Shimmer is applied by players to NPCs for 100 ticks, each tick the NPC is in the liquid. It is also done by multiple players at once.


hakusaro (dismissed):

AxisKriel:
I tested this and there is an additional issue when it comes to Shimmer...

The following lines performs an additional check if the NPC in question is NOT the Guide or the Clothier (because their Voodoo Dolls allow players to treat them as enemies):
https://github.com/Pryaxis/TShock/blob/07f65fc1c653ba29e1429a10fb890cc718f2644b/TShockAPI/Bouncer.cs#L1739-L1746

To prevent players from being kicked, you need to add Shimmer to the list of "whitelist" buffs.

drunderscore:
@AxisKriel Totally correct, I've implemented the proper checks for town NPCs now, thanks!

hakusaro (approved):

PR #2728, opened 2022-10-18
Start to update documentation

hakusaro:
This is the foundation of the new documentation system. This is incomplete, but I want to merge it anyway. Let me explain a bit of what I'm thinking and why I think that's the correct decision.

We've had a lot of lack of success with Readme. Their offering is nice on the surface but it's not easily tied to code versions and it got behind quickly. Submitting changes was cumbersome. The existing documentation is wrong and obsolete in a lot of ways. Based on this, I think I'd rather we just ship a new documentation system, in an incomplete state, and get it to a completed state for TShock 5. The `general-devel` branch is for general development. In that sense, I think the floor should be opened for anyone who wants to work on documentation. Many people have already stated that they want to help. But having this in a side branch is pointless and precludes that.

In `h/docs`, I tried to rewrite our documentation using Obsidian Publish. While I think this was a well-intentioned effort, the issue is that Obsidian Publish fundamentally isn't friendly for most people. You really have to install Obsidian to get the most out of it, and publishing is non-trivial. It's not even something that's like 2 clicks -- you have to select all of the files and hit publish, manually. It's a huge pain.  I like Obsidian, but it's not the right tool for the job here. The main / core issue is that Obsidian really has a problem with versioning. Everything is just online, and publishing changes overwrites the existing ones.

----

This iteration solves a lot of these problems. The documentation is in the repo. It's just markdown, and you can edit it with any editor. GitHub Pages automatically compiles markdown and serves it. Since it's versioned, when we cut a release, it'll be fairly trivial to just have a separate branch for "stable" versions of the docs (to map with a stable release) and unstable / edge docs (to map with general-devel).

Because anyone can just PR changes, it'll be better too, because you just need like, an editor (github.dev?) and github flow.

This PR specifically has the following things that I'd like to highlight, because I intend to crunch this branch down when merged so it's just one commit:

1. It updates the readme at the base of the repo to be focused on building / hacking on TShock on .NET 6.
2. It moves the "user specific" install instructions into the `docs` folder README. This readme is rendered live on ikebukuro.tshock.co.
3. It adds updated install instructions and upgrade instructions for TShock 5.
4. It rebuilds the command line flag documentation and the tile provider documentation (previously, undocumented, actually), so it's already better than readme for that.

**Because general-devel is already on TShock 5**, the existing readme is already out-of-date. That's why I want to merge this now -- to fix the developer documentation at minimum.

If approved, this will be squashed into one commit and merged. I'm going to work on this throughout the week with the goal of having readme fully replaceable by next week ish. Does that sound good?


hakusaro:
@Pryaxis/tshock if someone can review this so I can merge it in the next couple hours (if you agree), I'd really appreciate it. I re-summarized the purpose of this in the PR description, and intend to squash all of these commits into one before merge.

hakusaro:
I should clarify that in order to keep everyone as sane as possible (ASAP), this is really a conceptual review. If we're good with this approach, I won't PR other markdown files in the initial push to update docs, because that's a lot of nothing to review.

I'd like to get to a point where everyone is PRing docs and stuff, but in the initial push it's probably better to focus on just getting words on paper, since nobody likes doing this.

SignatureBeef (approved):
Looks pretty good to me so far - nothing jumps out to be incorrect either. Everything seems pretty reasonable and inline to current contributing practices too, so multiple can work on this.
I will try and get some time to add to the tile provider sections too, so merging is in my interest as well.

PR #2733, opened 2022-10-19
Add localization support for console spam reduction

KawaiiYuyu:


docs/changelog.md
```diff
@@ -15,6 +15,7 @@ Use past tense when adding new entries; sign your name off when you add or chang
   * If there is no section called "Upcoming changes" below this line, please add one with `## Upcoming changes` as the first line, and then a bulleted item directly after with the first change. -->
 
 ## Upcoming changes
+* Added localization support for console spam reduction. (@KawaiiYuyu)
```
drunderscore: This entry should be moved to the bottom, to try to keep changes chronological -- even though it's related to the change below, that was many months ago :^)

TShockAPI/Modules/ReduceConsoleSpam.cs
```diff
@@ -32,6 +33,13 @@ public class ReduceConsoleSpam : Module
 	/// </summary>
 	private string _lastStatusText = null;
 
+	private static readonly string ResettingObjectText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.47");
+	private static readonly string LoadingText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.51");
+	private static readonly string SettlingText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.27");
+	private static readonly string SavingText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.49");
+	private static readonly string ValidatingText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.73");
+	private static readonly string FinalizingText = LanguageManager.Instance.GetTextValue("LegacyWorldGen.87");
```
drunderscore: Although these being static might not actually cause issues, them being non-static makes more sense to me, and stays in-line with the existing field above.

drunderscore (changes requested):
I've tested this in English and it still works the same as it previously did. I'll test in other languages once I figure out how to convince `dotnet` to change locale.

Awesome patch though, sorry that it hadn't properly supported other locales originally :)

TShockAPI/Modules/ReduceConsoleSpam.cs
KawaiiYuyu: > I've tested this in English and it still works the same as it previously did. I'll test in other languages once I figure out how to convince `dotnet` to change locale.
> 
> Awesome patch though, sorry that it hadn't properly supported other locales originally :)

Thanks for reviewing, you can pass a cli flag like `-langauge de-DE` to set the locale, which is handled by vanilla code. All available locale codes are: en-US, de-DE, it-IT, fr-FR, es-ES, ru-RU, zh-Hans, pt-BR, pl-PL.

drunderscore (approved):
Awesome! Totally works for languages other than English now, LGTM!

hakusaro (approved):

hakusaro:
@KawaiiYuyu thank you for your contributions, and welcome! ✨ ❤️

PR #2736, opened 2022-10-20
Add i18n system

sowelipililimute:
automatic extraction CI is borked pending https://github.com/perpetualKid/GetText.NET/issues/43; but otherwise functional. contains translations of the handlecommand as a starting point and because the system doesn't function with empty everything.

### DIFF
--- a/TShockAPI/I18n.cs
+++ b/TShockAPI/I18n.cs
@@ -1,0 +1,46 @@
+/*
+TShock, a server mod for Terraria
+Copyright (C) 2022 Janet Blackquill
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+using System;
+using System.Globalization;
+using System.IO;
+using GetText;
+
+namespace TShockAPI
+{
+	static class I18n {
+		static string TranslationsDirectory => Path.Combine(AppContext.BaseDirectory, "i18n");
+		static CultureInfo TranslationCultureInfo
+		{
+			get
+			{
+				// cross-platform mapping of cultureinfos can be a bit screwy, so give our users
+				// the chance to explicitly spell out which translation they would like to use.
+				// this is an environment variable instead of a flag because this needs to be
+				// valid whether the passed flags are in a sane state or not.
+				if (Environment.GetEnvironmentVariable("TSHOCK_LANGUAGE") is string overrideLang)
+				{
+					return new CultureInfo(overrideLang);
+				}
+				return CultureInfo.CurrentUICulture;
+			}
+		}
+		/// <value>Instance of a <c>GetText.Catalog</c> loaded with TShockAPI translations for user's specified language</value>
+		public static Catalog C = new Catalog("TShockAPI", TranslationsDirectory, TranslationCultureInfo);
+	}
+}
--- a/TShockLauncher/TShockLauncher.csproj
+++ b/TShockLauncher/TShockLauncher.csproj
@@ -24,6 +24,7 @@
 		<PackageReference Include="MySql.Data" Version="8.0.31" />
 		<PackageReference Include="Microsoft.Data.Sqlite" Version="6.0.10" />
 		<PackageReference Include="ModFramework" Version="1.1.6" GeneratePathProperty="true" /> <!-- only used to extract out to ./bin. -->
+		<PackageReference Include="GetText.NET" Version="1.6.6" /> <!-- only used to extract out to ./bin. -->
 
 		<!-- the launcher doesnt need the direct OTAPI reference, but since PackageReference[ExcludeFromSingleFile] doesnt work, exclude the assets and copy manually -->
 		<PackageReference Include="OTAPI.Upcoming" Version="3.1.14" ExcludeAssets="all" GeneratePathProperty="true" />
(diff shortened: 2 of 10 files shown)


TShockLauncher/TShockLauncher.csproj
```diff
@@ -41,6 +41,30 @@
...
+		Name="GenerateMOFiles"
+		DependsOnTargets="CheckMsgfmtCallable"
+		AfterTargets="PostBuildEvent;Publish"
+		Inputs="..\i18n\**\*.po"
+		Outputs="$(OutDir)i18n\**\*.mo">
+		<ItemGroup Condition="'$(MsgfmtExitCode)' == '0'">
+			<POFiles Include="..\i18n\**\*.po" />
+		</ItemGroup>
+		<MakeDir Directories="$(OutDir)i18n/%(POFiles.RecursiveDir)" Condition="'$(MsgfmtExitCode)' == '0'" />
+		<Exec Command="msgfmt -o $(OutDir)i18n/%(RecursiveDir)%(Filename).mo @(POFiles)" Outputs="$(OutDir)i18n\**\*.mo" Condition="'$(MsgfmtExitCode)' == '0'">
```
Arthri, line 65: these all have `Condition="'$(MsgfmtExitCode)' == '0'"`. I think it should be put onto the target instead

sowelipililimute: hoisting it up to the target doesn't work

Arthri: ah, yeah it doesnt

TShockAPI/I18n.cs
```diff
@@ -0,0 +1,27 @@
...
+				// this is an environment variable instead of a flag because this needs to be
+				// valid whether the passed flags are in a sane state or not.
+				if (Environment.GetEnvironmentVariable("TSHOCK_LANGUAGE") is string overrideLang)
+				{
+					return new CultureInfo(overrideLang);
+				}
+				return CultureInfo.CurrentUICulture;
+			}
+		}
+		public static Catalog Catalog = new Catalog("TShockAPI", TranslationsDirectory, TranslationCultureInfo);
```
hakusaro: The only thing I'd suggest is documenting this. And, honestly, since this is going to be called everywhere, you might consider calling this something small like C or something. Lotta words to spell out catalog.

hakusaro (commented):
I tested this with `TSHOCK_LANGUAGE=tok dotnet run --project TShockLauncher` and verified that I do indeed get results. I'm happy to go with this solution, because you basically implemented all of my wishes...so...thanks!

A few minor things:

1. It would be really nice if the changelog is updated. I can handle that though, if you don't want to.
2. Umm, since you wrote new files...if you intend to add them to the project, are you cool with throwing GPL license headers on them too, and making them match the rest of the repo? Technically it's just your copyright, so you can just name them, but I wanna make sure you're cool with gpl v3 or later?
3. Do...you actually want to keep `tok` around as a language? I would prefer to keep the file around even if you don't.
4. Can you broadly document the steps required to update the template? Sorry for being new to this -- is that even required? Maybe not?

Screen Shot 2022-10-19 at 22 37 55

TShockAPI/I18n.cs
hakusaro: I say this but I mean with XML docs, because it'll add a warning if not.

hakusaro:
I will test this on windows before I commit to merging this though, that's the one hangup

TShockAPI/I18n.cs
Arthri: I think you can also pull this file <https://github.com/VitaliiTsilnyk/NGettext/blob/master/samples/NGettextShortSyntax.cs> then `global using static` on it, effectively allowing you to use the short C-style version of gettext throughout the TShock project

hakusaro: That might be ideal

sowelipililimute: not compatible with the gettext.net extractor; will look into terser options later

sowelipililimute:
image

hakusaro:
image

Yeah it works alright!

hakusaro (approved):
lgtm! thank you for all of your hard work on this!

Arthri:
Is the `.gitkeep` needed?

SignatureBeef (commented):
> Is the `.gitkeep` needed?

pending the decision of this (idc), code looks fine to me, and runs as expected on osx64(intel)+win64

one thing though - i assume this is expected: `'msgfmt' is not recognized as an internal or external command, operable program or batch file.` (edit: clarification, during build/publish)

happy to approve after clarification of the above, or if previously addressed @hakusaro.

hakusaro:
> one thing though - i assume this is expected: 'msgfmt' is not recognized as an internal or external command, operable program or batch file. (edit: clarification, during build/publish)

Does it stop you from building or publishing on windows?

It's expected that you'll see some error if you don't have it installed, but it shouldn't actually affect anything important. The intent is that you can still develop tshock if you don't have GetText, but if you do, it'll be fine.

hakusaro:
> Is the .gitkeep needed?

Even if it's not needed, it's not actually something to stop anything for imho

SignatureBeef:
> Is the `.gitkeep` needed?

pending the decision of this (idc), it looks fine to me, and runs as expected on osx64(intel)+win64
i assume `'msgfmt' is not recognized as an internal or external command, operable program or batch file.` is intended

> > one thing though - i assume this is expected: 'msgfmt' is not recognized as an internal or external command, operable program or batch file. (edit: clarification, during build/publish)
> 
> Does it stop you from building or publishing on windows?
> 
> It's expected that you'll see some error if you don't have it installed, but it shouldn't actually affect anything important. The intent is that you can still develop tshock if you don't have GetText, but if you do, it'll be fine.

nah build and publish completes fine, just seen it and figured id check - thats all.

SignatureBeef (approved):

Arthri:
> one thing though - i assume this is expected: `'msgfmt' is not recognized as an internal or external command, operable program or batch file.` (edit: clarification, during build/publish)

I believe this is intentional, the line that runs msgfmt explicitly ignores its exit code, so preventing the build from stopping

However, it does cause mo files to not be generated on windows, but this is acknowledged on the [Discord](https://discord.com/channels/479657350043664384/711282444933791774/1032514347181809745):
> Q(me): i havent checked but doesnt the build silently fail and not generate mo if it can't find msgfmt?
> A(author): generating mo is optional yea
> since localisation isn't really required for a dev build

PR #2747, opened 2022-10-22
Include geoip.dat file again

SignatureBeef:
This should allow the GeoIP.dat file to be found, and used the same way it used to on TShock 4. It appears I had missed including this in the upgrade, and simply allowing it to copy to the output will allow the system to start working again. Confirmed by enabling in config and joining via mobile (external network) and my country has appeared, N/A otherwise (local network)
Note: only ipv4 was tested, and both debug and publish variant on osx.

Image below of my test (external ip intentionally excluded)

Screen Shot 2022-10-22 at 9 39 47 pm

### DIFF
--- a/TShockLauncher/TShockLauncher.csproj
+++ b/TShockLauncher/TShockLauncher.csproj
@@ -20,6 +20,13 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<None Include="..\prebuilts\GeoIP.dat">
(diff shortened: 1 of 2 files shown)


hakusaro (approved):

hakusaro:
Thank you so much!

drunderscore:
wait I did not mean to do that I was just on my phone looking...

my bad!!

PR #2748, opened 2022-10-22
Dockerize TShock

pcider:
Not sure if this should be added to README. We could also use Github Actions to push this to Docker Hub.


Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# and run:
+#  docker run -p 7777:7777 -p 7878:7878 \
+#             -v <save path>:/tshock \
+#             -v <world path>:/worlds \
+#             -v <plugin path>:/plugins \
+#             --rm -it tshock -world /worlds/<world file> <flags>
+
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS builder
+
+ARG ARCH=linux-x64
```
hakusaro: Can you make this compatible with `docker buildx` for [multi-platform images](https://docs.docker.com/build/building/multi-platform/) instead? E.g., `docker buildx build --platform linux/arm64, linux/amd64`?

If possible I would suggest changing the source image to something that already is platform agnostic. Then, just build the target on the right version?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /TShock/TShockLauncher
+RUN dotnet publish -o output/ -r ${ARCH} -f net6.0 -c Release -p:PublishSingleFile=true --self-contained false
+
+# Runtime image
+FROM mcr.microsoft.com/dotnet/runtime:6.0 AS runner
+WORKDIR /server
+COPY --from=builder /TShock/TShockLauncher/output ./
+RUN mkdir -p /tshock /worlds /plugins
+
+VOLUME ["/tshock", "/worlds", "/plugins"]
```
hakusaro, line 35: Can you add documentation in `docs` (updating `_sidebar.md`) for how one might actually access their config, plugins, and worlds from the docker build? In general, can you maybe migrate the comment(s) above to that?

hakusaro (commented):
Overall this lgtm, but I would suggest at least adding compat with `docker buildx build` if possible because passing the architecture on the command line is kind of non-standard to my eyes. I would also suggest that it's going to build on whatever it's running on natively first, so it's probably safe to just use buildkit and not tell .net which arch to build for.

hakusaro:
> Docker Hub

They kinda are a wet blanket these days; I'd be open to exploring all of the available hosts just because I know Docker Hub tried (and maybe succeeded?) to add restrictions in terms of outgoing bandwidth/usage.

pcider:
Added docs and support for docker multi-platform images. Imma goto sleep now

PR #2750, opened 2022-10-23
Add ability for items given to players to be inserted directly into their inventory instead of spawned as an item drop

sowelipililimute:


punchready:
Beware that this can lead to item loss in case the client fills an empty slot before the packet reaches it.

sowelipililimute:
in practice @ penguin games we have far more cases of item loss from spawning dropped items than direct insertion. (i can't recall *any* case of direct insertion losing items, TBH)

either way, this is optional.

TShockAPI/TSPlayer.cs
```diff
@@ -1508,6 +1508,139 @@ public bool GiveItemCheck(int type, string name, int stack, int prefix = 0)
...
+				for (int i = 0; i < 10; i++)
+					if (depleted(item = GiveItem_FillEmptyInventorySlot(item, i)))
+						return;
+
+			for (int i = num - 1; i >= 0; i--)
+				if (depleted(item = GiveItem_FillEmptyInventorySlot(item, i)))
+					return;
+
+			// oh no, i can't give... guess i gotta spill it on the floor
+			GiveItemByDrop(type, stack, prefix);
```
pcider, line 1560: Should be using overflowed `item`? 
```suggestion
			GiveItemByDrop(item.type, item.stack, item.prefix);
```

pcider: Also maybe could handle going into void vault, though that's just a nice-to-have.

pcider (changes requested):
PR looks good to me. 

For the item loss issue, do you know why it happens? Couldn't that be solved by using `noGrabDelay` and `playerIndexTheItemIsReservedFor`?

sowelipililimute:
re. item loss, it's just that players aren't necessarily guaranteed to be still enough to pick up an item, even if it spawns directly on them. also, latency makes it possible for an item to spawn way off a player's current position.

hakusaro:
@pontaoski thank you for the context!

hakusaro (changes requested):
Would you be willing to consider moving this into the SSC config instead, and then regenerating the docs with `-dump`? Since this is mostly an SSC thing, it should probably be an SSC-config thing, imho.

sowelipililimute:
done

hakusaro:
If I’m completely honest, the biggest problem with this is that clearly, the code is copied directly from decompiled Terraria code. While not necessarily something we’re opposed to inherently, it does pose some problems from a licensing perspective. Is there any way we can just call game methods directly instead? I haven’t looked too far into these but it doesn’t seem implausible that we can’t just tap them directly.

The other big concern is that if these methods get changed in the future, updating them becomes something that we have to know to do, and becomes subtly difficult.

I know just as much as anyone else that we used to heavily use game code like this. It wouldn’t surprise me to learn that we may have instances of this technique still laying around. But the issue is that in hindsight, and given that this is a 2022 “post-relogic DMCA world” it would probably be best if we stray away from reimplementing their methods like this.

Is there any solution you can think of that doesn’t encumber us this way? If the issue is something stupid like accessors, we could just change the accessors on the methods to be publicly callable. I’m sure this is something we could tap OTAPI for, too.

pcider:
Calling the GrabItems directly and syncing the changed items could be an option. 

~~Another possible workaround for the item pickup would be to spawn an item with the velocity of the player, setting `noGrabDelay` to true with `playerItemIsReservedFor` set to the player's index. (I have not tested this)~~ Nevermind, this does not work with a 100ms lag

sowelipililimute:
made the procedure more distinct from how terraria accomplishes it (but either way, algorithims/procedures themselves aren't copyrightable, just the actual manifest code itself.)

pcider:
Sorry I didn't raise this issue further but shouldnt my suggested change be implemented? It would cause original stack of items to be dropped on the ground instead of the leftover items that had been calculated after filling the player's inventory.

PR #2753, opened 2022-10-23
Update Bouncer / OnNPCAddBuff

KawaiiYuyu:
- Increase the max allow time for Frostburn2 buff to 1200 bacause of the Elf Melter.

- The melee weapon Flymeal(ItemID: 5129) is able to hurt town NPCs. And there are some items could make a generic melee weapon inflicts debuff on NPC such as flasks. So we should allow those debuffs to be applied on town NPCs.

- Add a permission to skip this check, which can be used for admins doing some hack or as a workaround when there's an update.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1737,7 +1737,7 @@ namespace TShockAPI
 				{
 					bucket = 6;
 				}
-				else if (selectedItemType == ItemID.BottomlessHoneyBucket 
+				else if (selectedItemType == ItemID.BottomlessHoneyBucket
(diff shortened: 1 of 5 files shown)


hakusaro:
Can you run TShock with the `-dump` flag and commit the changes to the config file mapping?

KawaiiYuyu:
> Can you run TShock with the `-dump` flag and commit the changes to the config file mapping?

Done and I added a line to create `docs` directory before the dump.

hakusaro:
> Done and I added a line to create docs directory before the dump.

It already exists though — *and* you modified the changelog inside it! :P

hakusaro (approved):

PR #2754, opened 2022-10-23
Handle vanilla `-lang` and `-language` CLI flags for I18n

KawaiiYuyu:
In most cases, user would use the same language for both Terraria and TShock if the language is officially supported by this game. So we can just use those 2 flags to configure the culture.

Even if a user really want to use different langs they can also use envvars to override it.

### DIFF
--- a/TShockAPI/I18n.cs
+++ b/TShockAPI/I18n.cs
@@ -21,7 +21,10 @@ global using static TShockAPI.I18n;
 using System;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using GetText;
+using Terraria.Initializers;
+using Terraria.Localization;
 
 namespace TShockAPI
 {
@@ -29,8 +32,7 @@ namespace TShockAPI
 		static string TranslationsDirectory => Path.Combine(AppContext.BaseDirectory, "i18n");
 		static CultureInfo TranslationCultureInfo
 		{
-			get
-			{
+			get {
 				// cross-platform mapping of cultureinfos can be a bit screwy, so give our users
 				// the chance to explicitly spell out which translation they would like to use.
 				// this is an environment variable instead of a flag because this needs to be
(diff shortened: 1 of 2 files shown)


Arthri:
TShock already supports this through envvars https://github.com/Pryaxis/TShock/blob/4985c168637a70d58a511bf58156b7ad5f09960b/TShockAPI/I18n.cs#L38

TShockAPI/I18n.cs
```diff
@@ -39,6 +41,24 @@ static CultureInfo TranslationCultureInfo
 				{
 					return new CultureInfo(overrideLang);
 				}
+
+				static CultureInfo Redirect(CultureInfo cultureInfo)
+					=> cultureInfo.Name == "zh-Hans" ? new CultureInfo("zh-CN") : cultureInfo;
+
+				if (Terraria.Program.LaunchParameters.TryGetValue("-lang", out var langArg)
```
sowelipililimute, line 48: does terraria use the same language codes as the i18n system?

KawaiiYuyu:
> TShock already supports this through envvars
> 
> https://github.com/Pryaxis/TShock/blob/4985c168637a70d58a511bf58156b7ad5f09960b/TShockAPI/I18n.cs#L38

yep, but one need to set envvars **and** cli flags to change the whole server's language, I want to make it easier

TShockAPI/I18n.cs
KawaiiYuyu: > does terraria use the same language codes as the i18n system?

we are missing some languages supported by Terraria itself, they are `it-IT`, `pt-BR` and `pl-PL`, we can add them if needed. The only exception is Chinese(Simplified), Terraria uses `zh-Hans` and Crowdin has only `zh-CN`, so I manually converted it. Other than these, I think they are compatible

hakusaro: Can we normalize this on Crowdin maybe?

hakusaro (approved):

PR #2760, opened 2022-10-24
Update i18n

sgkoishi:
https://github.com/perpetualKid/GetText.NET/issues/44

Also some minor fixes pending

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2955,10 +2955,10 @@ namespace TShockAPI
 				}
 				else if (!TShock.Config.Settings.AnonymousBossInvasions)
 				{
-					TShock.Utils.Broadcast(string.Format($"{args.Player.Name} summoned the Empress of Light!"), 175, 75, 255);
+					TShock.Utils.Broadcast(GetString($"{args.Player.Name} summoned the Empress of Light!"), 175, 75, 255);
 				}
 				else
-					TShock.Utils.SendLogs(string.Format($"{args.Player.Name} summoned the Empress of Light!"), Color.PaleVioletRed, args.Player);
+					TShock.Utils.SendLogs(GetString($"{args.Player.Name} summoned the Empress of Light!"), Color.PaleVioletRed, args.Player);
 			}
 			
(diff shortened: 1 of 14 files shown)


sgkoishi:
There are still ~100 `player.Send(Info|Warning|Error)Message("literal string")` left.

hakusaro:
Wow that was fast

hakusaro:
This creates a few word puzzles -- I _really_ wish you'd made separate commits because it would be easier to merge smaller chunks at a time. I'm happy to approve this but I think @pontaoski should review it before it gets merged.

The reason why I was going (quite slowly) through `Commands.cs` was to try to fix the English consistency too at the same time, since it's just one pass. This doesn't change the fact that English strings aren't consistent, and if anything, makes me feel like it's a lower priority (because I won't be able justify doing it as part of the i18n work).

Like, if we merge this now and then make English consistent the strings will definitely change in i18n land. The latest Crowdin PR also has the issue with Russian, which means we can't merge more translations anyway.

hakusaro (dismissed):
(see notes)

hakusaro:
**The main thing I'm worried about is word puzzles, the rest of the PR is ok!**

TShockAPI/GetDataHandlers.cs
```diff
@@ -3523,52 +3523,52 @@ private static bool HandleSpawnBoss(GetDataHandlerArgs args)
 			switch (thingType)
```
sowelipililimute, line 3523: word puzzle, each of these cases should be a complete sentence rather than combining words

sowelipililimute (changes requested):

TShockAPI/GetDataHandlers.cs
sgkoishi: Resolved. The `/spawnboss` and `/tree` are also kind of word puzzle: spawnboss is a total of 120 combinations and tree is a `$"Tried to grow a {lots of tree names}."`.

hakusaro (approved):

hakusaro:
I'm okay with this -- and I'm willing to say we should merge this -- but please avoid code formatting changes in bulk in future pull-requests like this. It's _really_ a lot of lines to review already.

hakusaro (approved):

PR #2758, opened 2022-10-24
Fixed SSC not save `ateArtisanBread`, `usedAegisCrystal`, `usedAegisFruit`, `usedArcaneCrystal`, `usedGalaxyPearl`, `usedGummyWorm`, `usedAmbrosia`, `unlockedSuperCart`, `enabledSuperCart` flags, and Server will correct read them.

hufang360:
Fixed SSC not save `ateArtisanBread`, `usedAegisCrystal`, `usedAegisFruit`, `usedArcaneCrystal`, `usedGalaxyPearl`, `usedGummyWorm`, `usedAmbrosia`, `unlockedSuperCart`, `enabledSuperCart` flags, and Server will correct read them.

PR #2768, opened 2022-10-29
Fixed Server not tracks deaths, Now players can be viewed with the `/death`, `/pvpdeath`, `/alldeath` and `/allpvpdeath` commands.

hufang360:
Fixed Server not tracks deaths, Now players can be viewed with the `/death`, `/pvpdeath`, `/alldeath` and `/allpvpdeath` commands.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2700,16 +2700,6 @@ namespace TShockAPI
 			short numberOfDeathsPVE = args.Data.ReadInt16();
 			short numberOfDeathsPVP = args.Data.ReadInt16();
 			PlayerSpawnContext context = (PlayerSpawnContext)args.Data.ReadByte();
-			// When the player spawn into world, the numberOfDeathsPVE and numberOfDeathsPVP is not correct on the SSC Server.
-			// Send the data agrain fixed them.
-			if (context == PlayerSpawnContext.SpawningIntoWorld && args.Player.IsLoggedIn )
-			{
-				if(numberOfDeathsPVE != args.TPlayer.numberOfDeathsPVE || numberOfDeathsPVP != args.TPlayer.numberOfDeathsPVP)
-				{
-					args.Player.Spawn(PlayerSpawnContext.SpawningIntoWorld);
-					return true;
-				}
-			}
 			if (OnPlayerSpawn(args.Player, args.Data, player, spawnx, spawny, respawnTimer, numberOfDeathsPVE, numberOfDeathsPVP, context))
 				return true;
 
@@ -2746,6 +2736,17 @@ namespace TShockAPI
 				args.Player.Dead = true;
 			else
 				args.Player.Dead = false;
+
+			// When the player spawn into world, the numberOfDeathsPVE and numberOfDeathsPVP is not correct on the SSC Server.
+			// Send the data agrain fixed them.
+			if (context == PlayerSpawnContext.SpawningIntoWorld && args.Player.IsLoggedIn )
+			{
+				if(numberOfDeathsPVE != args.TPlayer.numberOfDeathsPVE || numberOfDeathsPVP != args.TPlayer.numberOfDeathsPVP)
+				{
+					args.Player.Spawn(PlayerSpawnContext.SpawningIntoWorld);
+					return true;
+				}
+			}
 			return false;
 		}
 
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -856,6 +856,14 @@ namespace TShockAPI
 			/// </summary>
 			public int RespawnTimer { get; set; }
 			/// <summary>
+			/// Player's deaths from PVE.
+			/// </summary>
+			public int NumberOfDeathsPVE { get; set; }
+			/// <summary>
+			/// Player's deaths from PVP.
+			/// </summary>
+			public int NumberOfDeathsPVP { get; set; }
+			/// <summary>
 			/// Context of where the player is spawning from.
 			/// </summary>
 			public PlayerSpawnContext SpawnContext { get; set; }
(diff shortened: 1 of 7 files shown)


hakusaro (dismissed):
Thank you so much for your contributions! This looks good to me, but another person needs to review this besides me before merging!

TShockAPI/PlayerData.cs
```diff
@@ -284,6 +288,11 @@ public void RestoreCharacter(TSPlayer player)
 			player.TPlayer.unlockedSuperCart = this.unlockedSuperCart == 1;
 			player.TPlayer.enabledSuperCart = this.enabledSuperCart == 1;
 
+			if (this.numberOfDeathsPVE > 0)
+				player.TPlayer.numberOfDeathsPVE = this.numberOfDeathsPVE;
+			if (this.numberOfDeathsPVP > 0)
+				player.TPlayer.numberOfDeathsPVP = this.numberOfDeathsPVP;
+
```
drunderscore: This looks suspicious... why are only death counts that are greater than `0` allowed to override the player's data?

drunderscore (changes requested):
It seems like logging into a brand-new user (never before logged into) will allow the player to retain the death counter they previously had (be it from our SSC of another user after `/logout`, or the character they are joining with), instead of defaulting to `0`.

Reproduction steps:
- Execute `/user add bruh bruh default` in console
- Join server
- Execute `/login bruh bruh`
- Observe that the death counts from the joining character are preserved.

Otherwise, once the user has existed, it seems it counts properly, and logging in and out to different users properly updates the count.

TShockAPI/PlayerData.cs
hufang360: > It seems like logging into a brand-new user (never before logged into) will allow the player to retain the death counter they previously had (be it from our SSC of another user after `/logout`, or the character they are joining with), instead of defaulting to `0`.
> 
> Reproduction steps:
> 
> * Execute `/user add bruh bruh default` in console
> * Join server
> * Execute `/login bruh bruh`
> * Observe that the death counts from the joining character are preserved.
> 
> Otherwise, once the user has existed, it seems it counts properly, and logging in and out to different users properly updates the count.

Thank you so much. I'll fixed it.

hufang360: > This looks suspicious... why are only death counts that are greater than `0` allowed to override the player's data?

Because, the `this.numberOfDeathsPVE` variable may be null, When "HandleSpawn" event can read the value is 9985.
It isn't always good, i need make change.

Refer: https://github.com/Pryaxis/TShock/pull/2758#event-7695789368

TShockAPI/PlayerData.cs
```diff
@@ -140,6 +142,8 @@ public void CopyCharacter(TSPlayer player)
 			this.usedAmbrosia = player.TPlayer.usedAmbrosia ? 1 : 0;
 			this.unlockedSuperCart = player.TPlayer.unlockedSuperCart ? 1 : 0;
 			this.enabledSuperCart = player.TPlayer.enabledSuperCart ? 1 : 0;
+			this.numberOfDeathsPVE = player.TPlayer.numberOfDeathsPVE;
+			this.numberOfDeathsPVP = player.TPlayer.numberOfDeathsPVP;
```
hufang360, line 146: I change this, It seems work. Can you review it agrain. @drunderscore

hakusaro (dismissed):

hakusaro:
@hufang360 I'm happy with this but it probably won't make 5.1, likely 5.2. This is because of the newest terraria release. Sorry.

TShockAPI/GetDataHandlers.cs
```diff
@@ -2687,9 +2697,20 @@ private static bool HandleSpawn(GetDataHandlerArgs args)
...
-
-			if (OnPlayerSpawn(args.Player, args.Data, player, spawnx, spawny, respawnTimer, context))
+			// When the player spawn into world, the numberOfDeathsPVE and numberOfDeathsPVP is not correct on the SSC Server.
+			// Send the data agrain fixed them.
+			if (context == PlayerSpawnContext.SpawningIntoWorld && args.Player.IsLoggedIn )
+			{
+				if(numberOfDeathsPVE != args.TPlayer.numberOfDeathsPVE || numberOfDeathsPVP != args.TPlayer.numberOfDeathsPVP)
+				{
+					args.Player.Spawn(PlayerSpawnContext.SpawningIntoWorld);
+					return true;
```
drunderscore: Dangerous change here -- when `SpawningIntoWorld` with non-matching death counts (which is likely), we will no longer call the `OnPlayerSpawn` handler, which is very much unexpected and will cause issue with plugins.

This entire check should be moved below the `OnPlayerSpawn` handler call.

drunderscore (changes requested):
The issue of newly-created users inheriting the previous death count is fixed now, but something is still wrong unfortunately...

Dying will properly increment the death counter (executing `/deaths` whilst dead shows the correct number), but once the player respawns, their local death count will override the server's.

hakusaro:
@drunderscore can you re-review this one?

drunderscore:
Something still isn't correct -- the same issue as above persists, where in death the counter holds the expected value, but once the player respawns, their local player's death count is respected. I've attached a video to show exactly what I'm talking about.

- Show that before login that local player has x deaths
- Login to new user
- Show new user has 0 deaths
- Die
- Before respawn, show user has 1 death
- After respawn, show user does not still have 1 death as expected.

[Screencast from 11-28-2022 09:13:16 AM.webm](https://user-images.githubusercontent.com/15949431/204299552-0f44f986-1702-4d92-9b2f-adea9b332fff.webm)

sgkoishi:
From the video, it seems that the player starts with 12544 times and ended up with 12545, which probably indicates that the client is maintaining their local count and send to the server and overwrite it. Does the server side 0-1 count ever reach the client?

sgkoishi:
> Something still isn't correct -- the same issue as above persists, where in death the counter holds the expected value, but once the player respawns, their local player's death count is respected. I've attached a video to show exactly what I'm talking about.
> 
>     * Show that before login that local player has x deaths
> 
>     * Login to new user
> 
>     * Show new user has 0 deaths
> 
>     * Die
> 
>     * Before respawn, show user has 1 death
> 
>     * After respawn, show user does not still have 1 death as expected.

The player initiates the Spawn process, and sends the local count to the server; to overwrite we need to let the server send Spawn again (double spawn!). We could initiate an extra server Spawn once per (after) login, or once after death. This might look awkward but probably the only way 🤔

sors89:
this might be a dumb question to ask but why don't create local variables to track player deaths instead of relying on vanilla deaths counter?

ACaiCat:
Done in #3130

PR #2778, opened 2022-10-31
Built-in Nougat support

sowelipililimute:
This PR allows using NuGet to simplify the management of plugins and their dependencies.

grafik

### DIFF
--- a/TShockLauncher/.gitignore
+++ b/TShockLauncher/.gitignore
@@ -1,0 +1,1 @@
+/packages
--- a/TShockLauncher/TShockLauncher.csproj
+++ b/TShockLauncher/TShockLauncher.csproj
@@ -14,6 +14,7 @@
 	<ItemGroup>
 		<ProjectReference Include="..\TerrariaServerAPI\TerrariaServerAPI\TerrariaServerAPI.csproj" ExcludeFromSingleFile="true" />
 		<ProjectReference Include="..\TShockAPI\TShockAPI.csproj" ExcludeFromSingleFile="true" ReferenceOutputAssembly="false" /> <!-- allow api to rebuilt with this project, so ServerPlugins are refreshed -->
+		<ProjectReference Include="..\TShockPluginManager\TShockPluginManager.csproj"/>
 		<Reference Include="HttpServer" ExcludeFromSingleFile="true">
 			<HintPath>..\prebuilts\HttpServer.dll</HintPath>
 		</Reference>
--- a/TShockPluginManager/Nuget.cs
+++ b/TShockPluginManager/Nuget.cs
@@ -1,0 +1,328 @@
+/*
+TShock, a server mod for Terraria
+Copyright (C) 2022 Janet Blackquill
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+using System.Reflection;
+using System.Runtime.InteropServices;
+using NuGet.Common;
+using NuGet.Configuration;
+using NuGet.Frameworks;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using NuGet.Packaging.Signing;
+using NuGet.Protocol.Core.Types;
+using NuGet.Resolver;
+using NuGet.Versioning;
+
+namespace TShockPluginManager
+{
+
+	public class Nugetter
+	{
+		// this object can figure out the right framework folders to use from a set of packages
+		private FrameworkReducer FrameworkReducer;
+		// the package framework we want to install
+		private NuGetFramework NuGetFramework;
+		// nuget settings
+		private ISettings Settings;
+		// this is responsible for bookkeeping the folders that nuget touches
+		private NuGetPathContext PathContext;
+		// this is responsible for managing the package sources
+		private PackageSourceProvider PackageSourceProvider;
+		// this is responsible for managing the repositories of packages from all of the package sources
+		private SourceRepositoryProvider SourceRepositoryProvider;
+		// this can tell us the paths of local packages
+		private PackagePathResolver PackagePathResolver;
+		// this is possible for bookkeeping the extraction state of packages
+		private PackageExtractionContext PackageExtractionContext;
+
+		public Nugetter()
+		{
+			FrameworkReducer = new FrameworkReducer();
+			NuGetFramework = NuGetFramework.ParseFolder("net6.0");
+			Settings = NuGet.Configuration.Settings.LoadDefaultSettings(root: null);
+			PathContext = NuGetPathContext.Create(Settings);
+			PackageSourceProvider = new PackageSourceProvider(Settings);
+			SourceRepositoryProvider = new SourceRepositoryProvider(PackageSourceProvider, Repository.Provider.GetCoreV3());
+			PackagePathResolver = new PackagePathResolver(Path.GetFullPath("packages"));
+			PackageExtractionContext = new PackageExtractionContext(
+				PackageSaveMode.Defaultv3,
+				XmlDocFileSaveMode.Skip,
+				ClientPolicyContext.GetClientPolicy(Settings, NullLogger.Instance),
+				NullLogger.Instance);
+		}
+
+		async Task GetPackageDependencies(
+			PackageIdentity package,
+			NuGetFramework framework,
+			SourceCacheContext cacheContext,
+			ILogger logger,
+			IEnumerable<SourceRepository> repositories,
+			ISet<SourcePackageDependencyInfo> availablePackages)
+		{
+			// if we've already gotten dependencies for this package, don't
+			if (availablePackages.Contains(package)) return;
+
+			foreach (var sourceRepository in repositories)
+			{
+				// make sure the source repository can actually tell us about dependencies
+				var dependencyInfoResource = await sourceRepository.GetResourceAsync<DependencyInfoResource>();
+				// get the try and dependencies
+				// (the above function returns a nullable value, but doesn't properly indicate it as such)
+				#pragma warning disable CS8602
+				var dependencyInfo = await dependencyInfoResource?.ResolvePackage(
+					package, framework, cacheContext, logger, CancellationToken.None);
+				#pragma warning restore CS8602
+
+				// oop, we don't have the ability to get dependency info from this repository, or
+				// it wasn't found. let's try the next source repository!
+				if (dependencyInfo == null) continue;
+
+				availablePackages.Add(dependencyInfo);
+				foreach (var dependency in dependencyInfo.Dependencies)
+				{
+					// make sure we get the dependencies of the dependencies of the dependencies ... as well
+					await GetPackageDependencies(
+						new PackageIdentity(dependency.Id, dependency.VersionRange.MinVersion),
+						framework, cacheContext, logger, repositories, availablePackages);
+				}
+			}
+		}
+
+		/// <returns>all the packages representing dependencies bundled with TShock.Server</returns>
+		public async Task<IEnumerable<SourcePackageDependencyInfo>> GetAllBuiltinDependencies()
+		{
+			// this is a convenient approximation of what dependencies will be included with TShock.Server
+			// and really only needs to be updated if new third-party dependencies are added
+
+			var knownBundles = new[] {
+				new PackageIdentity("GetText.NET", NuGetVersion.Parse("1.6.6")),
+				new PackageIdentity("OTAPI.Upcoming", NuGetVersion.Parse("3.1.8-alpha")),
+				new PackageIdentity("TSAPI", NuGetVersion.Parse("5.0.0-beta")),
+				new PackageIdentity("TShock", NuGetVersion.Parse("5.0.0-beta")),
+			};
+
+			return await GetAllDependenciesFor(knownBundles);
+		}
+
+		/// <returns>all the dependencies for the provided package identities</returns>
+		public async Task<IEnumerable<SourcePackageDependencyInfo>> GetAllDependenciesFor(IEnumerable<PackageIdentity> targets)
+		{
+			using var cacheContext = new SourceCacheContext();
+
+			// get all of the possible packages in our dependency tree
+			var possiblePackages = new HashSet<SourcePackageDependencyInfo>(PackageIdentityComparer.Default);
+			foreach (var target in targets)
+			{
+				await GetPackageDependencies(
+					target,
+					NuGetFramework,
+					cacheContext,
+					NullLogger.Instance,
+					SourceRepositoryProvider.GetRepositories(),
+					possiblePackages
+				);
+			}
+
+			var resolverContext = new PackageResolverContext(
+				// select minimum possible versions
+				DependencyBehavior.Lowest,
+				// these are the packages the user wanted
+				targets.Select(x => x.Id),
+				// we don't hard-require anything
+				Enumerable.Empty<string>(),
+				// we don't have a lockfile
+				Enumerable.Empty<PackageReference>(),
+				// we don't have fancy versioning
+				Enumerable.Empty<PackageIdentity>(),
+				// these are the packages that we figured out are in the dependency tree from nuget
+				possiblePackages,
+				// all the package sources
+				SourceRepositoryProvider.GetRepositories().Select(s => s.PackageSource),
+				NullLogger.Instance
+			);
+
+			var resolver = new PackageResolver();
+			var packagesToInstall =
+				// get the resolved versioning info from the resolver
+				resolver.Resolve(resolverContext, CancellationToken.None)
+					// and use that to select the specific packages to install from the possible packages
+					.Select(p => possiblePackages.Single(x => PackageIdentityComparer.Default.Equals(x, p)));
+
+			return packagesToInstall;
+		}
+
+		/// <returns>whether or not subPath is a subpath of basePath</returns>
+		public static bool IsSubPathOf(string subPath, string basePath)
+		{
+			var rel = Path.GetRelativePath(basePath, subPath);
+			return rel != "."
+				&& rel != ".."
+				&& !rel.StartsWith("../")
+				&& !rel.StartsWith(@"..\")
+				&& !Path.IsPathRooted(rel);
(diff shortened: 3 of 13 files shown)


Arthri:
IMO, the plugin manager code should be split into a separate executable/project

sowelipililimute:
the plugin manager code now lives in its own project

TShockLauncher/.gitignore
```diff
@@ -0,0 +1 @@
+/packages
```
hakusaro, line 1: Can you explain why you added this to gitignore?

TShockLauncher/TShockLauncher.csproj
```diff
@@ -50,27 +51,22 @@
 	</ItemGroup>
 
 	<Target Name="CheckMsgfmtCallable">
-		<Exec Command="msgfmt --help > NUL" IgnoreExitCode="True" Condition=" '$(OS)' == 'Windows_NT' ">
+		<Exec Command="msgfmt --help &gt; NUL" IgnoreExitCode="True" Condition=" '$(OS)' == 'Windows_NT' ">
```
hakusaro: Can you explain why this changed?

TShockLauncher/TShockLauncher.csproj
```diff
@@ -50,27 +51,22 @@
...
 			<Output TaskParameter="ExitCode" PropertyName="MsgfmtExitCode" />
 		</Exec>
-		<Exec Command="msgfmt --help 2>/dev/null >/dev/null" IgnoreExitCode="True" Condition=" '$(OS)' != 'Windows_NT' ">
+		<Exec Command="msgfmt --help 2&gt;/dev/null &gt;/dev/null" IgnoreExitCode="True" Condition=" '$(OS)' != 'Windows_NT' ">
```
hakusaro: Can you explain why this changed? I'm not sure I see the necessity to this.

TShockLauncher/TShockLauncher.csproj
```diff
@@ -50,27 +51,22 @@
...
 	</Target>
 
 	<!-- The condition for a Target can't come from the values of another target, so instead we have to put the same condition on all three of the items inside. -->
-	<Target
-		Name="GenerateMOFiles"
-		DependsOnTargets="CheckMsgfmtCallable"
-		AfterTargets="PostBuildEvent;Publish"
-		Inputs="..\i18n\**\*.po"
-		Outputs="$(OutDir)i18n\**\*.mo">
+	<Target Name="GenerateMOFiles" DependsOnTargets="CheckMsgfmtCallable" AfterTargets="PostBuildEvent;Publish" Inputs="..\i18n\**\*.po" Outputs="$(OutDir)i18n\**\*.mo">
```
hakusaro: This change seems unrelated entirely to the other changes in this PR -- and notably, makes it harder to read.

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+using NuGet.Versioning;
+
+namespace TShockPluginManager
+{
+	public static class NugetCLI
+	{
+		static public async Task<int> Main(List<string> args)
+		{
+			RootCommand root = new RootCommand(
+				description: "Manage plugins and their requirements"
```
hakusaro: At a really high level, this is true of all the strings in this change set: I think that anything added to TShock, going forward, should be internationalized. It seems like a server operator could use most of TShock, just not the nuget-enabled TShock plugin system in their own language. But this is going to be relatively important. Can you add this support?

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+			SyncManifest manifest;
+			try
+			{
+				string txt = await File.ReadAllTextAsync("packages.json");
+				manifest = JsonSerializer.Deserialize<SyncManifest>(txt, opts)!;
+			}
+			catch (System.IO.FileNotFoundException)
+			{
+				System.Console.WriteLine("You're trying to sync, but you don't have a packages.json file.");
+				System.Console.WriteLine("Without a list of plugins to install, I can't install any.");
```
hakusaro: In general, TShock conventionally uses third person or indirect speech / tone when speaking to users. Can you adjust the language on these strings to be more cohesive with the rest of the project?

```suggestion
				System.Console.WriteLine("Unable to install plugins without a list of plugins to install. Please supply a packages.json file to continue.");
```

To be as precise as possible: we try to avoid having the software speak as if it is a human or otherwise an intelligent agent. "Is it okay to connect to the internet?" vs "May I connect to the internet?" (in particular, I think this doesn't flow very well into other languages -- but the primary hangup is that it's tone that doesn't really exist in the project elsewhere.)

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+				return;
+			}
+			catch (System.Text.Json.JsonException e)
+			{
+				System.Console.WriteLine("I had an issue reading the packages.json, because...");
```
hakusaro:
```suggestion
				System.Console.WriteLine("An unexpected error occurred while reading packages.json. See exception details:");
```

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+			var initial = Console.ForegroundColor;
+			foreach (var item in manifest.GetPackageIdentities())
+			{
+				System.Console.ForegroundColor = System.ConsoleColor.Green;
+				System.Console.Write(item.Id);
+				System.Console.ForegroundColor = System.ConsoleColor.Black;
+				System.Console.Write($" [{item.Version}]\n");
+			}
+			Console.ForegroundColor = initial;
+			System.Console.WriteLine("These are the packages you asked me to install.");
```
hakusaro:
```suggestion
			System.Console.WriteLine("These packages are pending installation.");
```

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+			System.Console.WriteLine("May I connect to the internet to figure out what I need to download?");
```
hakusaro:
```suggestion
			System.Console.WriteLine("Connect to the internet to resolve packages?");
```

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+			System.Console.WriteLine("=== Dependencies ===");
+			foreach (var item in nonUserPackages)
+				DumpOne(item, builtins);
+
+			System.Console.WriteLine("Can I download and install the given packages?");
+			System.Console.WriteLine("Make sure that you trust the plugins you're installing.");
+			System.Console.WriteLine("If you want to know which plugins need which dependencies, press E.");
+
+			bool ok = false;
+			do
```
hakusaro, line 145: I can't really speak to other people who wrote C# more than I do, but do people actually use do-while loops in practice? I really don't like them.

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+					new NuGetVersionConverter()
+				}
+			};
+			SyncManifest manifest;
+			try
+			{
+				string txt = await File.ReadAllTextAsync("packages.json");
+				manifest = JsonSerializer.Deserialize<SyncManifest>(txt, opts)!;
+			}
+			catch (System.IO.FileNotFoundException)
```
hakusaro, line 82: We don't really have conventions around using directives in TShock, but I personally prefer that you throw something like `using System.IO;` in and then not fully-qualify the namespace throughout the project. The issue is that again, it just seems like a lot of work and it's extra verbosity that isn't needed, but that's just me.

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+						{
+							DumpGraph(pkg, packagesToInstall, builtins, 0);
+						}
+						System.Console.WriteLine("Can I download and install the given packages?");
+						System.Console.WriteLine("If you'd like to see which plugins need which dependencies again, press E.");
+						break;
+				}
+			} while (!ok);
+
+			await nugetter.DownloadAndInstall(userRequests);
```
hakusaro, line 167: I think I need to quit C# and TShock if it means that I have to deal with async/await in this language more.

TShockPluginManager/NugetCLI.cs
```diff
@@ -0,0 +1,207 @@
...
+			var initial = Console.ForegroundColor;
+
+			var indent = new String('\t', level);
+			System.Console.Write(indent);
+			if (level != 0)
+			{
+				Console.ForegroundColor = ConsoleColor.Black;
+				System.Console.Write("needs ");
+			}
+			Console.ForegroundColor = ConsoleColor.Green;
```
hakusaro: Can you provide a sample in a comment (e.g., ASCII in comment) that shows what this is supposed to look like?

docs/changelog.md
```diff
@@ -75,6 +75,7 @@ Use past tense when adding new entries; sign your name off when you add or chang
 
 * Add ability for items given to players to be inserted directly into their inventory instead of spawned as an item drop (@pontaoski)
 * Added support of `-lang` and `-language` flags for our i18n system. (@KawaiiYuyu)
+* Add built-in package management capabilities for plugins
```
hakusaro: Can you move this to the new upcoming changes section?
```suggestion
* Added built-in package management capabilities for plugins. (@pontaoski)
```

docs/packages.md
```diff
@@ -0,0 +1,26 @@
...
+```
+./TShock.Server plugins
+```
+
+Under Windows (cmd.exe):
+```
+TShock.Server plugins
+```
+
+The documentation for the commands is included in the help functionality.
```
hakusaro, line 17: We're trying to provide as much functionality / documentation in docs files as possible, in order to make it easier to explain to people how things work. Can you provide command syntax, usages, and examples here, instead?

docs/packages.md
```diff
@@ -0,0 +1,26 @@
...
+This file primarily exists to document the `packages.json`.
```
hakusaro, line 19: Can you explain this in more layman terms for people who have no idea what this is? What does it compare to in other languages?

More broadly, when are plugins loaded from this system? How manual is this? How does it update things?

TShockPluginManager/Nuget.cs
```diff
@@ -0,0 +1,269 @@
...
+				}
+				else
+				{
+					pkgPath = null;
+					// die somehow
+				}
+			}
+			return (pkgPath, packageReader);
+		}
+		public async Task<IEnumerable<SourcePackageDependencyInfo>> GetPackagesToInstallFor(PackageIdentity[] userRequest)
```
hakusaro, line 242: I think the preferred style is a blank line before/after all methods? Can you add these?

TShockPluginManager/Nuget.cs
```diff
@@ -0,0 +1,269 @@
...
+			libItems = libItems.Where(x => x.TargetFramework.Equals(libnearest));
+
+			var frameworkItems = packageReader.GetFrameworkItems();
+			var fwnearest = FrameworkReducer.GetNearest(NuGetFramework, frameworkItems.Select(x => x.TargetFramework));
+			frameworkItems = frameworkItems.Where(x => x.TargetFramework.Equals(fwnearest));
+
+			return libItems.Concat(frameworkItems);
+		}
+
+		public async Task<(string, PackageReaderBase)> GetOrDownloadPackage(SourcePackageDependencyInfo pkg)
```
hakusaro, line 195: In-general, we try to have XML documentation for all public methods and members. Can you add this?

TShockPluginManager/Nuget.cs
```diff
@@ -0,0 +1,269 @@
...
+		public static bool IsSubPathOf(string subPath, string basePath) {
+			var rel = Path.GetRelativePath(basePath, subPath);
+			return rel != "."
+				&& rel != ".."
+				&& !rel.StartsWith("../")
+				&& !rel.StartsWith(@"..\")
+				&& !Path.IsPathRooted(rel);
+		}
+
+		public IEnumerable<FrameworkSpecificGroup> ItemsWeCareAbout(PackageReaderBase packageReader)
```
hakusaro: Why do we care about them? Can you document this method more?

TShockPluginManager/Nuget.cs
```diff
@@ -0,0 +1,269 @@
...
+		private ISettings Settings;
+		private NuGetPathContext PathContext;
+		private PackageSourceProvider PackageSourceProvider;
+		private SourceRepositoryProvider SourceRepositoryProvider;
+		private PackagePathResolver PackagePathResolver;
+		private PackageExtractionContext PackageExtractionContext;
+
+		public Nugetter()
+		{
+			FrameworkReducer  = new FrameworkReducer();
```
hakusaro: Can you consistently add only one space between things in this block of code?

hakusaro (changes requested):
Thank you for your work on this! I'm sure it will really really help in terms of getting new plugins installed for server operators in the future.

I have a few review comments. If I were to summarize my main qualms with this it's the following:

1. Internally inconsistent style. Even within only additions and only new files in this project, the code style is internally inconsistent, in terms of sheer syntactical formatting. In particular, inconsistent spacing and indentation between methods, odd spacing between individual characters, and other syntactic choices make this difficult to review.
2. Lack of support for i18n -- we just worked really hard on internationalizing TShock. There are a few edge cases, namely in TSAPI (which doesn't currently use TShock i18n stuff), but broadly, I'm happy with what we've done so far. That being said, it necessarily implies that new code should support the i18n system. If not, we're just adding more debt to pay off in the future.
3. General maintainability / future documentation needs to be improved, imho. The biggest thing to remember with TShock is that this project has been around for a long, long time. We have had maintainers come and go and entire systems become maintained by us that we have no idea how they work. For example, the entire database engine / system was written by a charismatic New Zealander who hasn't been around in...a few years. We've had to put up with this, upgrade it, maintain it, etc. for a long time. Systems like this necessarily enter the realm of "nice to have", and therefore, the maintainability requirements go up significantly. If SSC doesn't work, that's a core feature we _have_ to fix. If this doesn't work and nobody knows how to fix it, the chances of someone fixing it go down with the complexity and documentation overload. We used to have a really fancy automatic update system that would download and install and reboot TShock with updates installed. We removed that system because it was a lot of overhead/complexity to maintain and it wasn't very well-understood. This same fate will befall this system unless the code is rock solid, easily understood, and maintainable.

To elaborate more on point 3, let me recall a couple variable declaration blocks:

```csharp
		private FrameworkReducer FrameworkReducer;
		private NuGetFramework NuGetFramework;
		private ISettings Settings;
		private NuGetPathContext PathContext;
		private PackageSourceProvider PackageSourceProvider;
		private SourceRepositoryProvider SourceRepositoryProvider;
		private PackagePathResolver PackagePathResolver;
		private PackageExtractionContext PackageExtractionContext;
```

and

```csharp
			var nugetter = new Nugetter();
			var userRequests = manifest.GetPackageIdentities();
			var packagesToInstall = await nugetter.GetPackagesToInstallFor(manifest.GetPackageIdentities());
			var builtins = await nugetter.GetAllBuiltinDependencies();
			var userRoots = packagesToInstall.Where(x => userRequests.Any(y => x.Id == y.Id));
			var nonUserPackages = packagesToInstall.Where(x => !userRequests.Any(y => x.Id == y.Id));
```

In each of these instances, we "setup the world" ala COBOL or another language that supports this pattern. Then, the usages of the declared items happen significantly later and aren't super well-understood just by looking at them. I know broadly how package managers work, so I'm at an advantage here. I know what some of these words mean, and I can guess what the outputs are. But the overall control flow is opaque to a newcomer. `userRoots` _deserves_ a line or two about what a root element is and why it's important and what it's used for.

I would even support a broad architecture document that spells out, in plain English, what the high level process that this system uses is.

----

The bar for a lot of code in TShock is, unfortunately, variable. For code that isn't necessarily "core" code like what already exists, we have a relatively high bar: we want new stuff to be relatively well understood and well-patterned. This means things like documenting all public members, and even documenting private members when necessary. For existing code, like what you see in `Commands.cs`, there's a much lower barrier to entry, because we quite frankly know that we're dealing with an unwieldy beast (Terraria). The issue here is that this is entirely greenfield, and thus, there's a higher standard / barrier to entry to direct merge. I don't want to see this merged and abandoned, or merged, forgotten, and then never used. I don't want us to have to rip it all out in 5 years because it's too cumbersome to update for some reason.

On behalf of the other maintainers, I think I speak for all of them when I say that we want new code to be as clear, obvious, simple, maintainable, and consistent as possible.

Also, as an unrelated note: If you're familiar enough with git, we really prefer commits to be sensible too. For code changes, it becomes very difficult for us to find things if we have to look at the history if the commits are total chaos. For documentation and CI stuff that's really not important, but when using `git blame` commit messages ideally should have paragraphs explaining what changed and why and why it's beneficial, etc. I know that again, this is mostly greenfield, but I want to put-it-forward that it is something we care about. We just don't typically enforce these style items on some reviews (e.g., PRs that are clearly one-shots from people who can't use git), but if you can grace us with good git conventions, we'd really appreciate it too.

TShockPluginManager/Nuget.cs
```diff
@@ -0,0 +1,269 @@
...
+			}
+		}
+		public async Task<IEnumerable<SourcePackageDependencyInfo>> GetAllBuiltinDependencies()
+		{
+			var knownBundles = new[] {
+				new PackageIdentity("GetText.NET", NuGetVersion.Parse("1.6.6")),
+				new PackageIdentity("OTAPI.Upcoming", NuGetVersion.Parse("3.1.8-alpha")),
+				new PackageIdentity("TSAPI", NuGetVersion.Parse("5.0.0-beta")),
+				new PackageIdentity("TShock", NuGetVersion.Parse("5.0.0-beta")),
+			};
```
QuiCM, line 117: I'd like to avoid hard-coding versions if possible. Otherwise, when we update OTAPI/TSAPI/TShock we'll also have to keep version strings in sync here too

TShockLauncher/TShockLauncher.csproj
sowelipililimute: dotnet tooling automatically poked it and i didn't notice

TShockPluginManager/NugetCLI.cs
bartico6: > but do people actually use do-while loops in practice? I really don't like them.  

Yes we do, for any task that requires you to do it once and then repeat it as long as a condition holds. The alternative would be to either duplicate code, or do a regular while loop with some `while(first_run || actual_condition)` gimmick.

hakusaro:
@pontaoski if you're ready to have this reviewed can you de-conflict this branch and request re-reviews from me et al?

hakusaro:
@pontaoski this will get re-reviewed, probably in 5.2 with the other PRs that are large-ish and unrelated to Terraria patch. Sorry, re-logic got us surprised again.

hakusaro:
@pontaoski can you rebase?

sowelipililimute:
rebased

PR #2780, opened 2022-10-31
Hide command args when not DoLog

### LINKED ISSUE #2779: Don't Log commands that don't log if they can't be executed
There was a issue before where if a player executed a command that had the 'DoLog' enabled but doesn't have permission to execute it, it will still show the message in console saying: 'Player tried to execute: /command'
I feel like this change should be added because if in some case where someone tried to execute /login or /password without knowing they didn't have permission will just show the password in console. its very easy to add just add cmd.dolog (sorry idk how to do pull requests) 

image

sgkoishi:
resolves #2779


punchready:
Might as well add the same thing to the successful execution imo

sgkoishi:
Sounds weird since GitHub close it after I force pushed, while showing merged
image

PR #2791, opened 2022-11-02
Use same logic for all GetByNameOrId

sgkoishi:
NPC, Item, Prefix, Buff

In the previous implementation,
* Item accepts both English and localized names
* NPC also accepts both but checking `FullName` and `TypeName` while they are the same
  * because we just called `SetDefaults`
* Prefix accepts both, but can never exact-match the localized name
  * because it uses `AffixName` which returns `Prefix ItemName` or `ItemName (PrefixWithParentheses)`
* Buff only accepts localized name, no English names.

This PR unify their behaviours:
* Exact match first, either English or Localized name
* Fuzzy match
  * If only one item `StartsWith`, return it (same behaviour as previous)
  * If more than one item `StartsWith`, show both `StartsWith` and `Contains` (it's already no match or multiple match error)
  * If no `StartsWith`, fallback to `Contains`


hakusaro (changes requested):
This pull-request has great documentation in the description. Can you add it to the changelog so that other people know what changed?

hakusaro:
This is really good but I'm going to wait until 5.2 because this probably requires at least one other person besides me to review.

hakusaro (approved):

PR #2817, opened 2022-11-08
Added two new config options, relating to the TShockAPI DB.

RenderBr:
In this pull request, I have implemented two new config options:

    "isStorageTypeSeperateFromTShockDBType": true/false
    "TShockDBType": "mysql"/"sqlite"
    
 Essentially, what this does is allow the actual TShock DB (the one that stores things like Users, Bans, etc.) to be stored in a separate the way than the rest of the plugins.

What is the use case of this? If someone wanted User groups to be global across multiple servers but wanted local plugin storage. This was a pretty simple implementation and the server seems to be running fine with no problems!

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -5254,6 +5254,7 @@ namespace TShockAPI
 					new PaginationTools.Settings
 					{
 						HeaderFormat = GetString("Commands ({{0}}/{{1}}):"),
+						HeaderTextColor = new Color(117, 188, 250),
 						FooterFormat = GetString("Type {0}help {{0}} for more.", Specifier)
 					});
 			}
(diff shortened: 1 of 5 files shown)


hakusaro:
@RenderBr I'm kinda ambivalent on this, and I'd like to see more discussion before we merge it. At minimum, I'm not excited by this, because TShock's database was not intended to be a plugin database. Just because plugins use it as a database doesn't mean that's the intent. We have changes that are upcoming that may break this too.

I'm happy to have a discussion about this but it seems like it's really not the direction we've intended to go.

csmir:
Out of curiosity. What *is* the intended way to set up plugin DB's in TShock, if the given tools are not intended for it?

RenderBr:
> @RenderBr I'm kinda ambivalent on this, and I'd like to see more discussion before we merge it. At minimum, I'm not excited by this, because TShock's database was not intended to be a plugin database. Just because plugins use it as a database doesn't mean that's the intent. We have changes that are upcoming that may break this too.
> 
> I'm happy to have a discussion about this but it seems like it's really not the direction we've intended to go.

tbh, do as you want. i thought something like this could convenience other users who would like to use mysql for their actual DB and seperate sqlite databases for their plugins, for the sake of multi-server support. Yes, it is more so the plugins (and the owners) fault for relying on the actual TShock DB however that is the reality. Either way, thanks for getting back to me!

hakusaro:
@QuiCM do you have any thoughts on this?

PR #2832, opened 2022-11-11
Add WorldTileProvider to tshock config

SignatureBeef:
This allows the tshock tile providers to be switched via the config instead of cli args via tsapi, for environments where the command line args cannot be changed.

### DIFF
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -384,7 +384,7 @@ namespace TShockAPI
 					Geo = new GeoIPCountry(geoippath);
 
 				// check if a custom tile provider is to be used
-				switch(Config.Settings.WorldTileProvider)
+				switch(Config.Settings.WorldTileProvider?.ToLower())
 				{
 					case "heaptile":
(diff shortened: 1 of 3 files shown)


TShockAPI/TShock.cs
```diff
@@ -383,6 +383,19 @@ public override void Initialize()
 				if (Config.Settings.EnableGeoIP && File.Exists(geoippath))
 					Geo = new GeoIPCountry(geoippath);
 
+				// check if a custom tile provider is to be used
+				switch(Config.Settings.WorldTileProvider)
```
csmir: Is it worth considering to call `.ToLower()` here? This gives the administrator some room for error in capitalization.

sgkoishi: Or maybe create an enum for it and let Newtonsoft.Json handle the input string?

SignatureBeef: both are valid, ill roll with `.ToLower()` for now since it wont need extra types to be defined.

sgkoishi (dismissed):

csmir (dismissed):
Lgtm

hakusaro (dismissed):

PR #2833, opened 2022-11-12
Fixed Super Sponge unable to absorb shimmer

sgkoishi:
Fixes super sponge unable to handle shimmer, reported by [Asanay](https://ptb.discord.com/channels/479657350043664384/479657350043664386/1040757045059670026).

image
image (holding sponge, not picksaw)

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1736,8 +1736,7 @@ namespace TShockAPI
 				{
 					bucket = 3;
 				}
-				else if (selectedItemType == ItemID.BottomlessBucket ||
-					selectedItemType == ItemID.SuperAbsorbantSponge)
+				else if (selectedItemType == ItemID.BottomlessBucket)
 				{
 					bucket = 4;
 				}
@@ -1762,6 +1761,10 @@ namespace TShockAPI
 				{
 					bucket = 9;
 				}
+				else if (selectedItemType == ItemID.SuperAbsorbantSponge)
+				{
+					bucket = 10;
+				}
 
 				if (!wasThereABombNearby && type == LiquidType.Lava && !(bucket == 2 || bucket == 0 || bucket == 5 || bucket == 6 || bucket == 9))
 				{
@@ -1783,7 +1786,7 @@ namespace TShockAPI
 					return;
 				}
 
-				if (!wasThereABombNearby && type == LiquidType.Water && !(bucket == 1 || bucket == 0 || bucket == 4 || bucket == 9))
+				if (!wasThereABombNearby && type == LiquidType.Water && !(bucket == 1 || bucket == 0 || bucket == 4 || bucket == 9 || bucket == 10))
 				{
 					TShock.Log.ConsoleDebug(GetString("Bouncer / OnLiquidSet rejected bucket check 2 from {0}", args.Player.Name));
 					args.Player.SendErrorMessage(GetString("You do not have permission to perform this action."));
@@ -1823,7 +1826,7 @@ namespace TShockAPI
 					return;
 				}
 
-				if (!wasThereABombNearby && type == LiquidType.Shimmer && !(bucket == 8 || bucket == 9))
+				if (!wasThereABombNearby && type == LiquidType.Shimmer && !(bucket == 8 || bucket == 9 || bucket == 10))
 				{
 					TShock.Log.ConsoleDebug(GetString("Bouncer / OnLiquidSet rejected bucket check 6 from {0}", args.Player.Name));
 					args.Player.SendErrorMessage(GetString("You do not have permission to perform this action."));
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1716,7 +1716,7 @@ namespace TShockAPI
 
 			// Liquid anti-cheat
 			// Arguably the banned buckets bit should be in the item bans system
-			if (amount != 0)
+			if (amount != 0 && !wasThereABombNearby)
 			{
 				int selectedItemType = args.Player.TPlayer.inventory[args.Player.TPlayer.selectedItem].type;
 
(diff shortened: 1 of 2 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -1704,132 +1704,72 @@ internal void OnLiquidSet(object sender, GetDataHandlers.LiquidSetEventArgs args
...
 					args.Player.SendErrorMessage(GetString("You do not have permission to perform this action."));
-					args.Player.Disable(GetString("Using banned honey bucket without permissions"), DisableFlags.WriteToLogAndConsole);
+					args.Player.Disable(reason, DisableFlags.WriteToLogAndConsole);
 					args.Player.SendTileSquareCentered(tileX, tileY, 1);
 					args.Handled = true;
-					return;
 				}
 
-				if (!wasThereABombNearby && type == LiquidType.Shimmer && !(bucket == 8 || bucket == 9))
+				if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(selectedItemType), args.Player))
```
sgkoishi, line 1732: I'm not sure if the banned item check should be here, I keep it because it used to be here and the `IsBeingDisabled` check does not check item-ban disabled
https://github.com/Pryaxis/TShock/blob/5d84192624a8780bd9f35283d0f0bc1281cdb206/TShockAPI/Bouncer.cs#L1656

sgkoishi:
There is also a behaviour change here:
Previously ban lava (or other liquid) bucket => ban lava bucket, bottomless lava bucket, lava sponge, ultra sponge (if absorbing lava)
Now ban lava bucket just means lava bucket.

hakusaro:
> Now ban lava bucket just means lava bucket.

Why?

This PR is genuinely daunting to review because the implication is that the _intent_ is to "Fixed Super Sponge unable to absorb shimmer" but it also "Now ban lava bucket just means lava bucket." and in-so-doing reorders all of the code related to this and makes it hard to review. At minimum can you separate changes like this into multiple commits?

For example:

Commit A (contents) could be the diff just for absorbing shimmer
Commit B could change the behavior for the lava bucket
Commit C could reorder everything in a different way

I've looked at this PR and other PRs in TShock multiple times but it's not easy to parse or understand the change deltas with very small descriptions of changes.

sgkoishi:
Now this adds back to the original behavior, banning the lava bucket will disable all lava-related actions, and so do water/honey/shimmer (via bottomless shimmer bucket).

sgkoishi:
The commit 49921cb seems to be confusing - view it with [`"patience diff" algorithm`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---patience) (`git diff --patience ac7ee7e..49921cb`) would be more clear:
```diff
diff --git a/TShockAPI/Bouncer.cs b/TShockAPI/Bouncer.cs
index cfa10b50..4b1ff297 100644
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -1729,54 +1729,70 @@ namespace TShockAPI
 					args.Handled = true;
 				}
 
+				if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(selectedItemType), args.Player))
+				{
+					Reject(GetString("Using banned {0} to manipulate liquid", Lang.GetItemNameValue(selectedItemType)));
+					return;
+				}
+
+				switch (type)
+				{
+					case LiquidType.Water:
+						if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(ItemID.WaterBucket), args.Player))
+						{
+							Reject(GetString("Using banned water bucket without permissions"));
+							return;
+						}
+						break;
+					case LiquidType.Lava:
+						if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(ItemID.LavaBucket), args.Player))
+						{
+							Reject(GetString("Using banned lava bucket without permissions"));
+							return;
+						}
+						break;
+					case LiquidType.Honey:
+						if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(ItemID.HoneyBucket), args.Player))
+						{
+							Reject(GetString("Using banned honey bucket without permissions"));
+							return;
+						}
+						break;
+					case LiquidType.Shimmer:
+						if (TShock.ItemBans.DataModel.ItemIsBanned(EnglishLanguage.GetItemNameById(ItemID.BottomlessShimmerBucket), args.Player))
+						{
+							Reject(GetString("Using banned shimmering water bucket without permissions"));
+							return;
+						}
+						break;
+					default:
+						Reject(GetString("Manipulating unknown liquid type"));
+						return;
+				}
+
 				if (type == LiquidType.Lava && !(selectedItemType == ItemID.LavaBucket || selectedItemType == ItemID.EmptyBucket || selectedItemType == ItemID.LavaAbsorbantSponge || selectedItemType == ItemID.BottomlessLavaBucket || selectedItemType == ItemID.UltraAbsorbantSponge))
 				{
 					Reject(GetString("Spreading lava without holding a lava bucket"));
 					return;
 				}
 
-				if (type == LiquidType.Lava && TShock.ItemBans.DataModel.ItemIsBanned("Lava Bucket", args.Player))
-				{
-					Reject(GetString("Using banned lava bucket without permissions"));
-					return;
-				}
-
 				if (type == LiquidType.Water && !(selectedItemType == ItemID.WaterBucket || selectedItemType == ItemID.EmptyBucket || selectedItemType == ItemID.BottomlessBucket || selectedItemType == ItemID.UltraAbsorbantSponge || selectedItemType == ItemID.SuperAbsorbantSponge))
 				{
 					Reject(GetString("Spreading water without holding a water bucket"));
 					return;
 				}
 
-				if (type == LiquidType.Water && TShock.ItemBans.DataModel.ItemIsBanned("Water Bucket", args.Player))
-				{
-					Reject(GetString("Using banned water bucket without permissions"));
-					return;
-				}
-
 				if (type == LiquidType.Honey && !(selectedItemType == ItemID.HoneyBucket || selectedItemType == ItemID.EmptyBucket || selectedItemType == ItemID.BottomlessHoneyBucket || selectedItemType == ItemID.HoneyAbsorbantSponge || selectedItemType == ItemID.UltraAbsorbantSponge))
 				{
 					Reject(GetString("Spreading honey without holding a honey bucket"));
 					return;
 				}
 
-				if (type == LiquidType.Honey && TShock.ItemBans.DataModel.ItemIsBanned("Honey Bucket", args.Player))
-				{
-					Reject(GetString("Using banned honey bucket without permissions"));
-					return;
-				}
-
 				if (type == LiquidType.Shimmer && !(selectedItemType == ItemID.BottomlessShimmerBucket || selectedItemType == ItemID.UltraAbsorbantSponge || selectedItemType == ItemID.SuperAbsorbantSponge))
 				{
 					Reject(GetString("Spreading shimmer without holding a shimmer bucket"));
 					return;
 				}
-
-				if (type == LiquidType.Shimmer &&
-					TShock.ItemBans.DataModel.ItemIsBanned("Bottomless Shimmer Bucket", args.Player))
-				{
-					Reject(GetString("Using banned bottomless shimmer bucket without permissions"));
-					return;
-				}
 			}
 
 			if (!args.Player.HasBuildPermission(tileX, tileY))
```

TShockAPI/Bouncer.cs
```diff
@@ -1770,28 +1770,54 @@ void Reject(string reason)
 						return;
 				}
 
-				if (type == LiquidType.Lava && !(selectedItemType == ItemID.LavaBucket || selectedItemType == ItemID.EmptyBucket || selectedItemType == ItemID.LavaAbsorbantSponge || selectedItemType == ItemID.BottomlessLavaBucket || selectedItemType == ItemID.UltraAbsorbantSponge))
+				switch (selectedItemType)
```
drunderscore, line 1773: I find this entire commit to be weird... why group the check of liquids based on item instead? It just feels so backwards when looking at the code.

TShockAPI/Bouncer.cs
```diff
@@ -1736,8 +1736,7 @@ internal void OnLiquidSet(object sender, GetDataHandlers.LiquidSetEventArgs args
 				{
 					bucket = 3;
 				}
-				else if (selectedItemType == ItemID.BottomlessBucket ||
-					selectedItemType == ItemID.SuperAbsorbantSponge)
+				else if (selectedItemType == ItemID.BottomlessBucket)
```
drunderscore: What is even the point of this first commit, if not to just remove this entirely the next commit?

TShockAPI/Bouncer.cs
```diff
@@ -1718,55 +1718,9 @@ internal void OnLiquidSet(object sender, GetDataHandlers.LiquidSetEventArgs args
 			// Arguably the banned buckets bit should be in the item bans system
 			if (amount != 0)
 			{
-				int bucket = -1;
 				int selectedItemType = args.Player.TPlayer.inventory[args.Player.TPlayer.selectedItem].type;
```
drunderscore, line 1721: And what's the point of this commit, if not to make all of this code a switch case later?

drunderscore (changes requested):
I'm not sure if this is what @hakusaro was exactly looking for when mentioning "separate commits", but I find these commits to be super weird:
- The first commit changes the magic numbers to include allow the Super Sponge
- ...but then the second commit removes those magic numbers completely
- ...but then the sixth commit totally removes all of that and replaces it with a `switch` statement on the items?

I do not recall how this PR previously looked, and that's certainly my own fault, but it's hard to imagine this is better?

sgkoishi:
The first commit resolves the issue, `SuperAbsorbantSponge` and `BottomlessBucket` behave differently and we need to split them.

However, it turns out that the `bucket` number is meaningless because basically, each item maps to a bucket, e.g. `bucket == 0` is abbr of `selectedItemType == ItemID.EmptyBucket`. With `ItemID` it is more readable to have `ItemID.EmptyBucket` than `0`. This is the second commit.

After the second commit, the lines are long and there are still many duplications, e.g. we have four `EmptyBucket` and four `UltraAbsorbantSponge` checks, group them by item type helps readability: if they share common behaviour, they use the same code instead of duplicating `selectedItemType == ItemID.EmptyBucket || selectedItemType == ItemID.UltraAbsorbantSponge` or `bucket == X` four times.
https://github.com/Pryaxis/TShock/blob/536c4d25459a61971af07d03453e6d46ce7f4b41/TShockAPI/Bouncer.cs#L1815-L1817

The original commit 3e61e68 is exactly the same but in one commit and without this, (restrict all lava actions if lava bucket banned, etc. to keep the same behaviour as old implementation)
https://github.com/Pryaxis/TShock/blob/536c4d25459a61971af07d03453e6d46ce7f4b41/TShockAPI/Bouncer.cs#L1738-L1771

TShockAPI/Bouncer.cs
sgkoishi: The first commit resolves the issue, it is standalone and can be cherry-picked.

TShockAPI/Bouncer.cs
```diff
@@ -1729,27 +1729,56 @@ void Reject(string reason)
...
+					Reject(GetString("Using banned {0} to manipulate liquid", Lang.GetItemNameValue(selectedItemType)));
 					return;
 				}
 
-				if (type == LiquidType.Lava && TShock.ItemBans.DataModel.ItemIsBanned("Lava Bucket", args.Player))
+				switch (type)
 				{
-					Reject(GetString("Using banned lava bucket without permissions"));
-					return;
+					case LiquidType.Water:
```
drunderscore, line 1740: It is a bit weird the basic bucket type being banned means the liquid is entirely banned, but that _is_ what the original code did, so that's alright I think.

drunderscore (approved):
I've tested that these changes fix the titled issue.

In the end, the commits will be squashed and merged into `general-devel`, so I'm alright with them staying how they are, as I was able to review the final code alright.

The `switch`ing on the selected item to check against the liquid type actually seems fine. I was a bit over-the-top with my previous dismissal of it, but the code is clean with it.

PR #2834, opened 2022-11-12
PVP with no team

### LINKED ISSUE #2617: Force players to not be able to switch to a team.
Add a option in the config menu to force players onto 'No Team' for 'Force PVP' usage.

ATFGK:
resolved #2617 

An additional option `pvpwithnoteam` is added at `PvPMode` to enable PVP with no team


drunderscore:
It feels a bit weird to specify team properties in a configuration option named `PvPMode`... how about creating a new configuration option (such as `TeamMode`), with similar options to `PvPMode`, where:
- `normal`, which is the default, and does nothing
- `disabled`, which prevents team switching
- A team name, such as `red` or `blue`, which forces players onto that team

That way all the existing options of `PvPMode` can be used alongside `TeamMode` in multiple configurations.

ATFGK:
> 

It's a good idea, but it doesn't seem to work because the same team will lose its PVP

ATFGK:
> It feels a bit weird to specify team properties in a configuration option named ... how about creating a new configuration option (such as ), with similar options to , where:`PvPMode``TeamMode``PvPMode`
> 
> * `normal`, which is the default, and does nothing
> * `disabled`, which prevents team switching
> * A team name, such as  or , which forces players onto that team`red``blue`
> 
> That way all the existing options of can be used alongside in multiple configurations.`PvPMode``TeamMode`

This idea seems good, but it doesn't seem to work because the same team will lose its PVP

sgkoishi:
Would it be better to use permission to restrict team switch and pvp?

ATFGK:
> Would it be better to use permission to restrict team switch and pvp?

Maybe so.

sgkoishi:
It's usually suggested to use another branch to merge unrelated changes since this branch is connected to a PR.

hakusaro:
@ATFGK can you clean this PR up or at least point to the exact commit(s) you want merged so that we can potentially just cherry-pick the changes? Each PR tree should be an independent set of changes. This is really confusing and not likely something we can merge without significant cleanup of this branch.

ATFGK:
> @ATFGK can you clean this PR up or at least point to the exact commit(s) you want merged so that we can potentially just cherry-pick the changes? Each PR tree should be an independent set of changes. This is really confusing and not likely something we can merge without significant cleanup of this branch.

I will resend it.....

Issue #1919, opened 2020-05-26
Digging Mole Cart is not supported


hakusaro:
Video: https://gyazo.com/4aba62f7e76050464a77d1b1af6c87cd

![image](https://user-images.githubusercontent.com/532647/82855504-57131800-9ec0-11ea-9207-8ad7d7d76385.png)

> While riding it, it tries to extend the minetrack by placing tiles in front of the cart. Normally this doesn't work for blocks such as Crimson and Sand but this can be "sort of exploited" in Terraria where it replaces the first block at the bottom with a minecart track. However the mole itself will not mine through the blocks if they are "not mineable" or so to say.

> However in TShock it mines the minecart track and not the ground next to it.

14Si:
Hey I think i'm getting this issue on the latest release... It's also consuming the minecart tracks when placing is failing... I see this was merged just after. Is general-devel stable enough right now?

sgkoishi:
> Hey I think i'm getting this issue on the latest release... It's also consuming the minecart tracks when placing is failing... I see this was merged just after. Is general-devel stable enough right now?

Should be fixed #2839 for the next release.

PR #2839, opened 2022-11-20
Fix Digging Molecart and Bomb Fish blocked

sgkoishi:
Digging Molecart can PlaceTile (tracks only) and KillTile.
Bomb Fish cause TileKill even when not holding.


sgkoishi:
Resolves #1919

PR #2830, opened 2022-11-11
Update GetDataHandlers.cs

ATFGK:
Fixes #2804 

Fix `GetDataHandlers.cs` file for version 144X, and correct various errors caused by it.

PR #2827, opened 2022-11-11
Add Shadow Candle and Brain of Confusion to `PlayerAddBuffWhitelist`, improve rejection messages

drunderscore:
Fixes #2799
Fixes [issue in Discord](https://discord.com/channels/479657350043664384/479657350043664386/1040411699632672858) reported by Kojiro (which was reported to them by another individual)

PR #2845, opened 2022-11-27
Multiple merges

ATFGK:
#2840 + #2833 + #2834 + #2830 + #2827 + #2768 

Is this allowed?
A version with fewer bugs is really desirable.


sgkoishi:
Likely this will not be merged, but [the ci always build it for you](https://github.com/ATFGK/TShock/actions) if you want to _use_ it.

ATFGK:
> Likely this will not be merged, but [the ci always build it for you](https://github.com/ATFGK/TShock/actions) if you want to _use_ it.

This function is really powerful. It is good enough that "forward-looking" can be experienced in advance.╮(￣▽ ￣)╭

hakusaro:
I approved a build for this.

> Likely this will not be merged, but [the ci always build it for you](https://github.com/ATFGK/TShock/actions) if you want to use it.

this is true.

this weekend, I have work related to the update server because of heroku shutting down free stuff, so that's taking a bit of time that isn't visible here.

PR #2847, opened 2022-11-28
Allow more tile and wall conversion

sgkoishi:
Fixes unable to use Purification/Evil Powder in jungle, reported by [toxic](https://ptb.discord.com/channels/479657350043664384/479657350043664386/1046594200931667978).

### DIFF
--- a/TShockAPI/Handlers/SendTileRectHandler.cs
+++ b/TShockAPI/Handlers/SendTileRectHandler.cs
@@ -390,6 +390,9 @@ namespace TShockAPI.Handlers
 			}
 		}
 
+		private static List<bool[]> _convertibleTiles = typeof(TileID.Sets.Conversion).GetFields().Select(f => (bool[])f.GetValue(null)).ToList();
+		private static List<bool[]> _convertibleWalls = typeof(WallID.Sets.Conversion).GetFields().Select(f => (bool[])f.GetValue(null)).ToList();
+
 		/// <summary>
 		/// Updates a single tile on the server if it is a valid conversion from one tile or wall type to another (eg stone -> corrupt stone)
 		/// </summary>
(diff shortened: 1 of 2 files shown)


sgkoishi:
This actually add more allowed conversion other than jungle:
`MergesWithDirtInASpecialWay` set:
```
TileID.Grass
TileID.CorruptGrass
TileID.HallowedGrass
TileID.CrimsonGrass
TileID.GolfGrass
TileID.GolfGrassHallowed
```

`JungleGrass` set:
```
TileID.JungleGrass
TileID.CorruptJungleGrass
TileID.CrimsonJungleGrass
```

`GolfGrass` set:
```
TileID.GolfGrass
TileID.GolfGrassHallowed
```
<hr>
and walls

`Dirt` set:
```
WallID.DirtUnsafe
WallID.Dirt
```

`Snow` set:
```
WallID.SnowWallUnsafe
WallID.SnowWallEcho
```

`Ice` set:
```
WallID.IceUnsafe
WallID.IceEcho
```

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -398,36 +401,22 @@ internal void ProcessFlowerBoots(int realX, int realY, NetTile newTile)
...
-				TileID.Sets.Conversion.MossBrick[tile.type] && TileID.Sets.Conversion.MossBrick[newTile.Type]
-			)
+			foreach (var tileType in _convertibleTiles)
 			{
-				TShock.Log.ConsoleDebug(GetString($"Bouncer / SendTileRect processing a tile conversion update - [{tile.type}] -> [{newTile.Type}]"));
-				UpdateServerTileState(tile, newTile, TileDataType.Tile);
+				if (tileType[tile.type] && tileType[newTile.Type])
+				{
+					TShock.Log.ConsoleDebug(GetString($"Bouncer / SendTileRect processing a tile conversion update - [{tile.type}] -> [{newTile.Type}]"));
+					UpdateServerTileState(tile, newTile, TileDataType.Tile);
```
drunderscore: Can we `break` early if this conversion is valid? I know the original code didn't, but it makes sense to do, and I believe it remains just fine and compatible with vanilla.

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -398,36 +401,22 @@ internal void ProcessFlowerBoots(int realX, int realY, NetTile newTile)
...
-				WallID.Sets.Conversion.NewWall4[tile.wall] && WallID.Sets.Conversion.NewWall4[newTile.Wall]
-			)
+			foreach (var wallType in _convertibleWalls)
 			{
-				TShock.Log.ConsoleDebug(GetString($"Bouncer / SendTileRect processing a wall conversion update - [{tile.wall}] -> [{newTile.Wall}]"));
-				UpdateServerTileState(tile, newTile, TileDataType.Wall);
+				if (wallType[tile.wall] && wallType[newTile.Wall])
+				{
+					TShock.Log.ConsoleDebug(GetString($"Bouncer / SendTileRect processing a wall conversion update - [{tile.wall}] -> [{newTile.Wall}]"));
+					UpdateServerTileState(tile, newTile, TileDataType.Wall);
```
drunderscore, line 447: Ditto

sgkoishi:
According to the client-side projectile AI `WorldGen.Convert(int i, int j, int conversionType, int size = 4)`, the `TileID.Sets.Conversion.Moss` and `TileID.Sets.Conversion.MossBrick` is never used by any vanilla code (perhaps we should remove them?)
(`Moss` is actually used in some checks like gemtree and shimmer, but neither is used in conversion)

drunderscore (changes requested):
This does solve the issue of not being able to use Purification Powder on Crimson Jungle Grass. This implementation for sure seems to future-proof us much better!

drunderscore:
@sgkoishi Yeah, we should only allow the conversions that vanilla allows, being as strict and accurate as possible. A filter on the field name should suffice?

hakusaro (dismissed):

hakusaro:
@drunderscore @sgkoishi hi! I didn't merge this because @drunderscore has a change request on this PR, fyi.

sgkoishi:
Yes, it's still under consideration which impl is better. The proposed version fixed the problem and seems to be future-proof, but it actually allows certain impossible conversion (which those are harmless for now).

sgkoishi:
The wall conversion is actually way more complicated - perhaps create dummy tiles and run `WorldGen.Convert` would be better to get an 'accurate' convertible table.

drunderscore (approved):
Code looks good, field filtering is also fine.
Tested and it does fix the issues talked about in Discord.

hakusaro (approved):

PR #2857, opened 2022-12-02
Whitelist the Mighty Wind and Hellfire buffs

drunderscore:
This was reported in [this comment](https://github.com/Pryaxis/TShock/pull/2827#issuecomment-1328418946), and then reminded to me by another Discord server.


ATFGK:
Thank you for your repair. During the game, we also found that the "Other Beast" in the "Old One's Army event" can also cause this problem.

PR #2872, opened 2022-12-13
Fixed startup error

ATFGK:
I don't know how this error came about, but I did encounter it when using the experimental version.
VA{QF}HCMG%)5(Y9(U~4ZNB


sgkoishi:
<del>Pretty sure this is not related and should not be merged 🤔 </del> 
Just realized there's another `Start`

sgkoishi (approved):

PR #2874, opened 2022-12-15
Seed initial group database with `default` and `guest` names from config

drunderscore:
The `GroupManager` constructor ensures the structure of the `GroupList` table, and if the table is newly-created, then some initial data is seeded.

We previously used the names `default` to represent the group that accounts are defaultly registered to, and `guest` to represent players that have not logged in yet. We now instead refer to the group names specified by the config (`DefaultRegistrationGroupName` and `DefaultGuestGroupName` respectively)


hakusaro (approved):

hakusaro:
@drunderscore this PR is a draft but can it be merged?

PR #2882, opened 2022-12-21
Detect invalid TShock installations

drunderscore:
The grammar introduced in this PR is _meant_ to be nit-picked! Please suggest what sounds most natural and understandable to an end-user.

Additionally, because we are in TShockLauncher and not in TShock yet, we cannot translate strings... and an error like this would benefit greatly from such. It may be worthwhile to hardcode some translations?


hakusaro:
Thanks! You've saved @CEO-Chair's bacon!

TShockLauncher/Program.cs
```diff
@@ -29,20 +29,34 @@
 using System.Reflection;
 using TShockPluginManager;
 
+// On occasion, users have been seen extracting TShock into their client installation directory -- this is of course incorrect, and is known
+// to cause issues. Let's attempt to catch this before anything happens (specifically, before Terraria assemblies are resolved) and prevent
+// TShock from launching.
+if (Directory.Exists("Content"))
```
Arthri: I believe the presence of Terraria.exe is a stronger indication

drunderscore: Yeah agreed, let's check for that instead

sgkoishi: Or perhaps `TerrariaServer.exe`, since it also indicates legacy TShock?

drunderscore: I agree even more -- I've added checks and updated the output to inform users about updating legacy TShock installations as well.

TShockLauncher/Program.cs
```diff
@@ -29,20 +29,38 @@
 using System.Reflection;
 using TShockPluginManager;
 
+// On occasion, users have been seen extracting TShock into their client installation directory -- this is of course incorrect, and is known
+// to cause issues. Let's attempt to catch this before anything happens (specifically, before Terraria assemblies are resolved) and prevent
+// TShock from launching.
+if (File.Exists("TerrariaServer.exe"))
```
Arthri, line 35: 😅 I forgot to mention the .exe part is windows-specific. should it take into account other OSes?

drunderscore: No, the `exe` part is not windows-specific, it also exists on Linux... and I'd assume Mac

sgkoishi: At least vanilla and TShock 4 are both using Mono - and they are just regular assembly and extension doesn't matters much. This doesn't work for tModLoader, but who would do that tho?

hakusaro:
I think I would check for Content instead of a platform-specific binary, personally. But I'd check for something inside Content too?

hakusaro (approved):

PR #2888, opened 2023-01-02
Remove unnecessary range check for quick stack

### LINKED ISSUE #2885: Quick Stack range is shortened by range check
* TShock version: 5.1.3.0
* TShock build number (if known):

* Description of issue:
In the current version of Terraria (1.4.4.9 PC), and I believe since 1.4.4, the range for quick stacking to nearby chests is ~37. This is verified by both the wiki and my own personal tests.
However, on servers using TShock, the quick stack range is maximally 32 by my own measurements.

* Expected problem and fix:
[Quick stack range check](https://github.com/Pryaxis/TShock/blob/24bb002e646c88dacb1d1d864f8c171b48784435/TShockAPI/TShock.cs#L685-L689)
[IsInRange](https://github.com/Pryaxis/TShock/blob/95d157fd5d24f8997578ca9a4d67f988b6b7d7f1/TShockAPI/TSPlayer.cs#L752-L762)
The code at the first link runs when a player tries to quick stack. If the chest is not in range, it returns without allowing the player.
The function IsInRange is called without a supplied value for range, and so defaults to 32. Adding a range value of, say, 40 on line 685 should fix  this issue.

1. Go to an isolated area.
2. Place a chest down and place gray bricks in it, leaving some in your personal inventory.
3. Walk a distance away using the ruler tool.
4. While, say, 35 tiles away, open your inventory.
5. Click the "Quick Stack to Nearby Chests" button.
6. Observe that the gray bricks in your inventory do not quick stack to the chest.

None, nor would I expect any.

#### Any screenshots?

image
Fig. 1 - Showing setup of inventories.
image
Fig. 2 - At range of 35 tiles, quick stacking to nearby chests does not work.
image
Fig. 3 - Now at a range of 32.
image
Fig. 4 - At the range of 32, you can now quick stack to the chest.

bcat:
Fixes #2885.

[As described here](https://github.com/Pryaxis/TShock/issues/2885#issuecomment-1367732843), vanilla Terraria already has a range check for quick stacking ([packet 85](https://tshock.readme.io/docs/multiplayer-packet-structure#force-item-into-nearest-chest-85) from the client only lists the inventory slot to quick stack, and the server decides what chests are close enough). As such, removing TShock's check is safe, and it enables quick stacking to the same range as vanilla (~36 tiles) rather than the 32 tiles TShock is limited to with the extra range check.

Example of distance that you couldn't quick stack from before, but can now after this fix:

image

### DIFF
--- a/TShockAPI/TShock.cs
+++ b/TShockAPI/TShock.cs
@@ -676,13 +676,9 @@ namespace TShockAPI
 
 			if (args.Chest != null)
 			{
+				// After checking for protected regions, no further range checking is necessarily because the client packet only specifies the
+				// inventory slot to quick stack. The vanilla Terraria server itself determines what chests are close enough to the player.
 				if (Config.Settings.RegionProtectChests && !Regions.CanBuild((int)args.WorldPosition.X, (int)args.WorldPosition.Y, tsplr))
-				{
-					args.Handled = true;
-					return;
-				}
-
-				if (!tsplr.IsInRange(args.Chest.x, args.Chest.y))
 				{
 					args.Handled = true;
 					return;
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -106,6 +106,7 @@ Use past tense when adding new entries; sign your name off when you add or chang
 * Allowed multiple test cases to be in TShock's test suite. (@drunderscore)
 * Fixed unable to use Purification/Evil Powder in jungle. (@sgkoishi)
 * Set the `GetDataHandledEventArgs.Player` property for the `SyncTilePicking` data handler. (@drunderscore)
+* Removed unnecessary range check that artifically shortened quick stack reach. (@boddyn, #2885, @bcat)
 
 ## TShock 5.1.3
 * Added support for Terraria 1.4.4.9 via OTAPI 3.1.20. (@SignatureBeef)


drunderscore (dismissed):
Yup, no range check needed at all, this works.

It's a bit funny actually, the only two checks in this method are actually wrong -- the one you just fixed, and the [region check discussed in Discord](https://discord.com/channels/479657350043664384/711282444933791774/1071951210346721360) is implemented incorrectly.

xtpkt (dismissed):

bcat:
Rebased against HEAD to resolve conflicts (which reset the approval status, sorry!).

Arthri:
> and the server decides what chests are close enough

Where does the server decide this? I don't see range checks in Packet 85's code path

punchready:
It's right there in `Chest.PutItemInNearbyChest` 🤔
```cs
Vector2 vector = new Vector2(Main.chest[i].x * 16 + 16, Main.chest[i].y * 16 + 16);
if ((vector - position).Length() >= 600f)
{
	continue;
}
```

Arthri:
Ahh. Found it

Arthri (dismissed):

QuiCM (approved):

Arthri (approved):

PR #2915, opened 2023-02-11
Improve the logic of the HandlePlayerSlot

sowelipililimute:
Somewhat shooting in the dark here, but it seems like we shouldn't ever accept inventory slot events from players after we've told them that the server is SSC


punchready:
This is bogus.

chairclr:
This is bogus.

PR #2936, opened 2023-04-04
Rewrite SendTileRrect handling

punchready:
The old handler had a pretty significant flaw allowing attackers to basically bypass most checks, so I went to rewrite it completely.
All categories of checks have been tested successfully, but I did not perform exhaustive checks for every potential action due to lack of time, so if needed someone could do that based on the list of operations in the code.
While writing this it also became apparent again that tShock does not have a way to restrict wall placement like it does with tiles, which would matter during biome conversion handling.


QuiCM (changes requested):
Rewriting Terraria code to protect against its own exploits makes me cry, but it is what it is.

Only real change I'd like is the removal of the existing `SendTileRectHandler`, and rename this to take its place

punchready:
Yeah I wanted to avoid making the diff unreadable, which at the very least requires a different file name which would also suck. I guess I can change the class name after everyone has had a chance to check it out while the diff is still clean.

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1,739 +1,1329 @@
...
+								if (tile.FrameX < 0 || tile.FrameX > MaxFrameX || tile.FrameX % FrameXStep != 0)
+								{
+									return false;
+								}
+							}
+							if (MaxFrameY != IGNORE_FRAME)
+							{
+								if (tile.FrameY < 0 || tile.FrameY > MaxFrameY || tile.FrameY % FrameYStep != 0)
+								{
+									// this is the only tile type sent in a tile rect where the frame have a different pattern (56, 74, 92 instead of 54, 72, 90)
```
QuiCM, line 214: Is there a separate check for this particular case? Does there need to be one?

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1,739 +1,1329 @@
...
-						!args.Player.IsInRange(realX, realY))
+			private bool MatchPlacement(TSPlayer player, TileRect rect)
+			{
+				for (int x = rect.X; x < rect.Y + rect.Width; x++)
+				{
+					for (int y = rect.Y; y < rect.Y + rect.Height; y++)
 					{
-						processed[x, y] = true;
-						continue;
+						if (Main.tile[x, y].active() && !(Main.tile[x, y].type != TileID.RollingCactus && (Main.tileCut[Main.tile[x, y].type] || TileID.Sets.BreakableWhenPlacing[Main.tile[x, y].type])))
```
QuiCM: Why is rolling cactus special? A comment would help clarify why this particular set of logic is necessary

QuiCM (dismissed):

QuiCM:
Couple minor Qs, but overall I'm happy to trust that this works

TShockAPI/Handlers/SendTileRectHandler.cs
punchready: The check being referred to is right below the comment?

TShockAPI/Handlers/SendTileRectHandler.cs
punchready: This is from `TileObject.Place` but actually the client will kill those tiles manually before sending that, so not sure why i kept that in, will remove in a minute.

TShockAPI/Handlers/SendTileRectHandler.cs
QuiCM: Gotcha. Tired brain + Terraria logic does not a good combination make.






I hate tile logic

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1,739 +1,1329 @@
...
+
+				for (int x = 0; x < rect.Width; x++)
 				{
-					// Do not process already processed tiles
-					if (processed[x, y])
+					for (int y = 0; y < rect.Height; y++)
 					{
-						continue;
+						NetTile tile = rect[x, y];
+						if (Type is MatchType.Placement or MatchType.StateChange)
```
Arthri, line 194: `StateChange` seems to have a duplicate check for this in `MatchStateChange`. And Removal seems to implement it separately

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1,739 +1,1329 @@
...
 				return;
 			}
 
-			if (TShock.TileBans.TileIsBanned((short)tileType))
+			// this handler handles the entire logic of this packet
+			args.Handled = true;
+
+			// as of 1.4 this is the biggest size the client will send in any case, determined by full code analysis
+			// see default matches above and special cases below
+			if (args.Width > 4 || args.Length > 4)
```
Arthri, line 425: The client sends a 5x5 SendTileSquare for beds and bathtubs in `PlaceThing_Tiles_PlaceIt_SpinBedsAndBaths`. I think there's a 1x6 in `ToggleLampPost`, though I am unsure if this is reachable client code. `HitSwitch` implies it is

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1,739 +1,1329 @@
...
+					args.Player.SendTileRect(args.TileX, args.TileY, args.Length, args.Width);
+					return;
 				}
 			}
 
-			// Conversion: only sends a 1x1 rect; has to happen AFTER grass mowing as it would otherwise also let mowing through, but without fixing vines
-			if (rectWidth == 1 && rectLength == 1)
+			// a few special cases
+			if (
+				MatchesConversionSpread(args.Player, rect) ||
```
Arthri: This check seems to happen twice, there's another one above

Arthri (changes requested):

TShockAPI/Handlers/SendTileRectHandler.cs
Arthri: Nevermind, upon further analysis, neither are true. `TripWire` prevents the latter case from happening, and all beds and bath tubs do not trigger `PlaceThing_Tiles_PlaceIt_SpinBedsAndBaths` at the moment

Arthri (dismissed):

TShockAPI/Handlers/SendTileRectHandler.cs
punchready: Explicit rotation on placement is a legacy implementation, in 1.4 rotation is handled based on the player's direction. The other STR cases are only sent by the server, mostly from wiring toggles as you noted.

TShockAPI/Handlers/SendTileRectHandler.cs
punchready: The check in `Matches` verifies the tile data that is incoming (`TileRect[x, y]`), the checks in `MatchPlacement` etc. verify the target tiles (`Main.tile[x, y]`).

Arthri: Ah I see, I missed that

Arthri (approved):

PR #2939, opened 2023-04-28
Use same TSPlayer.Active check

sgkoishi:
Currently some use `Player.active` while some use `TSPlayer.Active`. They usually behave the same but this inconsistency causes problems when trying to do something related - e.g. hide a player.
This PR unify them to use the same `TSPlayer.Active`.


Arthri (approved):

AgaSpace:
If some plugin will assign a value to `TSPlayer.Active`, such as the same `Ghost`, then we will get a player who can ignore most anti-cheats.

sgkoishi:
> If some plugin will assign a value to `TSPlayer.Active`, such as the same `Ghost`, then we will get a player who can ignore most anti-cheats.

Players are always active without plugin. If a player is not `TSPlayer.Active` or `Player.active`, it makes sense to reject them - which is the current behaviour (and the PR is not breaking it).

PR #2940, opened 2023-05-01
Changes to PlayerData, NetItem, TSPlayer and hooks

AgaSpace:


sgkoishi:
*Plugins that use `TSPlayer.DamagePlayer` or `KillPlayer` requires recompile (source level compatible).*
*Plugins that use `TShockAPI.Hooks.PlayerHooks.OnPlayerCommand` requires change.*

punchready (changes requested):
- The `OnAccountGroupChange` hook should be invoked in `UserManager.SetUserGroup` instead, right now it only fires when a group gets changed via command
- The `PlayerCommand` hook should probably be kept and called like it was before for binary backwards compatibility
- The `UpdateSection` method's `visibility` parameter should be called something like `isLoaded` instead to not be misleading, and the documentation adjusted
- The `DamagePlayer` and `KillPlayer` reason parameters should also be put into an overload instead for binary backwards compatibility

AgaSpace:
> * The `OnAccountGroupChange` hook should be invoked in `UserManager.SetUserGroup` instead, right now it only fires when a group gets changed via command
> * The `PlayerCommand` hook should probably be kept and called like it was before for binary backwards compatibility
> * The `UpdateSection` method's `visibility` parameter should be called something like `isLoaded` instead to not be misleading, and the documentation adjusted
> * The `DamagePlayer` and `KillPlayer` reason parameters should also be put into an overload instead for binary backwards compatibility

I wrote it down as you instructed

TShockAPI/TSPlayer.cs
```diff
@@ -1815,15 +1872,35 @@ public void SendFileTextAsMessage(string file)
 		/// <param name="damage">The amount of damage the player will take.</param>
 		public virtual void DamagePlayer(int damage)
 		{
-			NetMessage.SendPlayerHurt(Index, PlayerDeathReason.LegacyDefault(), damage, (new Random()).Next(-1, 1), false, false, 0, -1, -1);
+			NetMessage.SendPlayerHurt(Index, PlayerDeathReason.LegacyDefault(),
+				damage, (new Random()).Next(-1, 1), false, false, 0, -1, -1);
```
punchready: This should call the second overload with the default reason.

TShockAPI/TSPlayer.cs
```diff
@@ -1815,15 +1872,35 @@ public void SendFileTextAsMessage(string file)
...
 		}
 
 		/// <summary>
 		/// Kills the player.
 		/// </summary>
 		public virtual void KillPlayer()
 		{
-			NetMessage.SendPlayerDeath(Index, PlayerDeathReason.LegacyDefault(), 99999, (new Random()).Next(-1, 1), false, -1, -1);
+			NetMessage.SendPlayerDeath(Index, PlayerDeathReason.LegacyDefault(),
+				99999, (new Random()).Next(-1, 1), false, -1, -1);
```
punchready: Same here; call the second overload.

TShockAPI/TSPlayer.cs
```diff
@@ -1840,6 +1919,7 @@ public virtual void SetTeam(int team)
 		/// Sets the player's pvp.
 		/// </summary>
 		/// <param name="mode">The state of the pvp mode.</param>
+		/// <param name="withMsg">Is it necessary to send a message about changing pvp mode</param>
```
punchready: The grammar is slightly off, "Whether a chat message about the change should be sent." would be better.

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -449,6 +499,50 @@ public static bool OnPlayerCommand(TSPlayer player, string cmdName, string cmdTe
...
+			if (PrePlayerCommand == null)
+				return false;
+			PrePlayerCommandEventArgs playerCommandEventArgs = new PrePlayerCommandEventArgs()
+			{
+				Player = player,
+				Command = command,
+				CommandArgs = args
+			};
+
+			args = playerCommandEventArgs.CommandArgs;
```
punchready, line 521: This line is probably supposed to be after the hook invocation.

TShockAPI/DB/UserManager.cs
```diff
@@ -183,6 +185,38 @@ public void SetUserGroup(UserAccount account, string group)
...
+		public void SetUserGroup(TSPlayer author, UserAccount account, string group)
+		{
+			Group grp = TShock.Groups.GetGroupByName(group);
+			if (null == grp)
+				throw new GroupNotExistsException(group);
+
+			if (_database.Query("UPDATE Users SET UserGroup = @0 WHERE Username = @1;", group, account.Name) == 0)
+				throw new UserAccountNotExistException(account.Name);
+
+			if (AccountHooks.OnAccountGroupUpdate(account, author, ref grp))
```
punchready, line 203: This is not how an account update hook should be implemented:
- This still doesn't get invoked when a plugin changes a user's group with the other overload
- This gets invoked after the change has already been made in the database, not allowing proper cancellation
- This throws if canceled?? It should probably either not be cancelable or have a clear return value indicating whether the change was canceled, or cancel by setting the group to the previous group of the user silently

TShockAPI/Hooks/AccountHooks.cs
```diff
@@ -39,6 +40,26 @@ public AccountCreateEventArgs(UserAccount account)
...
 
+	public class AccountGroupUpdateEventArgs
+	{
+		public bool Handled = false;
+
+		public TSPlayer Author { get; private set; }
+		/// <summary>
+		/// Contains only the user name. Use <see cref="UserAccountManager.GetUserAccountByName(string)"/> for more information.
+		/// </summary>
+		public UserAccount Account { get; private set; }
```
punchready: This event should either only contain a user account name field or ensure that a full account instance has been fetched to not be confusing to consumers.

punchready (changes requested):

TShockAPI/Hooks/AccountHooks.cs
AgaSpace: > This event should either only contain a user account name field or ensure that a full account instance has been fetched to not be confusing to consumers.

Should I rename the next property to "AccountName" or should I leave it as it was, which is "Account"?

punchready: `string AccountName` or `UserAccount Account`

AgaSpace:
@punchready, Any other changes needed for the merge?

TShockAPI/DB/UserManager.cs
```diff
@@ -183,6 +185,38 @@ public void SetUserGroup(UserAccount account, string group)
...
+		/// <param name="account">The user account</param>
+		/// <param name="group">The user account group to be set</param>
+		public void SetUserGroup(TSPlayer author, UserAccount account, string group)
+		{
+			Group grp = TShock.Groups.GetGroupByName(group);
+			if (null == grp)
+				throw new GroupNotExistsException(group);
+
+			if (AccountHooks.OnAccountGroupUpdate(account, author, ref grp))
+				throw new ArgumentException("The hook blocked the change of the user group.");
```
punchready: This should throw a custom exception class at least.

TShockAPI/Hooks/AccountHooks.cs
```diff
@@ -39,6 +40,23 @@ public AccountCreateEventArgs(UserAccount account)
 		}
 	}
 
+	public class AccountGroupUpdateEventArgs
+	{
+		public bool Handled = false;
+
+		public TSPlayer Author { get; private set; }
```
punchready, line 80: Again, an account group update event is quite pointless if it only fires when changing a group via command and not from all other calls to the `SetUserGroup` method.

punchready (changes requested):
Account group change event needs to be reworked, otherwise LGTM.

TShockAPI/Hooks/AccountHooks.cs
AgaSpace: Then I need to specify that Author can be null (that is, I will add documentation to the Author property that explains this) and then add a hook call to the SetUserGroup overload?

punchready: I would make an abstract `AccountGroupUpdateEventArgs` base class and then have a class for plugin-initiated events like `AccountGroupUpdateByPluginEventArgs` and one for player-initiated events like `AccountGroupUpdateByPlayerEventArgs`.

AgaSpace: > I would make an abstract `AccountGroupUpdateEventArgs` base class and then have a class for plugin-initiated events like `AccountGroupUpdateByPluginEventArgs` and one for player-initiated events like `AccountGroupUpdateByPlayerEventArgs`.

Do I need to pass `TerrariaPlugin` to `AccountGroupUpdateByPluginEventArgs`?

punchready: That's probably too much effort for too little use cases.

sgkoishi: Account group update can only be triggered by either `/user` with `tshock.superadmin.user` permission - superadmin, server console, or plugin. Maintainers can access the database and bypass the hook, and so do the plugins.
What's the use case of this hook instead of using MonoMod?

AgaSpace: > Account group update can only be triggered by either `/user` with `tshock.superadmin.user` permission - superadmin, server console, or plugin. Maintainers can access the database and bypass the hook, and so do the plugins. What's the use case of this hook instead of using MonoMod?

Are you saying that it would still be better to keep 1 class `AccountGroupUpdateEventArgs`?

sgkoishi: > Are you saying that it would still be better to keep 1 class `AccountGroupUpdateEventArgs`?

At least it is confusing for me to tell which is the initiator if the plugins are chained together.

<hr>

If a plugin dev wants to be triggered when `SetUserGroup`, they can just go this (without this PR) - the whole group change event is like saving three lines of code
```csharp
using TShockAPI.DB;

new MonoMod.RuntimeDetour.Hook(
    () => TShock.UserAccounts.SetUserGroup(null, null), 
    (Action<UserAccountManager, UserAccount, string> originalSetGroup, UserAccountManager mgr, UserAccount acc, string a) => {
        Console.WriteLine("Before the group change");
        if (happyEnough && goodLuck)
        {
            originalSetGroup(mgr, acc, a);
        }
        Console.WriteLine("After the group change");
    });
```
now becomes
```csharp
using TShockAPI.DB;

Hooks.AccountHooks.AccountDelete += args => {
        Console.WriteLine("Before the group change");
        if (!happyEnough || !goodLuck)
        {
            args.Handled = true;
        }
        Console.WriteLine("Still before the change, no hook for after the group change");
    };
```

AgaSpace: @sgkoishi, In this way, you can do everything at all. Then why do we need other hooks?
I mean, our way would be more convenient.

AgaSpace: @sgkoishi Generally speaking, do you want me to add `MonoMod.RuntimeDetour.Hook` to the SetUserGroup method? Then how do I pass the "author" of that action?

sgkoishi: I would say the current version is quite ok and enough - the MonoMod one can provide post-event but I couldn't see much usage of it, it really doesn't matter that much if something happens right before/after a group change. The only required thing I think is the pre-event that can cancel it.

AgaSpace:
Since I have been asked to split this PR into several smaller PRs, I am closing this PR.

PR #2947, opened 2023-05-14
Added AccountGroupUpdate hook, when trying to change a user group.

AgaSpace:
Added a hook to change the user group. The hook is in `AccountHooks` and is named `AccountGroupUpdate`.
The event takes `AccountGroupUpdateEventArgs` as an argument, but it can be represented as `AccountGroupUpdateByPlayerEventArgs` in case a user group is updated by some player.

Examples of use:
```csharp
public void OnAccountGroupUpdate(AccountGroupUpdateEventArgs args)
{
	if (args.AccountName == "Zoom L1" && args is AccountGroupUpdateByPlayerEventArgs)
	{
		TSPlayer author = ((AccountGroupUpdateByPlayerEventArgs)args).Player;
		if (author.Account.Name != "Zoom L1")
			args.Handled = true;
	}
	else
	{
		TShock.Log.ConsoleInfo($"The user group {args.AccountName} has been updated.");
	}
}
```
```csharp
public void OnAccountGroupUpdate(AccountGroupUpdateEventArgs args)
{
	if (args is AccountGroupUpdateByPlayerEventArgs)
	{
		TSPlayer author = ((AccountGroupUpdateByPlayerEventArgs)args).Player;
		UserAccount account = TShock.UserAccounts.GetUserAccountByName(args.AccountName);
		if (account != null) // Must be used, because the user can be null.
		{
			if (account.Group.ToLower() == "superadmin" && author.Group is not SuperAdminGroup)
				args.Handled = true;
		}
	}
}
```


TShockAPI/DB/UserManager.cs
```diff
@@ -619,14 +654,28 @@ public UserAccountExistsException(string name)
...
 
+	/// <summary>The UserGroupUpdateLockedException used when the user group update failed and the request failed as a result.</summary>.
+	[Serializable]
+	public class UserGroupUpdateLockedException : UserAccountManagerException
+	{
+		/// <summary>Creates a new UserGroupUpdateLockedException object.</summary>
+		/// <param name="name">The name of the user who failed to change the group.</param>
+		/// <returns>New UserGroupUpdateLockedException object with a message containing the name of the user account that failed to change the group.</returns>
+		public UserGroupUpdateLockedException(string name) :
+			base(GetString($"The user {name} group could not be updated"))
```
Arthri: The exception message is confusing, `The user John group could not be updated`

Arthri (changes requested):

Arthri (dismissed):

Arthri (approved):

PR #2951, opened 2023-06-06
Group changes

AgaSpace:
I added "virtual" to all properties and methods.
Updated the R, G, B fields. These are now properties.
I added a property `Color` that gets and updates the values R,G,B but as `Microsoft.Xna.Framework.Color`.

The general idea of this PR is to allow developers to create their own groups, like `SuperAdminGroup`, but with more editing capabilities.


TShockAPI/Group.cs
```diff
@@ -154,15 +156,29 @@ public virtual List<string> TotalPermissions
...
+		public virtual byte R { get; set; } = 255;
 		/// <summary>
 		/// The group's chat color green byte.
 		/// </summary>
-		public byte G = 255;
+		public virtual byte G { get; set; } = 255;
 		/// <summary>
 		/// The group's chat color blue byte.
 		/// </summary>
-		public byte B = 255;
```
Arthri: In my opinion, these should be kept as fields for now. Field to property is a source and binary breaking change

AgaSpace: Can I mark them [Obsolete] and use only Color? @Arthri

AgaSpace: > Can I mark them [Obsolete] and use only Color? @Arthri

image

Arthri: I think it's best left as-is for now. Also Color.R/G/B is read-only, it can't be set to(`error CS1612: Cannot modify the return value of 'Group.Color' because it is not a variable`)

AgaSpace: I thought of this implementation: there will be a field of type Color called `_color` and property `Color`. The `Color` will be virtual to allow developers to change the color more correctly. To update the color - you will need to assign a value to the `Color` property.

The `_color` field will have to be `protected`.

If you look at how the color is set (say in the command /group color), it is set only together with other colors (for example, when changing B required R and G) `(Proper syntax: {0}group color <group name> [new color(000,000,000)]`)

AgaSpace: You can't assign a value to Color.R because it's a structure. Nothing will happen if the developer does `new Color(color.R, color.G, someRandomNumber)` a second time (I mean, it shouldn't complicate development)

AgaSpace: I don't really know what to write in `Obsolete`, so I gave the example "Use Group.Color...". I can write that to change `R` you need to assign a new value to the `Color` property and add `Use <see cref="Group.Color"/>` to the documentation.

Arthri: I mean, it's best to leave them as-is right now. Public byte fields without the obsolete attribute. In that manner it minimizes breaking changes and plugin authors having to possibly recompile. Perhaps in the next major update, what you proposed can be reconsidered

AgaSpace: @Arthri sorry to bother you, but is everything correct in my code now?

Arthri: Sorry, I reviewed the AccountHooks PR and forgot about this

Arthri (dismissed):

Arthri (approved):

PR #2960, opened 2023-06-15
New two hooks PrePlayerCommand and PostPlayerCommand

AgaSpace:
There is a hook `PlayerHooks.PlayerCommand` which is called before using a command. In my personal opinion, this hook is obsolete because it could be used before calling a command (`cmd.Run` in `HandleCommand`). This would make it easier to work with commands because you wouldn't have to look for a command that might not exist in `Commands.ChatCommands`. That's why I added the `PlayerHooks.PrePlayerCommand` hook.
Also, the `PlayerCommand` hook does not implement the ability to handle commands after they are used, so I added the `PostPlayerCommand` hook.

If you would like to add "invisible" commands, which can be done with PlayerCommand, we already have a similar alternative (TSPlayer.AwaitingResponse).

Now some examples of how to use hooks:
```csharp
PlayerHooks.PrePlayerCommand += (args) =>
{
	if (args.Command.Names.Contains("who"))
	{
		args.Arguments.Player.SendInfoMessage("Staff Online: {0}",
			string.Join(", ", TShock.Players.Where(i => i?.Active == true && i.HasPermission("staffonline"))));
	}
};
```
```csharp
PlayerHooks.PostPlayerCommand += (args) =>
{
	if (args.Command.Names.Contains("who"))
	{
		int somePlayersInDimensions = 0;
		args.Arguments.Player.SendInfoMessage("There are {0} players in dimensions.",
			somePlayersInDimensions);
	}
};
```
```csharp
PlayerHooks.PrePlayerCommand += (args) =>
{
	if (args.Command.Names.Contains("group"))
	{
		List<string> parameters = args.Arguments.Parameters;
		if (parameters.Count >= 2 && parameters[1].ToLower() == "owner")
		{
			TSPlayer player = args.Arguments.Player;
			player.SendErrorMessage("I'm sorry, but we have a monarchy. You are now banned.");

			player.SilentKickInProgress = true;
			player.Disconnect("You are banned!\n(but that's a joke)");
			args.Handled = true;
		}
	}
};
```


sgkoishi:
If there are already two `/who` commands, using this hook will result in two `Staff Online: a, b, c` or `There are 5 players in dimensions`; it's understandable for this hook but not quite expected for the actual use case 🤔

AgaSpace:
> If there are already two `/who` commands, using this hook will result in two `Staff Online: a, b, c` or `There are 5 players in dimensions`; it's understandable for this hook but not quite expected for the actual use case 🤔

You can get the command you need using `Commands.ChatCommands.Find`, but I was too lazy to describe it :p

Although yes, you're right. It was inappropriate to include this example in the description of my changes.

AgaSpace:
> If there are already two `/who` commands, using this hook will result in two `Staff Online: a, b, c` or `There are 5 players in dimensions`; it's understandable for this hook but not quite expected for the actual use case 🤔

Then how about marking the old hook (PlayerCommand) as obsolete?

AgaSpace:
@sgkoishi Is everything correct now?

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -119,6 +120,53 @@ public class PlayerCommandEventArgs : HandledEventArgs
...
+		/// </summary>
+		public Command Command { get; }
+		/// <summary>
+		/// Command arguments.
+		/// </summary>
+		public CommandArgs Arguments { get; }
+		/// <summary>
+		/// Is the command executed.
+		/// </summary>
+		public bool Handled { get; }
```
Arthri: this class should inherit from `HandledEventArgs` instead of redefining `Handled`

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -341,8 +389,29 @@ public static class PlayerHooks
 		/// <summary>
 		/// Fired by players when using a command.
 		/// </summary>
+		[Obsolete("There is an alternative to PlayerHooks.PrePlayerCommand")]
```
Arthri: The obsolete message says `PlayerCommand` is an alternative, rather than use `PrePlayerCommand`

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -341,8 +389,29 @@ public static class PlayerHooks
...
 		public static event PlayerCommandD PlayerCommand;
 
+		/// <summary>
+		/// The delegate of the <see cref="PrePlayerCommand"/> event.
+		/// </summary>
+		/// <param name="e">The EventArgs for this event.</param>
+		public delegate void PrePlayerCommandD(PrePlayerCommandEventArgs e);
+		/// <summary>
+		/// Fired before the command is run.
```
Arthri:
```suggestion
		/// Fired before a command is run.
```

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -341,8 +389,29 @@ public static class PlayerHooks
...
+		/// </summary>
+		public static event PrePlayerCommandD PrePlayerCommand;
+
+		/// <summary>
+		/// The delegate of the <see cref="PostPlayerCommand"/> event.
+		/// </summary>
+		/// <param name="e">The EventArgs for this event.</param>
+		public delegate void PostPlayerCommandD(PostPlayerCommandEventArgs e);
+		/// <summary>
+		/// Fired after the command is run.
```
Arthri:
```suggestion
		/// Fired after a command is run.
```

Arthri (changes requested):

Arthri (dismissed):

TShockAPI/Hooks/PlayerHooks.cs
```diff
@@ -341,8 +385,29 @@ public static class PlayerHooks
 		/// <summary>
 		/// Fired by players when using a command.
 		/// </summary>
+		[Obsolete("Use PlayerHooks.PrePlayerCommand.")]
 		public static event PlayerCommandD PlayerCommand;
```
Arthri, line 388: I think this could be possibly renamed to something else, because this delegate has a use: it can remove and add commands before they are checked.

Arthri (approved):

PR #2966, opened 2023-09-05
Update server autostart instructions

HerrSubset:
Specifying a combination of -world, -port and -maxplayers did not auto-start the TShock server for me. I found that -world and -autocreate were required. The other settings could be read from config.json.

As such, I updated the documentation.


hakusaro (approved):

hakusaro:
I applied these changes to the wiki to simplify the PR process etc, so that we don't have to merge a lot of markdown changes.

PR #2899, opened 2023-01-16
Fixed docker builds for arm64 and updated docker documentation

yeraylopez:
Might need @PotatoCider to review changes in .dockerignore for the possibility that dependencies are added in the future.

PR #2971, opened 2023-09-22
Rewrite the .dockerignore file into a denylist

timschumi:
This should help with not forgetting to add any new directories (such as TShockInstaller and TShockPluginManager, which have been missing until now).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,15 +1,20 @@
-# ignore every file
-*
+# Ignore all Git metadata and build output (in all directories).
+**/.git*
+**/bin/
+**/obj/
 
-# except for the ones required for building
-!i18n/
-!prebuilts/
-!TerrariaServerAPI/
-!TShockAPI/
-!TShockLauncher/
-!TShockLauncher.Tests/
-!TShock.sln
-
-# but exclude build artifacts
-*/bin/
-*/obj/
+# Ignore other specific files that aren't needed for the build itself.
+/.all-contributorsrc
+/.dockerignore
+/.editorconfig
+/.vscode
+/appveyor.yml
+/COPYING
+/crowdin.yml
+/Dockerfile
+/docs
+/README.md
+/README_cn.md
+/renovate.json
+/scripts
+/SECURITY.md
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -89,6 +89,7 @@ Use past tense when adding new entries; sign your name off when you add or chang
 * Added a method `TSPlayer.GiveItem`, which has `TShockAPI.NetItem` structure in its arguments. (@AgaSpace)
 * Added a property `TSPlayer.Hostile`, which gets pvp player mode. (@AgaSpace)
 * Fixed typo in `/gbuff`. (@sgkoishi, #2955)
+* Rewrote the `.dockerignore` file into a denylist. (@timschumi)
 
 ## TShock 5.2
 * An additional option `pvpwithnoteam` is added at `PvPMode` to enable PVP with no team. (@CelestialAnarchy, #2617, @ATFGK)


drunderscore (dismissed):
This LGTM, I'll let Potato have a look just in-case I missed anything silly :)

pcider:
This PR is related to #2899, but it seems that PR has gone stale. I originally added `.dockerignore` because Docker builds were copying lots of large files in the `.git/` . In hindsight, ignoring every file except project files in `.dockerignore` would break Docker builds with further changes (especially in the root directory).

Could we modify `.gitignore` to just exclude git and build directories (`.git/`, `bin/`, `obj/`)? You can take a look at [`.dockerignore`](https://github.com/Pryaxis/TShock/pull/2899/files) in #2899. Alternatively, we could try merging #2899 instead.

timschumi:
> In hindsight, ignoring every file except project files in .dockerignore would break Docker builds with further changes (especially in the root directory).

We could add building (not necessarily pushing) Docker images to the CI. However, that should probably be a separate PR. :^)

> Could we modify .gitignore to just exclude git and build directories (.git/, bin/, obj/)?

Yep, sure. I'll update the PR soon.

.dockerignore
```diff
@@ -1,15 +1,20 @@
...
-!TerrariaServerAPI/
-!TShockAPI/
-!TShockLauncher/
-!TShockLauncher.Tests/
-!TShock.sln
-
-# but exclude build artifacts
-*/bin/
-*/obj/
+# Ignore other specific files that aren't needed for the build itself.
```
timschumi, line 6: Note that this isn't really required, since we copy the build output into a new container during the end of the build.

We could drop it, but it keeps both copy operations and accidental cache misses low.

ghost:
this is... BIG

timschumi:
> this is... BIG

As said, the second part of the list isn't really necessary, and it doesn't have to be kept perfectly updated. It's just nice for Docker to copy and cache-track as little number of files as possible.

For now, I decided to go with the variant that keeps equivalency to the previous suggested change. If you want me to drop it, I'm happy to.

ghost:
Your good... I'm not a reviewer for the tshock program.

pcider (approved):
LGTM!

docs/changelog.md
```diff
@@ -89,6 +89,7 @@ Use past tense when adding new entries; sign your name off when you add or chang
 * Added a method `TSPlayer.GiveItem`, which has `TShockAPI.NetItem` structure in its arguments. (@AgaSpace)
 * Added a property `TSPlayer.Hostile`, which gets pvp player mode. (@AgaSpace)
 * Fixed typo in `/gbuff`. (@sgkoishi, #2955)
+* Rewrite the `.dockerignore` file into a denylist. (@timschumi)
```
drunderscore: SUPER nitpicky, but this should be in the past-tense, as in:
> Rewrote the `.dockerignore` file into a denylist. (@timschumi)

drunderscore (commented):
Ah, this makes much more sense. Hopefully this won't break in the future. CI is worth it cause it doesn't cost us anything except time, but that's a job for a future person at a future time!

timschumi:
> CI is worth it cause it doesn't cost us anything except time, but that's a job for a future person at a future time!

Well, if you want, you can already take a look at #2974. :^)

drunderscore (approved):
Awesome, thanks Tim! This is good to go.

hakusaro:
I guess I don't personally like this being a denylist, because it means we can accidentally include things we don't want. I'll let someone else approve this and merge it.

timschumi:
> I guess I don't personally like this being a denylist, because it means we can accidentally include things we don't want. I'll let someone else approve this and merge it.

Note that "accidentally include" only means that it will get copied to the build container without being needed. It wouldn't end up in the final image (not even its history) since we copy just the `dotnet publish` output to a new clean base (unless `dotnet publish` packages up files by excessive wildcard somewhere).

If we want to do keep it as an allowlist, we should really add CI for Docker though (#2974). Otherwise it will just silently break again, just like it did for basically the entire last year.

hakusaro (approved):

PR #2988, opened 2023-10-27
Fixed typo in ProjectileManager

RenderBr:
This is a very simple pull request. I have simply bulk renamed the class that was originally `ProjectileManagager` to `ProjectileManager`. This is minor, but I felt like it should be done.


hakusaro:
Won’t this break any third party consumers of the `ProjectileManagager`?

drunderscore:
Yeah, this is both source and binary incompatible, can't rename it.
If it bothers you enough, I suppose this will have to do:
```csharp
using ProjectileManager = ProjectileManagager;
```

ghost (commented):
Your very correct, there would have to be a change in all the plugins to support this, as this is hard to do by contacting all the plugin developers I believe that this should go unchanged...

RenderBr:
Fair enough, I didn't think about this. Feel free to close this PR.

ghost:
@RenderBr if you wish to close this then close the PR with the close button

Arthri:
For the record, I'm not against this PR, but it'll have to be delayed until the next major update to prevent unnecessary breaking changes in between feature versions

ghost:
guess this is a major update thing...

PR #2989, opened 2023-11-05
Added a configuration option to suppress player chat messages in the server console

ACaiCat:
`"DisplayChatToServerConsole": true` (default value is true)
if it is set to `false`, player chat messages will not be displayed in the server console

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2854,7 +2854,7 @@ namespace TShockAPI
 			{ BuffID.Poisoned, 3600 },              // BuffID: 20
 			{ BuffID.OnFire, 1200 },                // BuffID: 24
 			{ BuffID.Confused, short.MaxValue },    // BuffID: 31 Brain of Confusion Internal Item ID: 3223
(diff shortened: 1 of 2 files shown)


ghost (commented):
I would like to see a video on how it works...

bartico6:
> I would like to see a video on how it works...

if the option is set to false, chat isn't printed in server console.

ghost:
I never see chat printed on the console, i might have an inadequate version installed

ACaiCat:
https://github.com/Pryaxis/TShock/assets/62058454/e1b6514c-7a94-49fb-8f31-915422f5dac6

ghost (approved):
This can decrease clutter in the server console if the server has multiple players in it and it (according to my test) decreases GPU use, i ran these tests on my devices and gave them out to each one in the house at the time... i still have these devices but it works well with the current version. I suggest a full review from a collaborator because i do not understand much about this code but this is a good and working modification.

ghost:
@DaBomberNTG or @hakusaro please review this Pull Request as i have approved this... if there is an incorrection in my test then feel free to request changes as i will do a new test

PR #3018, opened 2024-02-12
Fix `/wind`

punchready:
The command was randomly changed in https://github.com/Pryaxis/TShock/commit/05a4e025f7831c6163daa8fbdb1c9436e61aa45e to parse as int instead but not do any conversions. This fixes it, allows entering the value in mph, and limits the range to the safe one also accessible through the journey menu.
See https://github.com/Pryaxis/TShock/discussions/3017.

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -4622,21 +4622,22 @@ namespace TShockAPI
 		{
 			if (args.Parameters.Count != 1)
 			{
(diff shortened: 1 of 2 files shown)


Arthri (approved):

TShockAPI/Commands.cs
ghost, line 1: May cause partially broken plugins, if no plugins use this, it will work. (I do belive the changes will not break any plugins due to me not finding with the wind thingy)

ghost (approved):

ghost:
These changes will run correctly (according to my smooth as heck brain)

PR #3027, opened 2024-04-09
Add permission for Moondial and unrecognized SFX

Arthri:
- Added explicit cases for cases 2, 4, and 5 of Packet 51. These are harmless and so are approved with no checks
- Added a handler for case 6 or Enchanted Moondial of Packet 51. Added a new permission, `tshock.world.time.usemoondial`, to regulate Enchanted Moondial's use
- Added a default case for Packet 51 which instead checks for the `tshock.specialeffects.{type}` permission if TShock does not recognize the case yet


Arthri:
Fixing `DisableDungeonGuardian` requires a non-trivial amount of work, so I've decided not to do it in this PR. The Dungeon Guardian is spawned by the server, deep inside the `NPC.SpawnNPC` method, which would require IL patching and all the strings attached with that

ghost (approved):
This should work correctly.

hakusaro (approved):

Issue #3011, opened 2024-01-16
SSC Bed Spawn Not Saving


Sancus101:
Setting a bed spawn for a SSC player is not saving after they logout.

TShock version 5.2

Reproduce bug steps

1. Set bed spawnpoint
2. Logout
3. Use Magic Mirror. Brings you to vanilla spawnpoint and not your bed.

We are all getting the message for a successful spawnpoint being set as well. The bedspawn point will work until you logout and back in.

[2024-01-16_14-10-57.log](https://github.com/Pryaxis/TShock/files/13955319/2024-01-16_14-10-57.log)

Arthri:
Known issue. #1966

Sancus101:
> Known issue. #1966

Sad to hear this is a known issue for years. I checked the sql database for my server to see what is being saved. I see the spawnX and sapwnY locations are updating just fine into the sql database. I can only assume the problem is the character datamanager is not retriving the data from the sql DB correctly.

punchready:
This is because it is **impossible** to save the bed spawn position from interacting with the bed without an additional user interaction, which is why most servers have something like a `/sethome` command.

Issue #2844, opened 2022-11-25
ssc enabled, bed set respawn point problem (ssc enabled, 床设置重生点的问题)


1-erick-zhang:
床设置的重生点，只要不退出游戏就可以一直存在，但是如果重进服务器重生点将消失

请求：保存玩家在服务器设置的重生点，即使退出可以正常使用

hakusaro:
Via DeepL:

> The respawn point set in the bed will always exist as long as you don't quit the game, but if you re-enter the server the respawn point will disappear

> Request: Save the respawn point set by the player in the server, even if you exit, you can use it normally

TomAltonator:
Can confirm this also happens on my server. Bed spawns reset when you leave the server.

Issue #1966, opened 2020-06-01
Beds are not saving spawn points in DB for SSC


kiroushi:
**TShock version**: TShock 4.4.0 Pre-release 9 for Terraria 1.4.0.4

Potentially related to https://github.com/Pryaxis/TShock/discussions/1845?
#### Reproduction steps
(Happens with all level of privileges):
1. Join server (will appear at initial spawn).
2. Set spawn point to a bed (map will show bed icon).
3. Die/use mirror -> will be ported to bed.
4. Check DB spawnX & spawnY values (won't match bed coords, it will be world's initial spawn coords).
5. Logout
6. Login again: use mirror or die -> will be ported to world's initial spawn point.

Expected behaviour is spawn point persistence in DB.

ATFGK:
agree

Unrealomega:
This is still currently an issue for us. We're running on the latest public release (4.5.17). I've noticed that it sporadically updates the SpawnX/SpawnY in the DB and folks will spawn in at their bed before being immediately moved to the initial world spawn.

Bestaford:
Confirm with latest TShock

Issue #1849, opened 2020-05-22
Beds cannot be removed as spawn points


QuiCM:
Copy/pasted straight from https://github.com/Pryaxis/TShock/discussions/1845:

How it should work:

When you use Magic Mirror it teleports you back to spawn.
How it does work:

    You click on the bed and "Spawn point set!" text appears in the chat.
    You use Magic Mirror and you are teleported to the bed.
    You click on the bed and "Spawn point removed!" text appears in the chat.
    You use Magic Mirror and you are teleported to the bed.

What I expected to happen:

    You click on the bed and "Spawn point removed!" appears
    Magic Mirror teleports you back to spawn.

iFocsy:
**This problem is still happening in the most recent commit 1eeb77d , The reproduction steps are:**

1. Have Spawn Erased to default.
2. Put down a bed (Any Type).
3. Set spawn point on bed. (Set OK)
4. Use Cellphone /Magic Mirror = You get Teleported to the Bed. (OK)
5. Click again the bed to remove Spawn (It shows it got Removed)
6. You still goes to the Bed.


**Optional steps:**

1. If you put a New bed and Set spawn, you are still returned to the previous bed.
2. If you destroy the bed, your next spawn will be the following bed you tried, if not, return to worldspawn.

**Note:**
No plugins, only SSC built-in enable.

sk9c00:
**I can confirm this issue is still happening in SSC, albeit with a minor functional improvement.**

1. Yes, the right-click remove spawn point function is still **broken**. Has been even after the patch. (Though I suspect it's because the 'fix' was only to check if the bed still exists or not and not if the right-click to remove spawn action was completed or not. I'm beginning to suspect that this is a client side authoritative issue, and that without a special 'packet' or dedicated way to detect/track if this right-click action was completed legitimately, this 'SSC' loop-hole will continue to exist in eventually three states: either trust the client, disable/limit the functionality, or some sort of 'hack' using the warp system. Possible thing to ask the original game devs to improve/assist on?)

2. The 'improvement' is basically as mentioned above by @iFocsy, if you remove the bed regardless of whether or not you've right-clicked on the bed, you will get teleported back to spawn. **However, if the bed is placed back in it's location, you will get teleported back.** This occurs regardless of logging out, restarting the server, etc. Short of manually editing the `spawnX`, `spawnY` of the player back to `-1` during server downtime, the spawn location essentially _persists_.

hakusaro:
[...]

PR #3031, opened 2024-06-02
Rewrite: Partially fix bed spawning for SSC

pcider:
This PR rewrites the player spawnpoint saving system for SSC.

When SSC is enabled, player spawnpoints are only saved server side but not client side. However, due to the game's netcode logic, it is most likely [impossible](https://github.com/Pryaxis/TShock/issues/3011#issuecomment-1894722408) for the server (TShock) to change the client's spawnpoint value `Terraria.Player.Spawn{X,Y}` since the server provided value gets overwritten by its own client sided spawnpoint value in `sp{X,Y}` in `Terraria.Player.FindSpawn()` after packet 12 has been received. 

In order to workaround this, TShock would keep track of the client's initial spawnpoint value on the first player spawn, it uses this value to compare if the player has changed his spawnpoint. This way, we can teleport the player with the correct server saved spawnpoint after the player tries to spawn with its incorrect client sided spawnpoint, at least before the player changes his spawnpoint mid-game.

What this means is that we purposely desync the client and server `Terraria.Player.Spawn{X,Y}` value before we know that the player had changed his spawnpoint. Once we know its changed, both client and server spawnpoint values would be synced, and we won't need to force teleport the player anymore. 

Note that there is one limitation with this approach - **this only works if the player respawns (via death or magic mirror) at his changed spawnpoint before logging off**, otherwise the server would have no idea that the player's spawnpoint has changed.

I'm not sure if I covered all test cases, but please let me know if there is any bugs :)

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2248,7 +2248,7 @@ namespace TShockAPI
 
 			var args = new SyncTilePickingEventArgs
 			{
-                                Player = player,
+				Player = player,
 				PlayerIndex = playerIndex,
 				TileX = tileX,
 				TileY = tileY,
@@ -2719,8 +2719,8 @@ namespace TShockAPI
 			}
 
 			byte player = args.Data.ReadInt8();
-			short spawnx = args.Data.ReadInt16();
-			short spawny = args.Data.ReadInt16();
+			short spawnX = args.Data.ReadInt16();
+			short spawnY = args.Data.ReadInt16();
 			int respawnTimer = args.Data.ReadInt32();
 			short numberOfDeathsPVE = args.Data.ReadInt16();
 			short numberOfDeathsPVP = args.Data.ReadInt16();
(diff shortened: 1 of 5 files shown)


hakusaro:
What problem are you trying to solve here?

pcider:
> What problem are you trying to solve here?

I'm mainly trying to solve #1966 (and #2844, #3011) which were known issues for quite some time. This PR doesn't fully solve the bed spawning problem with the limitations above. There are also other related issues like #1849.

pcider:
Need testers for this PR! You can help test out with this Docker image with:
`docker run -p 7777:7777 -v $(pwd)/tshock:/tshock -v $(pwd)/worlds:/worlds --rm -it ghcr.io/potatocider/tshock:general-devel`
(This will create two folders `tshock/` and `worlds/` in your current working directory. You can then enable SSC via `tshock/sscconfig.json`)

Please do let me know if there's any bugs or potential issues with this workaround.

xNarnia:
Upon dying and respawning, any action I take seems to not register.
Can anyone reproduce this?

**Steps to Reproduce**
- Run & initialize the server
  - `docker run -p 7777:7777 --rm -it ghcr.io/potatocider/tshock:general-devel`
- Enter the world
- Die
- Attempt to place a block or hurt enemies
  - The enemies will not register your hits server-side
- Leave and re-enter the server
  - The blocks placed before will be gone as if never placed

xNarnia:
Also, there seems to be some debugging output in the console. Is this intended?
image

pcider:
@xNarnia 
> Also, there seems to be some debugging output in the console. Is this intended?

Oops. That's not intended.

I've fixed the death issue you mentioned, could you help test it again?

xNarnia:
Tested scenarios in https://github.com/Pryaxis/TShock/issues/1966, https://github.com/Pryaxis/TShock/issues/2844, https://github.com/Pryaxis/TShock/issues/3011, and https://github.com/Pryaxis/TShock/issues/1849. All passing with no errors.

Also tested some scenarios with bombs destroying beds during death.
It works similarly to previous versions of TShock, where the client sees destroyed blocks but it doesn't register on the server.
In this scenario with the new change, the user is teleported back to the spawnpoint after the blocks are destroyed during death. However when relogging, the destroyed blocks and destroyed bed return. The player's spawnpoint is set back to the bed, and that is where they spawn in.

Tested with both SSC on & off.

Looks good to me! ✅

xNarnia (approved):

xNarnia (changes requested):
As per our conversation in Discord, Potato will implement the use case for the bed being broken

pcider:
Okay, the case where beds are broken/invalid is now handled. This PR should be ready for merging ^^

xNarnia:
Tested bed breaking with both the magic mirror and death.
Also tested `Player A` setting spawn point, `Player A` exiting server, `Player B` destroying `Player A's` bed spawnpoint, `Player A` logging back in. `Player A` returns to the world spawnpoint as intended.

xNarnia (approved):

hakusaro (approved):

Issue #2914, opened 2023-02-11
When tshock start with -lang 7 ,the /help command can't work.


fireflyoo:
* TShock version:5.1.3 for Terraria 1.4.4.9
* TShock build number (if known):?

#### Reproduction steps (if applicable)?

1. start tshock with "-lang 7"
2. then join in Game with Terraria Client..
3. type "/help" in the game.
4. it doesn't work,but in Server console it works well.

sgkoishi:
I actually saw lots of such reports with the error message `Invalid command entered. Type /help for a list of valid commands.` (and its `-lang 7` translation: `输入的命令无效。使用{0}help 获取可用的指令列表。`)
I was never able to reproduce it, but most said
* `.help` works but `/help` shows the error
* magically fixed after re-download

Some additional info:
* Only reproducible with mobile client
* Many of them are known to be using translated client (might be the cause)

HAHH9527:
I have same problem, I try disable `官方Wiki中文语言包`, `/help` command is work.
So `using translated client` maybe is main reason.

---

The following information is intended to make it easier for Chinese-speaking players to find this issue and troubleshoot it.
如果你碰到了`输入的命令无效。使用/help 获取可用的指令列表`，请尝试禁用语言包类mod。

sgkoishi:
> I have same problem, I try disable `官方Wiki中文语言包`, `/help` command is work. So `using translated client` maybe is main reason.
> 
> The following information is intended to make it easier for Chinese-speaking players to find this issue and troubleshoot it. 如果你碰到了`输入的命令无效。使用/help 获取可用的指令列表`，请尝试禁用语言包类mod。

The cause is not the language pack. The `官方Wiki中文语言包` is not doing anything wrong, but TShock is the source of the bug. Follow the above link in yaaiomi:

> Terraria will translate chat commands into command id. TShock translate them back to keep the command working.
> However, when the server and the client have different locale, a `enUS` player send `/help` will be sent as `CommandId: Help` and a `deDE` server will translate it back to `/hilfe`, thus the command is broken.

PR #3039, opened 2024-07-31
fix: /help, /me, and /p commands can't work in non-English languages

ACaiCat:
### When TShock runs in a non-English language, vanilla commands prefix will be localized.
eg.(Chinese) `/help` (English command prefix) => `/帮助`  (Chinese command prefix) 

### `TShock.OnChat` cannot correctly handle vanilla command prefixes
e.g.(Chinese) When a player use `/help`, TShock will execute `/帮助` (causing the command to not be found).
image

> Invalid command entered. Type /help for a list of valid commands.

### This PR add `EnglishLanguage.GetCommandPrefixByName` to get English vanilla command prefix.
image


ACaiCat:
Solve #2914

ACaiCat:
Could someone review this PR?

hakusaro (approved):

hakusaro:
Can someone else look at this too?

TShockAPI/Localization/EnglishLanguage.cs
```diff
@@ -71,6 +75,15 @@ internal static void Initialize()
...
 				}
+
+				ChatInitializer.Load();
+				foreach (var command in ChatManager.Commands._localizedCommands)
+				{
+					if (VanillaCommandsPrefixs.ContainsKey(command.Value._name))
+						continue;
+					VanillaCommandsPrefixs.Add(command.Value._name,command.Key.Value);
+				}
+				ChatManager.Commands._localizedCommands.Clear();
```
Arthri, line 86: This line effectively disables TShock processing Terraria commands, because TShock uses `_localizedCommands` here https://github.com/ACaiCat/TShock/blob/7419205548fb9779f83bc924222c1df9741ddff5/TShockAPI/TShock.cs#L1494

TShockAPI/Localization/EnglishLanguage.cs
```diff
@@ -71,6 +75,15 @@ internal static void Initialize()
 					var i = (int)field.GetValue(null);
 					Prefixs.Add(i, Lang.prefix[i].Value);
 				}
+
+				ChatInitializer.Load();
+				foreach (var command in ChatManager.Commands._localizedCommands)
```
Arthri, line 80: This doesn't seem to request the English localization?

Arthri (changes requested):

sgkoishi:
This is requesting the enUS locale as set earlier
https://github.com/Pryaxis/TShock/blob/0b6bf9ef4050f39f468d6782ea2992d8ce82e8bb/TShockAPI/Localization/EnglishLanguage.cs#L50

I guess the `Clear` was done for similar reason like my implementation (but I did set the `_localizedCommands` back after)
https://github.com/sgkoishi/yaaiomni/blob/debf82951dd84bd307fc7bce06a1c03160479976/Core/Enhancements.cs#L129

TShockAPI/Localization/EnglishLanguage.cs
sgkoishi: Just double checked and this Clear happened right before vanilla `Main.Initialize`, and it was empty before we call `ChatInitializer.Load`.

bartico6:
> Can someone else look at this too?

I would, but I don't understand the issue very well.. Can you ELI5 what is wrong?

sgkoishi:
> > Can someone else look at this too?
> 
> I would, but I don't understand the issue very well.. Can you ELI5 what is wrong?

Terraria will translate chat commands into command id. TShock translate them back to keep the command working.
However, when the server and the client have different locale, a `enUS` player send `/help` will be sent as `CommandId: Help` and a `deDE` server will translate it back to `/hilfe`, thus the command is broken.

Arthri (approved):
Tested and can confirm that it works as intended

PR #3056, opened 2025-01-26
Automatic language detection

### LINKED ISSUE #2957: Unable to detect system locale
https://github.com/Pryaxis/TShock/blob/a7360b635a38b57b11236f7386ab8532d0374ce4/docs/lang.md?plain=1#L1

Theoretically TShock will detect the locale, but `Terraria.Program.RunGame` (`orig_RunGame` for OTAPI) calls `LanguageManager.SetLanguage(GameCulture.DefaultCulture)`, which indirectly overwrite the `CurrentUICulture` to `enUS`.

Thus, the `I18n.TranslationCultureInfo` is always `enUS` if no `TSHOCK_LANGUAGE`, `-lang` or `-language`.
https://github.com/Pryaxis/TShock/blob/a7360b635a38b57b11236f7386ab8532d0374ce4/TShockAPI/I18n.cs#L63

The call of `SetLanguage` happens before TShock and plugins, so the original value of `CurrentUICulture` is no longer available.

Some potential fix:
* Set the (private) field `CultureInfo.s_currentThreadUICulture` to null.
* Read the (private) property `CultureInfo.GetUserDefaultCulture`.
* Read the public property `CultureInfo.InstalledUICulture`.

sgkoishi:
Set the (private) field `CultureInfo.s_currentThreadUICulture` to null and retrieve the original default language

Fix #2957

### DIFF
--- a/TShockAPI/I18n.cs
+++ b/TShockAPI/I18n.cs
@@ -60,6 +60,12 @@ namespace TShockAPI
 					}
 				}
 
+				if (LanguageManager.Instance.ActiveCulture == GameCulture.DefaultCulture)
+				{
+					var bf = System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static;
+					// LanguageManager.SetLanguage will change this so we need to reset it back to null
+					typeof(CultureInfo).GetField("s_currentThreadUICulture", bf)?.SetValue(null, null);
+				}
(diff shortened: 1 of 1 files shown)


hakusaro (approved):

hakusaro:
@sgkoishi  can you update the changelog on the wiki? thanks xxx

PR #3057, opened 2025-01-26
Detect xterm compatibility to avoid console spam

sgkoishi:
This helps #2820. Some Linux control panels and random screem/tmux settings might also experiencing this.

Regression: some Linux terminals might got excluded and will not get their title set even when they do supports it (but I haven't actually find one, perhaps not that common).


hakusaro (approved):

hakusaro:
@sgkoishi blah blah blah, changelog, blah blah blah, wiki

PR #3063, opened 2025-02-28
Fixes related to `SendTileRectHandler`

LaoSparrow:
Fixed incorrect validating range in `TileRectMatch.MatchRemoval`.
Fixed tile rect changes (e.g. turning on and off campfires) are not synced between clients.
Fixed unable to place Hat Rack without permission `tshock.ignore.sendtilesquare`.

### DIFF
--- a/TShockAPI/Handlers/SendTileRectHandler.cs
+++ b/TShockAPI/Handlers/SendTileRectHandler.cs
@@ -327,7 +327,7 @@ namespace TShockAPI.Handlers
 
 			private bool MatchRemoval(TSPlayer player, TileRect rect)
 			{
-				for (int x = rect.X; x < rect.Y + rect.Width; x++)
+				for (int x = rect.X; x < rect.X + rect.Width; x++)
 				{
 					for (int y = rect.Y; y < rect.Y + rect.Height; y++)
 					{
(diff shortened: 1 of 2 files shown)


hakusaro:
@LaoSparrow is this code from a fork of TShock perhaps? Why did you change the types of Matches etc., to a `MatchResult` enum in the same commit? I assume this is because you've already done this in a fork and want to upstream these changes...?

LaoSparrow:
The use of `MatchResult` is to distinguish between these three cases:
- NotMatched: the tile rect is not matched (e.g. tileid not matched, size not matched etc. ). In this case we will skip to the next match
- RejectChanges: changes from a player is rejected (e.g. doesn't have a build permission etc. ). In this case the tiles on the server side are unchanged, and we want to revert the changes on the player's client
- BroadcastChanges: changes from a player is accepted (e.g. normal operations, like turning on or off campfires). In this case the tiles on the server side are changed, and we want to broadcast these changes to other players as well

So the purpose of adding `MatchResult` is to solve this issue:
> Fixed tile rect changes (e.g. turning on and off campfires) are not synced between clients.

LaoSparrow:
and yes I would like to upstream these bugfixes

punchready (approved):

hakusaro (approved):

hakusaro:
@LaoSparrow 
lgtm_trick

PR #3066, opened 2025-03-10
Fix invasions not working

ohayo:
Uhm, sorry. I'll stop contributing to TShock once this is merged.
Tested and works now, I never thought to test invasions when implementing this check and for some reason it blocks invasions such as the pirate invasion from summoning? Hm..


TShockAPI/GetDataHandlers.cs
```diff
@@ -3720,7 +3720,7 @@ private static bool HandleSpawnBoss(GetDataHandlerArgs args)
 					break;
 			}
 
-			if (NPCID.Sets.MPAllowedEnemies[thingType])
+			if (!invasions.Contains(thingType) && NPCID.Sets.MPAllowedEnemies[thingType])
```
sgkoishi: The logic looks weird. If `MPAllowedEnemies[thingType]` is true, thing must be a positive number. `invasions` are negatives so definitely they can't contains a positive number.

```suggestion
			if (invasions.Contains(thingType) || NPCID.Sets.MPAllowedEnemies[thingType])
```

<hr>

... or perhaps we can just remove the `if` check here - if the NPC is un-spawnable, we should reject it instead of hiding it from the log, right?

ohayo: Well, before we were logging every npc that was summoned for some reason (Stuff like "the bunny", etc which arent spawnable but alright). I thought this check would prevent spam from bad actors on servers with anonymous boss invasions turned off, however, it turns out this has caused more harm than good.

ohayo:
Tested and appears to work, bosses spawn fine & invasions.

image
image
 (Anonymous Boss Invasions turned off)

ohayo:
image
image

Seems a lot better than before, nothing related to "the bunny", or "the nurse", etc.

hakusaro:
PR not mergable

PR #3082, opened 2025-03-25
fix(SendTileRectHandler): unable to place LogicSensors

LaoSparrow:
Fixed unable to place LogicSensors (@LaoSparrow)

Maximum frameX should be 18 (inactivated=0, activated=18)
Maximum frameY should be 108
image


LaoSparrow:
~~looks like clients only send TileRect with frameX 0 when these sensors are placed, is it better to be stricter (maxFrameX=0) or not?~~
ignore what i have said, they send both

I some how think there are more   wrong matches/rules similar to this, anyone testing?

hakusaro (approved):

hakusaro:
@LaoSparrow can you update the changelog on the wiki?

LaoSparrow:
ok

PR #3088, opened 2025-04-03
Fix losing a item in ItemFrame upon interacting it on protected region you do not have access

NightKLP:
## this is the best thing i could come up with
- i tested it in my main server
> - where spamming it doesn't dupe
> - you do not lose your item modifier/prefix
- it is a common issue when players lose their item which cause by a itemframe
- if there's better ways than this i don't know honestly


TShockAPI/Bouncer.cs
```diff
@@ -2491,6 +2491,11 @@ internal void OnPlaceItemFrame(object sender, GetDataHandlers.PlaceItemFrameEven
 
 			if (!args.Player.HasBuildPermission(args.X, args.Y))
 			{
+				int num = Item.NewItem(new EntitySource_DebugCommand(), (args.X * 16) + 8, (args.Y * 16) + 8, args.Player.TPlayer.width, args.Player.TPlayer.height, args.ItemID, args.Stack, noBroadcast: true, args.Prefix, noGrabDelay: true);
+				Main.item[num].playerIndexTheItemIsReservedFor = args.Player.Index;
+				NetMessage.SendData((int)PacketTypes.ItemDrop, -1, -1, NetworkText.Empty, num, 1f);
+				NetMessage.SendData((int)PacketTypes.ItemOwner, -1, -1, NetworkText.Empty, num);
```
Arthri:
```suggestion
				NetMessage.SendData((int)PacketTypes.ItemDrop, args.Player.Index, -1, NetworkText.Empty, num, 1f);
				NetMessage.SendData((int)PacketTypes.ItemOwner, args.Player.Index, -1, NetworkText.Empty, num);
```
I'm suggesting a change so that the item drop is not visible to anyone else

TShockAPI/Bouncer.cs
```diff
@@ -2491,6 +2491,11 @@ internal void OnPlaceItemFrame(object sender, GetDataHandlers.PlaceItemFrameEven
 
 			if (!args.Player.HasBuildPermission(args.X, args.Y))
 			{
+				int num = Item.NewItem(new EntitySource_DebugCommand(), (args.X * 16) + 8, (args.Y * 16) + 8, args.Player.TPlayer.width, args.Player.TPlayer.height, args.ItemID, args.Stack, noBroadcast: true, args.Prefix, noGrabDelay: true);
```
Arthri:
```suggestion
				int num = Item.NewItem(null, (args.X * 16) + 8, (args.Y * 16) + 8, args.Player.TPlayer.width, args.Player.TPlayer.height, args.ItemID, args.Stack, noBroadcast: true, args.Prefix, noGrabDelay: true);
```
Better to leave item source null rather than say it's from something it's not

Arthri:
Thanks for the pull request! I have a few suggestions, if you don't mind

Arthri (changes requested):

lost-werewolf:
Why not just use the UpdateItemDrop (90) packet? In DropItemInstanced method from NPC.cs, the item slot is set to be unusable for 54000 frames after spawned, and then the item is set to inactive right after sending UpdateItemDrop. This eliminates sending a second packet afterwards and also replicates vanilla behavior for locally instanced items.
Of course, you need to send this packet to only the player who's being rejected.

bartico6:
> Why not just use the UpdateItemDrop (90) packet? In DropItemInstanced method from NPC.cs, the item slot is set to be unusable for 54000 frames after spawned, and then the item is set to inactive right after sending UpdateItemDrop. This eliminates sending a second packet afterwards and also replicates vanilla behavior for locally instanced items. Of course, you need to send this packet to only the player who's being rejected.

Repeatedly rightclicking a blocked item frame would fill up the locked out slots (54000 frames is 900 seconds or 15 minutes) and allow a malicious client to prevent item drops from functioning on the server at all by locking out all slots.

lost-werewolf:
> > Why not just use the UpdateItemDrop (90) packet? In DropItemInstanced method from NPC.cs, the item slot is set to be unusable for 54000 frames after spawned, and then the item is set to inactive right after sending UpdateItemDrop. This eliminates sending a second packet afterwards and also replicates vanilla behavior for locally instanced items. Of course, you need to send this packet to only the player who's being rejected.
> 
> Repeatedly rightclicking a blocked item frame would fill up the locked out slots (54000 frames is 900 seconds or 15 minutes) and allow a malicious client to prevent item drops from functioning on the server at all by locking out all slots.

I see, thank you for the insight.
The unusable timer could perhaps be reduced to just a few seconds or even just one second with little to no impact on behavior as it is likely the client would pick up the item as soon as it's spawned on them. Not sure if this would be the best, as you said - a malicious client could indeed spam the item frame to temporarily disable item drops, but significantly reducing the timer could remedy this if it's set to something very low (perhaps just half a second - a second).

TShockAPI/Bouncer.cs
NightKLP: good point which making it clean 👍

TShockAPI/Bouncer.cs
NightKLP: this changes a bit when a player is running and a item doesn't reach the player it will update the item which other players can pick it up but its a good improvement 👍

Arthri (approved):

hakusaro:
@Arthri @ACaiCat is this one okay to merge in your opinion?

ACaiCat:
> @Arthri @ACaiCat is this one okay to merge in your opinion?

Tested.It can work as intended.

hakusaro:
Can someone update the wiki for this?

PR #3092, opened 2025-04-29
Feat: Add Postgres database support + Refactor database handling

SakuraIsayeki:
**Added initial support for PostgreSQL database, as an alternative to MySQL or SQLite.** 
Foundational database elements have been refactored, to account for new implementation. Should actually be easier to work with, if another DB is to be be supported in the future.

Most of the initial flow has been tested, with register/login, groups, warps, being known good, regions still being tested. Bugs may arise with the new DB support, testing would be advised. 

Leaving this PR as a draft until I clear most of the bugs (which are mostly if not wholly due to casing and escaping issues), would be happy to hear some feedback in the meantime.


hakusaro (commented):
A very cursory glance from me indicates no major issues with this, but I wanna rope in @QuiCM who has more thoughts on database structure before I go too far.

SakuraIsayeki:
Thanks for the feedback. Indeed, more thoughts on the DB specialist(s) of the project would be greatly appreciated.
I was unsure of whether further evolutions on the database handling would be appropriate, and how they could/would take form. For now, decided to do the bare minimum to add support for PGSQL, all whilst keeping the structure as close to original.
I'm willing to go a few steps further, if you're okay with that. But further discussions may be needed. Hit me up on Discord if you have some time, and let's discuss it.

hakusaro:
@Pryaxis/tshock I intend to merge this without any other feedback in **one week**.

SignatureBeef:
doing a bit of a quick test, unfortunately getting some errors but not ruling out something my end.
posting the exception should it be known:
image

running via launcher with `PostgreSQL 15.8 on aarch64-unknown-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit`

aside from these that may need ironing out, the code changes are looking good to me, and personally happy to see this one pop up - nice stuff!.

TShockAPI/DB/SqlTable.cs
```diff
@@ -69,36 +72,50 @@ public bool EnsureTableStructure(SqlTable table)
...
+					break;
 				}
-			}
-			else
-			{
-				throw new NotSupportedException();
+				case SqlType.Postgres:
+				{
+					// HACK: Using "ilike" op to ignore case, due to weird case issues adapting for pgsql
+					using QueryResult reader = database.QueryReader("SELECT column_name FROM information_schema.columns WHERE table_name ILIKE @0", table.Name);
```
SignatureBeef: ah think we might need the table schema check here too, e.g. `AND table_schema=current_schema()` or similar which seems to work for me.
im currently utilising supabase's postgres which already has `auth.users`

SignatureBeef (commented):
think i have found it - partially me, but think this might be a good case to check the table schema (e.g. in code review comment)

after this, seems that other error is resolved: 🎉 

image

TShockAPI/DB/SqlTable.cs
SakuraIsayeki: Fixed in latest commit, thanks for pointing that oversight out!

TShockAPI/Configuration/TShockConfig.cs
```diff
@@ -552,6 +552,22 @@ public class TShockSettings
 		[Description("The password used when connecting to a MySQL database.")]
 		public string MySqlPassword = "";
 
+ 		///<summary>The Postgres hostname and port to direct connections to.</summary>
+ 		[Description("The Postgres hostname and port to direct connections to.")]
```
hakusaro: Can you fix the inconsistent indendation / whitespace here?

hakusaro:
@SakuraIsayeki I think you can rebase nao

SakuraIsayeki:
> @SakuraIsayeki I think you can rebase nao

Rebased as requested. However, just before we merge anything, I'll quickly bump the Npgsql package too, to align with our current .NET version.

TShockAPI/Configuration/TShockConfig.cs
```diff
@@ -552,6 +552,22 @@ public class TShockSettings
...
+	    public string PostgresHost = "";
+
+	    /// <summary>The database name to connect to when using Postgres as the database type.</summary>
+	    [Description("The database name to connect to when using Postgres as the database type.")]
+	    public string PostgresDbName = "";
+
+	    /// <summary>The username used when connecting to a Postgres database.</summary>
+	    [Description("The username used when connecting to a Postgres database.")]
+	    public string PostgresUsername = "";
```
hakusaro: The inconsistent spacing/indentation in this file will be a problem later. While my previous commit is only on one line, all changes in this file have the problem.

TShockAPI/DB/DbBuilder.cs
```diff
@@ -0,0 +1,115 @@
...
+using TShockAPI.Configuration;
+
+namespace TShockAPI.DB;
+
+/// <summary>
+/// Provides logic to build a DB connection.
+/// </summary>
+public sealed class DbBuilder
+{
+	private readonly TShock _caller;
```
hakusaro, line 17: The indentation and spacing in this file doesn't seem to match the rest of the repo but I could be wrong. Can you double check?

hakusaro (approved):

hakusaro:
@SakuraIsayeki please update the changelog on the wiki, kthx!

PR #3095, opened 2025-05-06
feat: Add spawning pets perm to default group

SakuraIsayeki:
Add permission for users to spawn pets for default usergroup.
Codebase currently warns about possible high net usage over this permission, so discussions are in order in regards to enhancements to make this possible.

### DIFF
--- a/TShockAPI/DB/GroupManager.cs
+++ b/TShockAPI/DB/GroupManager.cs
@@ -74,6 +74,7 @@ namespace TShockAPI.DB
 						Permissions.canchangepassword,
 						Permissions.canlogout,
 						Permissions.summonboss,
+						Permissions.spawnpets,
 						Permissions.worldupgrades,
 						Permissions.whisper,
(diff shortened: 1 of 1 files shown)


hakusaro:
> Codebase currently warns about possible high net usage over this permission, so discussions are in order in regards to enhancements to make this possible.

In basic terms I think the point is that out-of-the-box it should behave like vanilla, we can simply document this on the wiki as a suggestion for large servers to change the permissions for/revoke.

hakusaro (approved):

hakusaro:
@SakuraIsayeki please update the changelog on the wiki when this merges

SakuraIsayeki:
> @SakuraIsayeki please update the changelog on the wiki when this merges

Changed as requested : https://github.com/Pryaxis/TShock/wiki/Changelog/_compare/d757495a2dd004d213388373f4f2ffa509403d41

PR #3099, opened 2025-05-11
fix(OnTileEdit): item disappeared when hitting an ItemFrame without permission

ACaiCat:
**Fix item disappeared when hitting an ItemFrame without permission. (@ACaiCat)**
PixPin_2025-05-11_12-34-26


hakusaro:
I guess better than #3088 in your opinion?

ACaiCat:
> I guess better than #3088 in your opinion?

hmmm, it is another issue

hakusaro:
> hmmm, it is another issue

oh i cannot read

ACaiCat:
- #3088: if player without permission try to use a holding item to replace the item on the item frame, the holding will not return
- This PR: if player without permission try
to fetch the item on the item frame, the item will disapper

hakusaro (approved):

hakusaro:
@ACaiCat  can you update the changelog on the wiki?

PR #3106, opened 2025-05-24
Allow for reloading of player bans

Terrabade:
Allows all bans to be reloaded from the database (similarly to item bans, projectile bans, groups etc) via the `UpdateBans` method in `BanManager` (`EnsureBansCollection`, but no longer single-use), and now does so when the reload command is executed. The method has been made public. I refrained from restructuring/refactoring `BanManager` any more than absolutely necessary.

I would like to update the `/reload` response string to be more verbose on specifically what has been reloaded, but am unsure of i18n implications so have left it be.


hakusaro (approved):

hakusaro:
@Terrabade plz update changelog on wiki kthx!

PR #3107, opened 2025-05-24
Feature: Whitelist overhaul

### LINKED ISSUE #3091: Allow IP ranges in whitelist
**Is your feature request related to a problem? Please describe.**
Me and my girlfriend are trying to run a private Terraria server. Her ISP often changes her public IP. 

**Describe the solution you'd like**
Our best solution would be to use IP ranges to fix it entirely.

**Describe alternatives you've considered**
There's not a whole lot of other ways to go about this, really. We _could_ just allowlist every single IP in the IP range, but I don't think I have the patience for that, to be honest.

SakuraIsayeki:
Overhauled the whitelist system to use a new dedicated service. 
This one is memory-based to optimise lookups, rather than accessing the file at each server join for WL comparison. 

Added support for in-file comments, by using `#` prefix (considered invalid and ignored by parser). Could come in handy for those massive files, or just keeping track of "whose IP is this" cases.
Added support for ranges using CIDR notation, to allow for entire subnets to be whitelisted. Particularly useful for LAN parties. 

New default for the `whitelist.txt` file looks like this : 
```ini
# Localhost
127.0.0.1
::1

# Uncomment to allow IPs within private ranges
# 10.0.0.0/8
# 172.16.0.0/12
# 192.168.0.0/16
# fe80::/10
# fd00::/8
```

Resolves #3091.


hakusaro:
@ACaiCat if you think the code is good, can you add a code review too?

hakusaro (changes requested):
I tested this and after adding 127.0.0.1 to the whitelist and using /reload, I was not able to whitelist myself.

Also, Terraria doesn't support IPv6.

hakusaro:
CleanShot 2025-05-25 at 12 00 39

ACaiCat:
> @ACaiCat if you think the code is good, can you add a code review too?

Maybe you should find someone more professional, but it's not me.

hakusaro:
> Maybe you should find someone more professional, but it's not me.

You are professional

hakusaro:
Also, if the whitelist is blank, I think it would be fine to populate it with the example whitelist instead of leaving it blank. e.g., if whitelist is disabled, and is empty, add the default config to it.

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,270 @@
...
+
+				_file.Delete();
+				File.Move(tempFile, _file.FullName);
+			}
+			else
+			{
+				File.Delete(tempFile);
+			}
+
+            return removed;
```
hakusaro: CleanShot 2025-05-25 at 12 04 54


Indentation

TShockAPI/Commands.cs
```diff
@@ -1742,12 +1742,8 @@ void BanDetails()
 
 		private static void Whitelist(CommandArgs args)
 		{
-			if (args.Parameters.Count == 1)
+			if (args.Parameters is [{ } ip] && TShock.Whitelist.AddToWhitelist(ip))
```
sgkoishi: Perhaps show something even when not added (already existing)?

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,270 @@
...
+	/// <param name="host">string ip of the user</param>
+	/// <returns>true/false</returns>
+	public bool IsWhitelisted(string host)
+	{
+		if (!Enabled)
+		{
+			return true;
+		}
+
+		if (!IPAddress.TryParse(TShock.Utils.GetRealIP(host), out IPAddress? ip))
```
sgkoishi: The `GetRealIP` is IPv4 only and will breaks IPv6... while Terraria doesn't support v6 🤣

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,270 @@
...
+		}
+		else if (IPAddress.TryParse(content, out IPAddress? ip))
+		{
+			_whitelistAddresses.Add(ip);
+		}
+		else
+		{
+			// If we reach here, the line is not a valid IP address or network.
+			// We could throw this, but for now we just log and ignore it.
+			TShock.Log.Warn($"Invalid whitelist entry at line {line}: \"{content.ToString()}\", skipped");
```
sgkoishi: Not a big problem but a line with just whitespace will ended up with a warning

ACaiCat:
> I tested this and after adding 127.0.0.1 to the whitelist and using /reload, I was not able to whitelist myself.
> 
> Also, Terraria doesn't support IPv6.

It seems can not be reloaded through `/reload` command (Need restart whole server)

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,270 @@
...
+
+	private bool RemoveFromWhitelist(IPAddress ip)
+		=> _whitelistAddresses.Remove(ip)
+		&& RemoveLine(ip.ToString());
+
+	private bool RemoveFromWhitelist(IPNetwork network)
+		=> _whitelistNetworks.Remove(network)
+		&& RemoveLine(network.ToString());
+
+	private bool AddLine(scoped ReadOnlySpan<char> content)
```
sgkoishi, line 219: Both the `AddLine` and `RemoveLine` don't sync the in-memory list with the file, but instead add or remove from each separately. Not sure if this is intended, but sounds weird?

SakuraIsayeki: Fully intended, Add/RemoveLine is only intended to combine the logic for file ops. They're internal calls destined to be used upstream by public-facing methods 😌

TShockAPI/Whitelist.cs
SakuraIsayeki: _May one day my Rider install be finally able to stop this lunatic disorder of switching between tabs and spaces on a whim..._

hakusaro: https://www.sublimetext.com/

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,291 @@
...
+			return true;
+		}
+
+		if (!IPAddress.TryParse(host, out IPAddress? ip))
+		{
+			throw new ArgumentException($"The provided host '{host}' is not a valid IP address.", nameof(host));
+		}
+
+		// HACK: Terraria doesn't support IPv6 yet, so we can't check for it.
+		// Remove once TShock supports IPv6.
```
sgkoishi, line 96: This branch will never be true unless both underlying TShock and Terraria change - but let's just leave it without changing back and forth on these minor points...

TShockAPI/Whitelist.cs
sgkoishi: Yes, I'm referring to the public methods: for example, the whitelist in memory is empty and the file is `127.0.0.1` and `10.0.0.1`. 
Attempt to call the `public RemoveFromWhitelist("127.0.0.1")` will **not** remove `127.0.0.1` from the file (because it was not in memory), nor load `10.0.0.1` from the file. Similarly, `Add(a.b.c)` will not add it to the file if a.b.c is already in the memory.


(Due to the use of short-circuit logical AND `&&`)

sgkoishi (commented):
Other note:
* There's a submodule update that actually reverts 57039ed002, probably unwanted.
* The include of `.idea` folder?

SakuraIsayeki:
> Other note:
> 
> * There's a submodule update that actually reverts 57039ed002, probably unwanted.
> 
> * The include of `.idea` folder?

Oops, indeed unwanted changes. Lemme fix that right away 😅

TShockAPI/Whitelist.cs
hakusaro: It *was* briefly supported in a released version of Terraria, but support was removed when an unexpected bug happened. I don't know if it will be re-added in a future update. It was live for about 8 hours before the hotfix removed it.

.idea/.idea.TShock/.idea/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+# Rider ignored files
+/.idea.TShock.iml
+/modules.xml
+/contentModel.xml
+/projectSettingsUpdater.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
```
hakusaro:
```suggestion
.idea
```

hakusaro (changes requested):

TShockAPI/Whitelist.cs
SakuraIsayeki: Added this on the offchance Relogic does actually (re-)release IP6 support in a near future. Let's keep it there, if no one objects too sternly on it :)

TShockAPI/Whitelist.cs
SakuraIsayeki: True point, I actually argued the use of the in-memory list being the source of truth for validation purposes, especially with dupes or any invalid notations in-file.
Stood to reason that, if duplicate (IIRC HashSet doesn't allow for duplicate values), adding to the list would fail, ergo shouldn't write to file. Similarly, pointless attempting to remove from file some value that isn't there or wasn't parsed.

Either I've gone crazy (that's what I get for missing my morning coffee...), or perhaps my logic wasn't very clear. Does this shed a bit more light as to why the && shorting ? Would you agree with this point, or would you like me to go with & operators instead to force the file ops in case of failures?

sgkoishi: It works great when the memory and the file were syncing, but unintuitive when they are already **desynced** - might be kind of edge case:
* Run the server with default config
* Cleared the file content of whitelist.txt
* `/whitelist 127.0.0.1`
* `/reload` or restart the server
People would probably expect 127.0.0.1 to be whitelisted, but actually no it wasn't

SakuraIsayeki: At this point, it does beg the question of file-watching for changes to sync the whitelist... 

I'll give it a shot.

sgkoishi: Really edge case and I'm ok if we just keep the current behaviour - just a note in case someone encountered

hakusaro: Just my 2c but this is why the original implementation, written by yours truly, is basic and just uses the file as a source of truth rather than syncing data structures.

SakuraIsayeki: Aight, your call. I'm okay going the extra step and implementing a watcher, but would kinda defeat the purpose of a reload, I feel. 
Still, very thorough work, thanks for pointing out the edge case.

hakusaro: I mean you don't have to implement a watcher -- you could just synchronize the data and write out the file if someone adds a new player to the list.

SakuraIsayeki: Fair enough, let's just KISS it, I'll change the short-circuit operators.

TShockAPI/Commands.cs
```diff
@@ -1742,13 +1742,20 @@ void BanDetails()
...
+				else
 				{
-					tw.WriteLine(args.Parameters[0]);
+					args.Player.SendErrorMessage(GetString($"Failed to add {ip} to the whitelist. Perhaps it is already whitelisted?"));
 				}
-				args.Player.SendSuccessMessage(GetString($"Added {args.Parameters[0]} to the whitelist."));
+			}
+			else
+			{
+				args.Player.SendErrorMessage(GetString($"Invalid Whitelist syntax. Refer to {Specifier}whitelist help for details on how to use the {Specifier}whitelist command"));
```
ACaiCat: It seems that the `/whitelist help` subcommand doesn't exist.

ACaiCat:
I test it on my win server, `/whitelist` command can't work.I am sure that no other process was using `whitelist.txt`
### Log:
```
2025-05-30 21:28:08 - Utils: INFO: Server execute: /whitelist 221.12.2.1。
2025-05-30 21:28:08 - Command: ERROR: System.IO.IOException: The process cannot access the file 'C:\Users\Administrator\Desktop\a\tshock\whitelist.txt' because it is being used by another process.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options) in System.Private.CoreLib.dll:token 0x60001ce+0xa7
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode) in System.Private.CoreLib.dll:token 0x60001cd+0x6
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode) in System.Private.CoreLib.dll:token 0x6008cfe+0x14
   at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode) in System.Private.CoreLib.dll:token 0x6008cd5+0xa
   at System.IO.FileInfo.OpenRead() in System.Private.CoreLib.dll:token 0x6008742+0x0
   at TShockAPI.Whitelist.AddLine(ReadOnlySpan`1 content) in /home/runner/work/TShock/TShock/TShockAPI/Whitelist.cs:line 227
   at TShockAPI.Whitelist.AddToWhitelist(IPAddress ip) in /home/runner/work/TShock/TShock/TShockAPI/Whitelist.cs:line 203
   at TShockAPI.Whitelist.AddToWhitelist(ReadOnlySpan`1 ip) in /home/runner/work/TShock/TShock/TShockAPI/Whitelist.cs:line 175
   at TShockAPI.Commands.Whitelist(CommandArgs args) in /home/runner/work/TShock/TShock/TShockAPI/Commands.cs:line 1747
   at TShockAPI.Command.Run(CommandArgs args) in /home/runner/work/TShock/TShock/TShockAPI/Commands.cs:line 158
```

SakuraIsayeki:
> I test it on my win server, `/whitelist` command can't work.I am sure that no other process was using `whitelist.txt`

By any chance, would you have the repro steps for this one ? Because I can't repro the issue.

ACaiCat:
> By any chance, would you have the repro steps for this one ? Because I can't repro the issue.

Just try to add a ip through `/whitelist 192.168.1.2`

ACaiCat:
@THEXN tested it too and got the same error.
c34ee5cc3b502444c2953575df78f4d8

SakuraIsayeki:
Okay. Seems I derped handling the file streams. 
Now fixed, give it a go on last commit :)

ACaiCat:
It works now.
image

TShockAPI/Whitelist.cs
```diff
@@ -0,0 +1,291 @@
...
+		}
+		else if (IPAddress.TryParse(content, out IPAddress? ip))
+		{
+			_whitelistAddresses.Add(ip);
+		}
+		else
+		{
+			// If we reach here, the line is not a valid IP address or network.
+			// We could throw this, but for now we just log and ignore it.
+			TShock.Log.Warn($"Invalid whitelist entry at line {line}: \"{content.ToString()}\", skipped");
```
ACaiCat: Adding GetString might be better
```suggestion
			TShock.Log.Warn(GetString($"Invalid whitelist entry at line {line}: \"{content.ToString()}\", skipped"));
```

ACaiCat (approved):

hakusaro (approved):

PR #3116, opened 2025-06-15
fix(Bouncer): fix players bypass region protection and build permissions when using Quick Stack

ACaiCat:
**Fixed players bypass region protection and build permissions when using Quick Stack**
## Problem
- Players who aren't logged in can use Quick Stack to put items into chests.
- Many players report stacking items into protected regions when using Quick Stack.
image

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -137,7 +137,7 @@ namespace TShockAPI
 			GetDataHandlers.KillMe += OnKillMe;
 			GetDataHandlers.FishOutNPC += OnFishOutNPC;
 			GetDataHandlers.FoodPlatterTryPlacing += OnFoodPlatterTryPlacing;
-			OTAPI.Hooks.Chest.QuickStack += OnChestOnQuickStack;
+			OTAPI.Hooks.Chest.QuickStack += OnQuickStack;
 
 
 			// The following section is based off Player.PlaceThing_Tiles_PlaceIt and Player.PlaceThing_Tiles_PlaceIt_GetLegacyTileStyle.
@@ -2877,7 +2877,7 @@ namespace TShockAPI
 		/// </summary>
 		/// <param name="sender"></param>
 		/// <param name="args"></param>
-		internal void OnChestOnQuickStack(object sender, OTAPI.Hooks.Chest.QuickStackEventArgs args)
+		internal void OnQuickStack(object sender, OTAPI.Hooks.Chest.QuickStackEventArgs args)
 		{
 			var id = args.ChestIndex;
 			var plr = TShock.Players[args.PlayerId];
@@ -2898,13 +2898,6 @@ namespace TShockAPI
 			if (!plr.HasBuildPermission(Main.chest[id].x, Main.chest[id].y) && TShock.Config.Settings.RegionProtectChests)
 			{
 				TShock.Log.ConsoleDebug(GetString("Bouncer / OnQuickStack rejected from region protection? from {0}", plr.Name));
-				args.Result = HookResult.Cancel;
-				return;
-			}
-
-			if (!plr.IsInRange(Main.chest[id].x, Main.chest[id].y, 600/16))
-			{
-				TShock.Log.ConsoleDebug(GetString("Bouncer / OnQuickStack rejected from range check from {0}", plr.Name));
 				args.Result = HookResult.Cancel;
 				return;
 			}
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -19,7 +19,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Terraria.ID;
-using TShockAPI.Net;
 using Terraria;
 using Microsoft.Xna.Framework;
 using TShockAPI.Localization;
@@ -29,6 +28,7 @@ using Terraria.DataStructures;
 using Terraria.Localization;
 using TShockAPI.Models.PlayerUpdate;
 using System.Threading.Tasks;
+using OTAPI;
 using Terraria.GameContent.Tile_Entities;
 
 namespace TShockAPI
@@ -137,6 +137,7 @@ namespace TShockAPI
 			GetDataHandlers.KillMe += OnKillMe;
 			GetDataHandlers.FishOutNPC += OnFishOutNPC;
 			GetDataHandlers.FoodPlatterTryPlacing += OnFoodPlatterTryPlacing;
+			OTAPI.Hooks.Chest.QuickStack += OnChestOnQuickStack;
 
 
 			// The following section is based off Player.PlaceThing_Tiles_PlaceIt and Player.PlaceThing_Tiles_PlaceIt_GetLegacyTileStyle.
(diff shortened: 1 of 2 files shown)


hakusaro (commented):
It looks like you setup a new TShock event for `ForceItemIntoNearestChest` in 939d1582f5506f014a38510e1809309abd97cd5e but then in 6c500dfa70ebba1a0fdcda12deabb4ca5b175d12 you don't ever fire the event?

TShockAPI/Bouncer.cs
```diff
@@ -2871,6 +2872,44 @@ internal void OnFoodPlatterTryPlacing(object sender, GetDataHandlers.FoodPlatter
...
+			if (!plr.HasBuildPermission(Main.chest[id].x, Main.chest[id].y) && TShock.Config.Settings.RegionProtectChests)
+			{
+				TShock.Log.ConsoleDebug(GetString("Bouncer / OnQuickStack rejected from region protection? from {0}", plr.Name));
+				args.Result = HookResult.Cancel;
+				return;
+			}
+
+			if (!plr.IsInRange(Main.chest[id].x, Main.chest[id].y, 600/16))
+			{
+				TShock.Log.ConsoleDebug(GetString("Bouncer / OnQuickStack rejected from range check from {0}", plr.Name));
```
bartico6: It appears the items are not rolled back here - if we roll back when the player is being disabled and for not having build priv, I figure range desync should also try to restore the items to the user - unless this is standard and I'm unaware of it

bartico6 (commented):
Undecided, would like feedback from PR author and other maintainers about the comment I left in bouncer code

sgkoishi:
> It looks like you setup a new TShock event for `ForceItemIntoNearestChest` in [939d158](https://github.com/Pryaxis/TShock/commit/939d1582f5506f014a38510e1809309abd97cd5e) but then in [6c500df](https://github.com/Pryaxis/TShock/commit/6c500dfa70ebba1a0fdcda12deabb4ca5b175d12) you don't ever fire the event?

The `ForceItemIntoNearestChestEventArgs` was added to the handler list so should be fired whenever `PacketTypes.ForceItemIntoNearestChest` received...?

~~Would be better if we could move items to other accessible chests~~

TShockAPI/Bouncer.cs
lost-werewolf: IIRC, quick-stacking to nearby chests is your client requesting the server to place the items in a chest - the logic runs server-side after the client requests it, in which the client just flags items in their inventory to be modifiable by the server before sending the request. The item is only removed once the server informs the client to remove it.

If you simply tell the client "no" without doing anything, I _believe_ it results in an item that the client can't modify without reconnecting.

Proper approach to rejecting it would simply be to send their item slot back without any modifications to it, so the item doesn't get stuck. The client will un-flag the item as being quick-stacked if it receives the item slot packet for that item, and it won't be stuck anymore (if it even does get stuck).

sgkoishi: Related logic done in OTAPI, this hook might be fired multiple times for the same inventory against different chests. No rollback here because OTAPI will find other possible chests and eventually send the result back.

ACaiCat: `OTAPI.Hooks.Chest.QuickStack` acts as an extra filtering condition for chests in `Chest.PutItemInNearbyChest` and does not block `ServerPlaceItem`. After `Chest.PutItemInNearbyChest`, item will roll back through vanilla logic.
```csharp
public static void ServerPlaceItem(int plr, int slot)
  {
    if (slot >= PlayerItemSlotID.Bank4_0 && slot < PlayerItemSlotID.Bank4_0 + 40)
    {
      int index = slot - PlayerItemSlotID.Bank4_0;
      Main.player[plr].bank4.item[index] = Chest.PutItemInNearbyChest(Main.player[plr].bank4.item[index], Main.player[plr].Center, plr);
      NetMessage.SendData(5, number: plr, number2: (float) slot, number3: (float) Main.player[plr].bank4.item[index].prefix);
    }
    else
    {
      if (slot >= 58)
        return;
      Main.player[plr].inventory[slot] = Chest.PutItemInNearbyChest(Main.player[plr].inventory[slot], Main.player[plr].Center, plr);
      NetMessage.SendData(5, number: plr, number2: (float) slot, number3: (float) Main.player[plr].inventory[slot].prefix);
    }
  }

public static Item PutItemInNearbyChest(Item itemToDeposit, Vector2 playerPosition, int playerId)
{
    if (Main.netMode == NetmodeID.Client)
        return itemToDeposit;

    bool performActualTransfer = true;
    
    for (int chestIndex = 0; chestIndex < Main.maxChests; chestIndex++)
    {
        bool hasMatchingItem = false;
        bool hasEmptySlot = false;
        
        Chest currentChest = Main.chest[chestIndex];
        
        if (currentChest == null || Chest.IsPlayerInChest(chestIndex) || Chest.IsLocked(currentChest.x, currentChest.y))
            continue;

        Vector2 chestCenterPosition = new Vector2(
            currentChest.x * Tile.Size + Tile.Size / 2, 
            currentChest.y * Tile.Size + Tile.Size / 2
        );
        float distanceToChest = (chestCenterPosition - playerPosition).Length();

        //QuickStackHook fired here
        if (distanceToChest < 600f && OTAPI.Hooks.Chest.InvokeQuickStack(playerId, itemToDeposit, chestIndex)) 
        {
            for (int slotIndex = 0; slotIndex < currentChest.item.Length; slotIndex++)
            {
                Item chestItem = currentChest.item[slotIndex];
                
                if (chestItem.IsAir)
                {
                    hasEmptySlot = true;
                }
                else if (itemToDeposit.IsTheSameAs(chestItem))
                {
                    hasMatchingItem = true;
                    
                    int availableSpace = chestItem.maxStack - chestItem.stack;
                    if (availableSpace > 0)
                    {
                        int transferAmount = Math.Min(availableSpace, itemToDeposit.stack);
                        
                        Chest.VisualizeChestTransfer(playerPosition, chestCenterPosition, itemToDeposit, transferAmount);
                        
                        if (performActualTransfer)
                        {
                            itemToDeposit.stack -= transferAmount;
                            chestItem.stack += transferAmount;
                        }
                        
                        if (itemToDeposit.stack <= 0)
                        {
                            itemToDeposit.SetDefaults();
                            return itemToDeposit;
                        }
                    }
                }
                else
                {
                    hasEmptySlot = true;
                }
            }
            
            if (hasMatchingItem && hasEmptySlot && itemToDeposit.stack > 0)
            {
                for (int slotIndex = 0; slotIndex < currentChest.item.Length; slotIndex++)
                {
                    if (currentChest.item[slotIndex].IsAir)
                    {
                        Chest.VisualizeChestTransfer(playerPosition, chestCenterPosition, itemToDeposit, itemToDeposit.stack);
                        
                        if (performActualTransfer)
                        {
                            currentChest.item[slotIndex] = itemToDeposit.Clone();
                            itemToDeposit.SetDefaults();
                        }
                        return itemToDeposit;
                    }
                }
            }
        }
    }
    
    return itemToDeposit;
}
```

hakusaro:
> The ForceItemIntoNearestChestEventArgs was added to the handler list so should be fired whenever PacketTypes.ForceItemIntoNearestChest received...?

I am awake

QuiCM (approved):

PR #3146, opened 2025-09-26
fix: allow evil grass to grow during world generation even when `Allow...Creep` is `false`

ACaiCat:
### Normal
img

### `AllowCrimsonCreep` or `AllowCorruptionCreep` is `false`
img

### Changelog
**Fixed evil grass growth blocked during world generation when `Allow...Creep` is `false`**


bartico6 (approved):
👍

QuiCM (approved):

PR #3150, opened 2026-01-17
fix: initial ssc spawn always (-1,-1) for partial player

ksqeib:
always found this:
`TShock: VERBOSE: OnSecondUpdate / initial ssc spawn for white01 at (-1, -1)`

so I fixed this by copy TeleportSpawnpoint code to HandleSpawn,this make beahavior keep same.


hakusaro (approved):

Zaicon (approved):

AniAnanas (approved):

TShockAPI/GetDataHandlers.cs
```diff
@@ -2753,10 +2753,19 @@ private static bool HandleSpawn(GetDataHandlerArgs args)
 				// handle initial connection
 				if (args.Player.State == 3)
 				{
+					//same as TeleportSpawnpoint
```
ZakFahey:
```suggestion
					// Same as TSPlayer.TeleportSpawnpoint
```

TShockAPI/GetDataHandlers.cs
```diff
@@ -2753,10 +2753,19 @@ private static bool HandleSpawn(GetDataHandlerArgs args)
...
+					int x = args.TPlayer.SpawnX;
+					int y = args.TPlayer.SpawnY;
+					if ((x == -1 && y == -1) ||
```
ZakFahey: Instead of copy/pasting this code, extract both to a shared util?

ksqeib: That's a great idea, but since I'm not familiar with the TShock structure, I'm not sure where to place it. Could you modify it directly?

> Instead of copy/pasting this code, extract both to a shared util?

hakusaro: You could just make a method in this file

ksqeib: > You could just make a method in this file

Thank you for your guidance. I have already submitted the new commit, and I hope there are no misunderstandings.

PR #3161, opened 2026-01-28
1.4.5: Fix PlayerInfo Packet Format, solve "Kicked: You have been Bounced."

Reisenbug:
This PR addresses the changes introduced in Terraria 1.4.5 to the PlayerInfo packet (PacketType 4). The following updates have been made:

New Fields Added:  
   - Added reading of `voiceVariant` (byte) and `voicePitchOffset` (float) to handle the new voice-related data.
   
`hideVisual` Format Change:  
   - Updated the reading of `hideVisual`, which has changed from two `BitsByte` values to a single `ushort` in 1.4.5. The logic has been adjusted to account for this new structure.
   
Updated `hideVisibleAccessory` Logic:  
   - Reworked how `hideVisibleAccessory` is populated, using the new `ushort` format for `hideVisualFlags` and ensuring correct flag assignments to the array.

Compatibility Fixes:  
   - The changes were made to maintain compatibility with the updated data structure and ensure proper reading and assignment of player info in the new Terraria version.
   - Avoiding player been kicked by Bouncer mistakenly.(mainly because of voice field ig)
   
488439db84a4cc319b8a8a65c38dcdc5
Now player can join the server normally.


ServerLog.txt
sgkoishi, line 1: I would assume you don't want your server logs and configuration files be added to TShock repository...

TShockAPI/GetDataHandlers.cs
```diff
@@ -2511,12 +2511,15 @@ private static bool HandlePlayerInfo(GetDataHandlerArgs args)
...
+			// 1.4.5: new voice fields
+			byte voiceVariant = args.Data.ReadInt8();
+			float voicePitchOffset = args.Data.ReadSingle();
 			var hair = args.Data.ReadInt8();
 			string name = args.Data.ReadString();
 			byte hairDye = args.Data.ReadInt8();
 
-			BitsByte hideVisual = args.Data.ReadInt8();
-			BitsByte hideVisual2 = args.Data.ReadInt8();
+			// 1.4.5: hideVisibleAccessory is now a ushort instead of two BitsByte
```
sgkoishi: Is there any difference between one ushort and two bytes? They are identical and unchanged.

sgkoishi: With bisect this non-functional change can be found in the first version of 1.4.4.9 (manifest 6735719959859348558).

grosscol: It appears as though this change is about consolidating `hideVisual` and `hideVisual2` into a single variable.  Same 16 bits, but no longer bifurcated so it matches the length of  `args.Player.TPlayer.hideVisibleAccessory`.  Ostensibly this is easier to reason about and maintain.

I think `(hideVisualFlags & (1 << i))` is a clever little idiom for converting a bit to a boolean.

ServerLog.txt
Reisenbug: Already deleted

TShockAPI/GetDataHandlers.cs
Reisenbug: So whether should I withdraw this change or not

sgkoishi: It's fine to keep this. However, you might want to revert dfb61524 which you are using your fork of TSAPI.

Reisenbug: Revert https://github.com/Pryaxis/TShock/commit/dfb6152475cd663f14216a047e37aaa30ce39361 done.

ZakFahey: Superfluous comments. We don't need to specify that certain fields are 1.4.5 or what the old behavior used to be.

Reisenbug: removed already

QuiCM:
I think #3166 superseded all the changes included here, but I'll make sure that this contribution is noted however as it was here first :)

PR #3163, opened 2026-01-29
Update OTAPI and .net version

ksqeib:
Update the versions of OTAPI and .NET as indicated in the title.
This may not be necessary.


hakusaro:
We're not updating .net until after updating otapi.

PR #3164, opened 2026-01-30
Fix SSC (Server Side Characters) for Terraria 1.4.5

Reisenbug:
ssc mode broken in 1.4.5
`/uploadssc` throwing `IndexOutOfRangeException` in log
ssc failed. Players can join an ssc-enabled server without being reset, but when they click on items in their inventory, those items turn into whatever ssc already has in that slot. Also, the items can't actually be picked up.
etc.

Root Cause:
1.4.5 source code:
image
1.4.4.9 source code:
image
They allocate 200 slots per bank instead of 40.

Changes:
Use `PlayerItemSlotID` constants instead of sequential slot counter in `RestoreCharacter`
Add try-finally to ensure `IgnoreSSCPackets` is reset even if exception occurs
Update `HandlePlayerSlot` slot validation to use `PlayerItemSlotID.Count` (990) instead of `NetItem.MaxInventory` (350)

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2677,7 +2677,7 @@ namespace TShockAPI
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Loadout3_Dye_0 + NetItem.LoadoutDyeSlots - 1) // equals to 989, avoid hardcoding
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Count - 1) // equals to 989, avoid hardcoding
 			{
 				args.Player.HasSentInventory = true;
 				bypassTrashCanCheck = true;
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2676,8 +2676,9 @@ namespace TShockAPI
 			bool blockedSlot = slotFlags[1];
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
+			// The last slot the client sends is Count - 1 (Count is total slots, so Count - 1 is the last index)
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Count - 1) // equals to 989, avoid hardcoding
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Count - 1)
 			{
 				args.Player.HasSentInventory = true;
 				bypassTrashCanCheck = true;
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2676,15 +2676,16 @@ namespace TShockAPI
 			bool blockedSlot = slotFlags[1];
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
+			// 1.4.5: The last actual item slot is Loadout3_Dye_0 + LoadoutDyeSlots - 1
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == NetItem.MaxInventory)
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Loadout3_Dye_0 + NetItem.LoadoutDyeSlots - 1)
 			{
 				args.Player.HasSentInventory = true;
 				bypassTrashCanCheck = true;
 			}
 
 			if (OnPlayerSlot(args.Player, args.Data, plr, slot, stack, prefix, type, favorited, blockedSlot)
-				|| plr != args.Player.Index || slot < 0 || slot > NetItem.MaxInventory)
+				|| plr != args.Player.Index || slot < 0 || slot >= PlayerItemSlotID.Count)
 				return true;
 			if (args.Player.IgnoreSSCPackets)
 			{
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2676,9 +2676,8 @@ namespace TShockAPI
 			bool blockedSlot = slotFlags[1];
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
-			// 1.4.5: The last actual item slot is Loadout3_Dye_0 + LoadoutDyeSlots - 1
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Loadout3_Dye_0 + NetItem.LoadoutDyeSlots - 1)
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Loadout3_Dye_0 + NetItem.LoadoutDyeSlots - 1) // equals to 989, avoid hardcoding
 			{
 				args.Player.HasSentInventory = true;
 				bypassTrashCanCheck = true;
@@ -2701,7 +2700,11 @@ namespace TShockAPI
 
 			if (args.Player.IsLoggedIn)
 			{
-				args.Player.PlayerData.StoreSlot(slot, type, prefix, stack, favorited);
+				int internalSlot = NetworkSlotToInternalSlot(slot);
+				if (internalSlot >= 0)
+				{
+					args.Player.PlayerData.StoreSlot(internalSlot, type, prefix, stack, favorited);
+				}
 			}
 			else if (Main.ServerSideCharacter && TShock.Config.Settings.DisableLoginBeforeJoin && !bypassTrashCanCheck &&
 					 args.Player.HasSentInventory && !args.Player.HasPermission(Permissions.bypassssc))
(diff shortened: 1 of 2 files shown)


Reisenbug:
conflicts solved

TShockAPI/PlayerData.cs
```diff
@@ -539,165 +541,135 @@ public void RestoreCharacter(TSPlayer player)
...
 			{
-				NetMessage.SendData(5, -1, -1, NetworkText.FromLiteral(Main.player[player.Index].armor[k].Name), player.Index, slot, (float)Main.player[player.Index].armor[k].prefix);
-				slot++;
+				NetMessage.SendData(5, -1, -1, NetworkText.FromLiteral(Main.player[player.Index].armor[k].Name), player.Index, PlayerItemSlotID.Armor0 + k, (float)Main.player[player.Index].armor[k].prefix);
 			}
 			for (int k = 0; k < NetItem.DyeSlots; k++)
 			{
-				NetMessage.SendData(5, -1, -1, NetworkText.FromLiteral(Main.player[player.Index].dye[k].Name), player.Index, slot, (float)Main.player[player.Index].dye[k].prefix);
-				slot++;
+				NetMessage.SendData(5, -1, -1, NetworkText.FromLiteral(Main.player[player.Index].dye[k].Name), player.Index, PlayerItemSlotID.Dye0 + k, (float)Main.player[player.Index].dye[k].prefix);
```
ACaiCat:
```suggestion
				NetMessage.SendData(5, -1, -1, NetworkText.Empty, player.Index, PlayerItemSlotID.Dye0 + k);
```
The `SendData` parameters for the `SyncSlot (5)` packet now only include the player index and the slot ID.

ACaiCat (changes requested):

TShockAPI/GetDataHandlers.cs
```diff
@@ -2677,14 +2677,14 @@ private static bool HandlePlayerSlot(GetDataHandlerArgs args)
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == NetItem.MaxInventory)
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Loadout3_Dye_0 + NetItem.LoadoutDyeSlots - 1) // equals to 989, avoid hardcoding
```
ACaiCat: `PlayerItemSlotID.Count - 1` may be better

Reisenbug: yeah clearer

TShockAPI/PlayerData.cs
Reisenbug: Thanks for correction

ACaiCat: Hmmm, there seems to be more than one SendData.

Reisenbug: Classic me lmao
Now should be fine

ACaiCat: 要不加个群 816771079 👀

Reisenbug: 彳亍

TShockAPI/GetDataHandlers.cs
Reisenbug, line 1: These changes successfully solved at least the issue that players lose all of their items in Void Vault and Defender's Forge

TShockAPI/PlayerData.cs
Reisenbug, line 1: `try` and `finally` ensure `player.IgnoreSSCPackets` is set back to `false`, because i haven't figure out how https://github.com/Pryaxis/TShock/blob/01e742305ea6e47871e79bfb87a2266f933a3b68/TShockAPI/GetDataHandlers.cs#L3080
(HandleItemOwner: `id == 400 && owner == 255`) is supposed to work
at least it lets players pick up item without their inventory being reset by a stuck `IgnoreSSCPackets = true`.
The sendData lines use `PlayerItemSlotID` constants instead of a counter to fix the `/uploadssc` IndexOutOfRangeException

ACaiCat: The server sends an `ItemOwner` packet when SSC restoration is complete. Upon receiving this packet, the client sends back an `ItemOwner` packet with `id == 400 and owner == 255`. However, this no longer works because ReLogic added a validation check on the `ItemOwner` packet in 1.4.5

Reisenbug: I see.
I think it is harmless

TShockAPI/GetDataHandlers.cs
```diff
@@ -2677,14 +2677,14 @@ private static bool HandlePlayerSlot(GetDataHandlerArgs args)
 
 			// Players send a slot update packet for each inventory slot right after they've joined.
 			bool bypassTrashCanCheck = false;
-			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == NetItem.MaxInventory)
+			if (plr == args.Player.Index && !args.Player.HasSentInventory && slot == PlayerItemSlotID.Count - 1) // equals to 989, avoid hardcoding
```
hakusaro: Can we derive this somehow in a way that makes it clear what -1 is?

TShockAPI/PlayerData.cs
```diff
@@ -539,165 +541,133 @@ public void RestoreCharacter(TSPlayer player)
 			NetMessage.SendData((int)PacketTypes.SyncLoadout, remoteClient: player.Index, number: player.Index, number2: player.TPlayer.CurrentLoadoutIndex);
 			NetMessage.SendData((int)PacketTypes.SyncLoadout, ignoreClient: player.Index, number: player.Index, number2: player.TPlayer.CurrentLoadoutIndex);
 
-			float slot = 0f;
 			for (int k = 0; k < NetItem.InventorySlots; k++)
 			{
-				NetMessage.SendData(5, -1, -1, NetworkText.FromLiteral(Main.player[player.Index].inventory[k].Name), player.Index, slot, (float)Main.player[player.Index].inventory[k].prefix);
-				slot++;
+				NetMessage.SendData(5, -1, -1, NetworkText.Empty, player.Index, PlayerItemSlotID.Inventory0 + k);
```
hakusaro, line 546: Can you explain why this is changing?

TShockAPI/PlayerData.cs
```diff
@@ -743,6 +713,12 @@ public void RestoreCharacter(TSPlayer player)
 					*/
 				}
 			}
+			}
+			finally
+			{
+				// Ensure IgnoreSSCPackets is reset even if an exception occurs
```
hakusaro: Is this really the failure scenario we want?

hakusaro (changes requested):

TShockAPI/GetDataHandlers.cs
Reisenbug: PlayerItemSlotID.Count (990) is the total number of slot IDs, slots range from 0 to 989, so Count - 1 is the last valid slot index. Would a descriptive comment or a local variable like const int lastSyncSlot = PlayerItemSlotID.Count - 1 be clearer

TShockAPI/PlayerData.cs
Reisenbug: Old code use slot counter, but in 1.4.5, bank slots expanded from 40 to 200 each. The sequential counter produces wrong slot IDs. New code use `PlayerItemSlotID` constants to address each section's correct starting offset.

### 1.4.4.9
```csharp
private static void TrySendingItemArray(int plr, Item[] array, int slotStartIndex)
{
    for (int i = 0; i < array.Length; i++)
    {
        NetMessage.TrySendData(5, -1, -1, null, plr, slotStartIndex + i, (int)array[i].prefix);
    }
}
```

### 1.4.5
```csharp
private static void TrySendingItemArray(int plr, Item[] array, int slotStartIndex)
{
    for (int i = 0; i < array.Length; i++)
    {
        NetMessage.TrySendData(5, -1, -1, null, plr, slotStartIndex + i);
    }
}
```
The forth parameters is null
`prefix` is not necessary now

TShockAPI/PlayerData.cs
Reisenbug: You are right. Added a catch block.

bartico6:
random thoughts screamed into the ether on the ssc changes in 1.4.5.0+ (not related to _this PR_ per se, just something to note for future reference if anyone wants to make changes)

- packet 5 no longer requires an `Item` instance to send, so all senditem methods (or any SendData(5, ...) calls for that matter) could be collapsed into array-less overloads like so:
    - SendItem(int player, int index);
    - SendItems(int player, int index, int count); => count * SendItem(player, index + i);

- be careful around personal banks. you want to use the offsets from PlayerItemSlotID (because they're correct) but use array lengths for the bounds of each bank (because PlayerItemSlotID is incorrect) - once the 40-200 discrepancy is resolved, PlayerItemSlotID is Terraria's internal version of what you currently have in NetItem, and the one in NetItem should be deprecated/removed.

- it could be worthwhile to import these values from PlayerItemSlotID at runtime instead of hardcoding them, as otherwise you'll be returning to this code and reasoning about it every update - though this is guaranteed to temporarily happen anyway due to the bad indexes issue with personal banks

- the massive loop sending everything from the server in RestorePlayer could be simplified: instead of manually deciding which slots to broadcast and which slots to send only to the player being restored, you can use `PlayerItemSlotID.CanRelay[int]` with the global index of that item slot. if it's false, send to owner only, if it's true, send to everyone including owner. You could also make it into one loop going 0-990 since you're using NetMessage.SendData() which will automatically convert the index back to the correct Item array anyway


overall fix looks okay from experience making similar changes elsewhere but I cannot personally test tshock's version anytime soon. 👍

TShockAPI/GetDataHandlers.cs
hakusaro: Either, yes

hakusaro (commented):
Most likely this can be merged if you wanna make a small comment or change on the magic number. Otherwise I’m reasonably satisfied!

hakusaro (approved):
seems legit on paper

hakusaro:
@ACaiCat can you re-review and approve if you're good?

ACaiCat (approved):
Looks good to me

ACaiCat:
image all slots are tracked normally

hakusaro:
@Nitori-Lab if you can review #3164 I would really appreciate it, but if you can't, nbd!

PR #3167, opened 2026-01-31
Draft: Add Terraria 1.4.5 support

hcross:
- Update OTAPI.Upcoming to 3.3.4 (supports Terraria 1.4.5.3)
- Rewrite CreativeUnlocksHandler to use NetCreativeUnlocksPlayerReportModule instead of the removed NetCreativeUnlocksModule
- Fix SSC research sync by explicitly sending updates to all players including the sender
- Add anti-cheat validation for player ID in research packets
- Fix PlayerData research sync for connecting players

The old NetCreativeUnlocksModule was removed in Terraria 1.4.5.0. The new NetCreativeUnlocksPlayerReportModule has different semantics but this implementation works around its vanilla limitations.

### DIFF
--- a/TShockAPI/Handlers/NetModules/CreativeUnlocksHandler.cs
+++ b/TShockAPI/Handlers/NetModules/CreativeUnlocksHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.IO.Streams;
 using Terraria;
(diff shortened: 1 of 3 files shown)


TShockAPI/Handlers/NetModules/CreativeUnlocksHandler.cs
```diff
@@ -9,7 +9,8 @@
 namespace TShockAPI.Handlers.NetModules
 {
 	/// <summary>
-	/// Handles creative unlock requests
+	/// Handles creative unlock requests (Journey mode research)
+	/// Updated for Terraria 1.4.5 compatibility
```
ZakFahey, line 13: Second line is superfluous (+ other references to it being a 1.4.5 change)

ZakFahey: Overall clean up the comments and remove stuff that's not needed.

PR #3173, opened 2026-02-06
Update  bouncer for 1.4.5 (basic)

ACaiCat:
**Update bouncer for 1.4.5 (basic)**

Thanks to @xuyuwtu @Cjx8848 @1242509682 ([WindFrost-CSFT/TShock](https://github.com/WindFrost-CSFT/TShock)).

> [!NOTE]
> **Test Needed**

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2462,16 +2462,6 @@ namespace TShockAPI
 					return;
 				}
 			}
-			else if (type == TileID.GardenGnome)
-			{
-				if (style is > 4 or < 0)
-				{
-					TShock.Log.ConsoleDebug(GetString("Bouncer / OnPlaceObject rejected {0} due to invalid garden gnome style {1}", args.Player.Name, style));
-					args.Player.SendTileSquareCentered(x, y, 4);
-					args.Handled = true;
-					return;
-				}
-			}
 			else if (type == TileID.KiteAnchor)
 			{
 				if (style != 0)
@@ -2492,30 +2482,36 @@ namespace TShockAPI
 					return;
 				}
 			}
-			else if (type == TileID.Books)
-			{
-				if (style is > 4 or < 0)
-				{
-					TShock.Log.ConsoleDebug(GetString("Bouncer / OnPlaceObject rejected {0} due to invalid book style {1}", args.Player.Name, style));
-					args.Player.SendTileSquareCentered(x, y, 4);
-					args.Handled = true;
-					return;
-				}
-			}
 			else
 			{
+				List<int> allowTypes = [args.Player.TPlayer.inventory[args.Player.TPlayer.selectedItem].createTile];
+				List<int> allowStyles = [args.Player.TPlayer.inventory[args.Player.TPlayer.selectedItem].placeStyle];
+				var flexibleTileWand = args.Player.SelectedItem.GetFlexibleTileWand();
+				if (flexibleTileWand != null)
+				{
+					var flexibleTypes = flexibleTileWand._options
+						.SelectMany(kvp => kvp.Value.Options)
+						.Select(option => option.TileIdToPlace);
+
+					allowTypes.AddRange(flexibleTypes);
+
+					var flexibleStyles = flexibleTileWand._options
+						.SelectMany(kvp => kvp.Value.Options)
+						.Select(option => option.TileStyleToPlace);
+
+					allowStyles.AddRange(flexibleStyles);
+				}
 				// This is necessary to check in order to prevent special tiles such as
 				// queen bee larva, paintings etc that use this packet from being placed
 				// without selecting the right item.
-				if (type != args.Player.TPlayer.inventory[args.Player.TPlayer.selectedItem].createTile)
+				if (!allowTypes.Contains(type))
 				{
-					TShock.Log.ConsoleDebug(GetString("Bouncer / OnPlaceObject rejected awkward tile creation/selection from {0}", args.Player.Name));
+					TShock.Log.ConsoleError(GetString("Bouncer / OnPlaceObject rejected object placement with invalid tile type {1} (expected {2}) from {0}", args.Player.Name, type, string.Join(',', allowTypes)));
 					args.Player.SendTileSquareCentered(x, y, 4);
 					args.Handled = true;
 					return;
 				}
-
-				if (args.Player.SelectedItem.placeStyle != style)
+				if (!allowStyles.Contains(style))
 				{
 					int biomeTorchPlaceStyle = args.Player.SelectedItem.placeStyle;
 					{
@@ -2533,7 +2529,7 @@ namespace TShockAPI
 					var validCampfire = args.Player.SelectedItem.createTile == TileID.Campfire && biomeCampfirePlaceStyle == style;
 					if (!args.Player.TPlayer.unlockedBiomeTorches || (!validTorch && !validCampfire))
 					{
-						TShock.Log.ConsoleError(GetString("Bouncer / OnPlaceObject rejected object placement with invalid style {1} (expected {2}) from {0}", args.Player.Name, style, args.Player.SelectedItem.placeStyle));
+						TShock.Log.ConsoleError(GetString("Bouncer / OnPlaceObject rejected object placement with invalid style {1} (expected {2}) from {0}", args.Player.Name, style, string.Join(',', allowStyles)));
 						args.Player.SendTileSquareCentered(x, y, 4);
 						args.Handled = true;
 						return;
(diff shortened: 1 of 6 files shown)


TShockAPI/Bouncer.cs
```diff
@@ -260,6 +260,19 @@ internal Bouncer()
 					}
 				});
 
+			PlaceStyleCorrectors.Add(TileID.Grass, // it's TileID.ImmatureHerbs actually...
```
hakusaro, line 263: What does this comment mean?

TShockAPI/Bouncer.cs
```diff
@@ -1116,54 +1188,51 @@ internal void OnItemDrop(object sender, GetDataHandlers.ItemDropEventArgs args)
 			if (type < -48 || type >= Terraria.ID.ItemID.Count)
 			{
 				// Causes item duplications. Will be re added later if necessary
-				//args.Player.SendData(PacketTypes.ItemDrop, "", id);
+				//args.Player.SendData(PacketTypes.SyncItemDespaw, "", id);
```
hakusaro: Why is this changed, if it's a comment, and this is a typo?

TShockAPI/Bouncer.cs
```diff
@@ -1116,54 +1188,51 @@ internal void OnItemDrop(object sender, GetDataHandlers.ItemDropEventArgs args)
...
-					args.Handled = true;
-					return;
-				}
-
-				args.Handled = false;
-				return;
-			}
+			// Item removed by PacketTypes.SyncItemDespawn now
+			// if (type == 0)
+			// {
```
hakusaro: Why is this commented out?

TShockAPI/Bouncer.cs
```diff
@@ -2089,6 +2158,12 @@ internal void OnNPCAddBuff(object sender, GetDataHandlers.NPCAddBuffEventArgs ar
 
 			if (!args.Player.HasPermission(Permissions.ignorenpcbuffdetection))
 			{
+				// "TorchSlime" will add a super long burn buff
+				if (type == BuffID.OnFire && Main.getGoodWorld && npc.aiStyle == 1 && (short)npc.ai[1] == ItemID.Torch)
```
hakusaro: Is there any way this can be non-magic numbers?

TShockAPI/Bouncer.cs
```diff
@@ -2990,34 +3075,39 @@ internal static int GetMaxPlaceStyle(int tileID)
...
-			{ BuffID.GelBalloonBuff, 1800  },       // BuffID: 320
-			{ BuffID.OnFire3, 1200 },               // BuffID: 323
-			{ BuffID.Frostburn2, 1200 },            // BuffID: 324
-			{ BuffID.TentacleSpike, 540 },          // BuffID: 337
-			{ BuffID.BloodButcherer, 540 },         // BuffID: 344
-			{ BuffID.Shimmer, 100 },		        // BuffID: 353
+			{ BuffID.Shimmer, 100 },
+			{ BuffID.Venom, 1800 },
+			{ BuffID.CursedInferno, 600 },
+			{ BuffID.OnFire, 19392 }, // It's supposed to be 600, but it's actually 19392. And I can't find 19132 anywhere.
```
hakusaro: Who says it's supposed to be 600, why is it 19392, and what do you mean by you can't find 19392 anywhere?

TShockAPI/GetDataHandlers.cs
```diff
@@ -2849,14 +2850,14 @@ private static bool HandleSpawn(GetDataHandlerArgs args)
 			if (Main.ServerSideCharacter)
 			{
 				// As long as the player has not changed his spawnpoint since initial connection,
-				// we should not use the client's spawnpoint value. This is because the spawnpoint 
-				// value is not saved on the client when SSC is enabled. Hence, we have to assert 
-				// the server-saved spawnpoint value until we can detect that the player has changed 
+				// we should not use the client's spawnpoint value. This is because the spawnpoint
```
hakusaro, line 2915: unnecessary changes contaminate the diff and slow the review process down by making it larger and harder to understand and change. in the future, please consider not changing arbitrary / random things like spaces after comments

hakusaro (changes requested):

TShockAPI/Bouncer.cs
ACaiCat: Removed, the duration of this buff is the same as the maximum OnFire buff duration.

TShockAPI/Bouncer.cs
ACaiCat: In the FTW world, when adding a burning buff to a slime carrying a torch, the integer 216000 is cast to ushort, causing an overflow, so the result becomes 19392 (`NPC.AI_001_Slimes()`). However, I noticed that non-slime creatures also trigger this. In fact, the value 19392 was previously obtained based on player feedback.600 is the maximum value I found in the Terraria source code besides 216000.

TShockAPI/Bouncer.cs
ACaiCat: The key of `PlaceStyleCorrectors` is `selectedItem.createTile` (`actualTileToBeCreated`). In this case, it is `TileID.Grass`, but the requested tile is `TileID.ImmatureHerbs`.

TShockAPI/Bouncer.cs
ACaiCat: Now, when a client removes a dropped item, it will send the `PacketTypes.SyncItemDespawn` packet. At the same time, clients should no longer send the `PacketTypes.ItemDrop` packet with `item.type` set to 0. This logic should be moved to a new method. However, I'm not sure if there are any special cases that need to be handled

TShockAPI/Bouncer.cs
ACaiCat: We should now use `PacketTypes.SyncItemDespawn` to remove dropped items, so I think we should update the comment.

TShockAPI/Bouncer.cs
ACaiCat: My mistake

TShockAPI/GetDataHandlers.cs
ACaiCat: Sorry, it seems my IDE automatically removed those strange spaces

TShockAPI/Bouncer.cs
hakusaro: We can remove the comment perhaps?

hakusaro (approved):

TShockAPI/Bouncer.cs
```diff
@@ -854,6 +854,15 @@ containerTile.type is not (TileID.ClayPot or TileID.RockGolemHead or TileID.Plan
 							}
 						}
 					}
+
+					// Acorn Slingshot can only place saplings, reject any other tile placement
```
lost-werewolf: This implies the item itself can place acorns. Shouldn't the existance of an acorn projectile at the requested coordinates be checked for instead? Similar to how rope placement from rope coils is checked. Checking held item can cause incorrect rejection if the player switches items before the acorn lands.

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -369,13 +369,15 @@ private MatchResult MatchRemoval(TSPlayer player, TileRect rect)
 		private static readonly TileRectMatch[] Matches = new TileRectMatch[]
 		{
 			TileRectMatch.Placement(2, 3, TileID.TargetDummy, 54, 36, 18, 18),
```
ZakFahey, line 371: Magic numbers in these args aren't very readable.

TShockAPI/Bouncer.cs
```diff
@@ -799,27 +825,73 @@ internal void OnTileEdit(object sender, GetDataHandlers.TileEditEventArgs args)
 						args.Handled = true;
 						return;
 					}
+					// Handle placement action for Regrowth tools to ensure they only replant herbs on valid containers.
```
ZakFahey, line 829: This and maybe other existing code ought to go in a helper method.

TShockAPI/Bouncer.cs
```diff
@@ -2394,20 +2462,56 @@ internal void OnPlaceObject(object sender, GetDataHandlers.PlaceObjectEventArgs
 					return;
 				}
 			}
+			else if (type == TileID.KiteAnchor)
+			{
```
ZakFahey, line 2459: Also here. This method is overall too long.

hakusaro:
@ZakFahey @lost-werewolf I'm inclined to merge this with the intent of revising it after e.g., improving code quality of Bouncer in general but I think this is probably a minimum viable changeset for initial support. If you have objections, please raise them in the next 24 hours from this post.

lost-werewolf:
No objections, looks good to me for now.

TShockAPI/Handlers/SendTileRectHandler.cs
ZakFahey: Also everywhere else?

ZakFahey:
No objections on the functionality.

PR #3187, opened 2026-02-21
Bump TSAPI + Configurable Chat Message Length

lost-werewolf:
This PR removes the fixed 500 character limit on chat messages and instead uses a config option. Also bumps TSAPI.
Changelog:
- Removed fixed 500 character chat message length. Servers can now configure it.
- Added ``TruncateExcessiveChatMessages``, which will truncate a message which exceeds the ``MaximumChatMessageLength`` instead of rejecting it outright.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -2906,19 +2906,28 @@ namespace TShockAPI
 			// This was formerly marked as a crash check; does not actually crash on this specific packet.
 			if (playerDeathReason != null)
 			{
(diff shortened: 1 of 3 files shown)


lost-werewolf:
Would like to know if handling for excessive death text length & rejecting custom death messages should be moved out of the Bouncer so we can run all relevant logic on rejected deaths (since I opted to instead kill the player with the default death reason if rejecting).

lost-werewolf:
Also apologies, I ended up bumping TSAPI in my local repo and it kept nagging me to commit it, so I included it here in this PR

hakusaro (approved):

PR #3188, opened 2026-02-22
Add voice variant and voice pitch offset

bryldd:
Set player's voiceVariant and voicePitchOffset in HandlePlayerInfo


lost-werewolf (commented):
Good catch on this, players can change their voice at a Dresser in-game.

lost-werewolf (approved):

hakusaro (approved):

PR #3189, opened 2026-02-23
Fix missing initialization of favorited state in NetItem

ACaiCat:
**Fix favorited state tracking**


AniAnanas (approved):

lost-werewolf (approved):
Looks good as-is.

hakusaro (approved):

PR #3193, opened 2026-02-24
Filter out [ct:xxx] tags to prevent PE client crashes

ACaiCat:
Invalid `[ct:xxx]` tags can crash the PE client.

**Filtered `[ct:xxx]` tags from chat messages to prevent mobile client crashes**


lost-werewolf:
Only malformed tags can cause crashing, such as invalid item ID or unfinished tag. They are legitimate tags that show up fine when not invalid (example: [ct:7,757]). And some of them don't respect the item ID after the comma, either. Last I remember there were about 20 total variants of this tag, from 0-19.

It would be good to report this to the mobile devs. I do have my problems with blocking them outright, but do not object to blocking malformed ones specifically.

hakusaro:
@ACaiCat can you write a filter for the invalid values? Or add a config option for this if you don’t change it?

ACaiCat:
> @ACaiCat can you write a filter for the invalid values? Or add a config option for this if you don’t change it?  

It might be too complex to write and maintain a filter that targets invalid values. I think we should block all [ct:] tags by default for now, until the mobile team fixes the underlying crash issue.

hakusaro (approved):

PR #3197, opened 2026-02-25
Improvements to TSPlayer.Teleport, more coordinate utiils, Spawn command respects team-based spawn seed

lost-werewolf:
- Most uses of ``TSPlayer.Teleport`` are now aligned to the player's bottom position, working around mounts changing player size
- ``TSPlayer.Teleport`` now uses ``RemoteClient.CheckSection`` when teleporting player to hopefully prevent cases where a player falls through blocks due to tiles not being loaded (because ``SendTileSquareCentered`` wasn't enough, apparantly)
- New overloads for ``TSPlayer.Teleport`` that take tile coordinates (Point) or world coordinates (Vector2), and can specify teleporting the player via their bottom position
- New position properties for TSPlayer, with some specifically working around mounts
- Spawn command now supports team-based spawns seed, added ``TeleportToWorldSpawn`` method to only teleport player to proper spawn point
- Adding a warp no longer places it incorrectly when using mounts that affect the player's hitbox size
- Teleporting to a region now teleports the player via their center position

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -585,10 +585,7 @@ namespace TShockAPI
 							args.Player.SendErrorMessage(GetString("You need to rejoin to ensure your trash can is cleared!"));
 						}
 
-						// ??
-						var lastTileX = args.Player.LastNetPosition.X;
-						var lastTileY = args.Player.LastNetPosition.Y - 48;
-						if (!args.Player.Teleport(lastTileX, lastTileY))
+						if (!args.Player.Teleport(args.Player.LastNetPosition))
 						{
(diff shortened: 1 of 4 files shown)


lost-werewolf:
Apologies with the 4 pointless commits. Mildly irritated by those but couldn't remove them even when re-creating the branch. If it would be preferable for me to simply make a new fork & migrate this PR to said new fork, I can do so

hakusaro:
Use rebase?

lost-werewolf:
Don't see the option to. Will just redo this on a fresh fork for now to avoid headaches, I haven't slept yet and it's almost 8 AM here.

PR #3198, opened 2026-02-25
Improvements to TSPlayer.Teleport, more coordinate utiils

lost-werewolf:
- Most uses of ``TSPlayer.Teleport`` are now aligned to the player's bottom position, working around mounts changing player size
- ``TSPlayer.Teleport`` now uses ``RemoteClient.CheckSection`` when teleporting player to hopefully prevent cases where a player falls through blocks due to tiles not being loaded (because ``SendTileSquareCentered`` wasn't enough, apparantly)
- New overloads for ``TSPlayer.Teleport`` that take tile coordiantes (Point) or world coordinates (Vector2), which can also target the bottom position of the player
- New position properties for TSPlayer, with some specifically working around mounts
- Spawn command now supports team-based spawns seed, added ``TeleportToWorldSpawn`` method to only teleport player to proper spawn point

(This time, without the headache of an improperly based fork)

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -586,9 +586,7 @@ namespace TShockAPI
 						}
 
 						// ??
-						var lastTileX = args.Player.LastNetPosition.X;
-						var lastTileY = args.Player.LastNetPosition.Y - 48;
-						if (!args.Player.Teleport(lastTileX, lastTileY))
+						if (!args.Player.Teleport(args.Player.LastNetPosition))
 						{
 							args.Player.Spawn(PlayerSpawnContext.RecallFromItem);
 						}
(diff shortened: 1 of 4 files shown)


lost-werewolf:
Extra notes:
This changes warp positioning slightly. While the actual tile-based position remains the same, the player is no longer shifted to the right and is instead centered on the warp's tile.
The /tppos command is unchanged, in case anything took advantage of the way it was aligned.

hakusaro:
>  This changes warp positioning slightly. While the actual tile-based position remains the same, the player is no longer shifted to the right and is instead centered on the warp's tile.

I feel like servers that have particularly well-designed warps will be thrown off by this change?

lost-werewolf:
> I feel like servers that have particularly well-designed warps will be thrown off by this change?

Reverted the X positioning, should be consistent with previous behavior without affecting the Y position change.

hakusaro (approved):

PR #3201, opened 2026-02-26
Check SyncLoadout only when connection status is complete

THEXN:


hakusaro:
@THEXN why were players being disabled when this occurred on your server?

THEXN:
> @THEXN why were players being disabled when this occurred on your server?

PixPin_2026-02-26_21-28-28
When the server has 'RequireLogin = true' and SSC is disabled:
As shown in the screenshot for players A and B:When Player B joins the game using Loadout 2, the debug log shows:
`Debug: GetDataHandlers / HandleSyncLoadout rejected loadout index sync 肝帝熙恩射手`
After Player B switches to Loadout 1 and then back to the original loadout, from Player A’s perspective, Player B appears to have no armor, accessories, or dyes equipped, and also suffers from obvious movement lag.

hakusaro:
I think this is fine for now, but it does make me question the check in the first place: https://github.com/Pryaxis/TShock/pull/2695

hakusaro (approved):

PR #3205, opened 2026-02-26
Packet hot-path optimizations and deterministic projectile cleanup

Xekep:
## Summary
- optimize handler dispatch using copy-on-write snapshots
- make projectile cleanup deterministic in second-update path

## Why
Reduces lock/iteration overhead in packet/handler paths while preserving plugin compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,11 @@ Thumbs.db
 *.user
 .vs/*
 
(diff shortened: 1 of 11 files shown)


hakusaro:
@Xekep thank you for your contributions but realistically we cannot merge any of these PRs because they are too big. if the size is over size/M and you change random unrelated lines it is too difficult to review.

PR #3207, opened 2026-02-26
Staggered stack-check scheduling to smooth second-update load

Xekep:
## Summary
- stagger stack-hack scans across ticks
- optimize second-update stack-check access path
- fix duplicate stack-check timestamp declaration

## Why
Spreads anti-cheat work over time and reduces periodic spikes during gameplay.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,11 @@ Thumbs.db
 *.user
 .vs/*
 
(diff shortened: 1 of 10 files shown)


hakusaro:
Please put one set of changes per PR. It is very difficult to review large changesets. If you’re pulling from a fork you can make a new branch on tshock and use git cherry-pick and pluck individual commits.

Issue #3180, opened 2026-02-16
In-game command to switch evils


Kojirremer:
Currently there is a seed (Drunk world) that switches the world's internal evil flag between corruption and crimson every sunrise. That is great for getting loot specific for corrupt/crimson-type worlds. Unfortunately, these worlds also force on you the Terraria OW soundtrack, which I'm not really a fan of.

So a command that toggles evil biomes would be nice. For instance, using /evil in a corrupt world would toggle it into a crimson one. I don't mean changing any world tiles, just the flag the world uses to identifie itself. 

Since there are just two evils, just a toggle should be enough IMO. Maybe just using
/evil
could show the user the current type of world, whereas
/evil switch
could change the world's evil to its alternative.

lost-werewolf:
Added as a basic toggle in #3208, cheers and thanks for the suggestion!

PR #3208, opened 2026-02-26
Evil switch command

lost-werewolf:
Added a command to switch the current world evil (#3180).


hakusaro:
#3180 wants /evil though

hakusaro (approved):

PR #3212, opened 2026-02-27
[perf] Make projectile cleanup deterministic in OnSecondUpdate

Xekep:
## Summary
- run recent projectile cleanup directly in second-update path
- remove stale entries in place with RemoveAll under lock

## Scope
- focused change in Bouncer.OnSecondUpdate only
- no unrelated file changes


hakusaro:
@greptile review

PR #3213, opened 2026-02-27
[perf] Stagger stack-hack scans to smooth second-update load

Xekep:
## Summary
- add per-player stack-scan timestamp state
- schedule stack checks at 5s intervals instead of every second per player
- initialize timestamps with player-index offset to spread work

## Scope
- focused changes in TSPlayer and TShock.OnSecondUpdate`n- no unrelated file changes


hakusaro:
@greptile review

PR #3215, opened 2026-02-27
[refactor/perf] Deduplicate HasHackedItemStacks slot checks

Xekep:
## Summary
- remove large duplicated slot-check branches in HasHackedItemStacks`n- keep equivalent checks via a single helper and per-container loops

## Scope
- focused change in TSPlayer.HasHackedItemStacks only
- no unrelated file changes

### DIFF
--- a/TShockAPI/TSPlayer.cs
+++ b/TShockAPI/TSPlayer.cs
@@ -486,71 +486,81 @@ namespace TShockAPI
 			Item[] loadout3Dye = TPlayer.Loadouts[2].Dye;
 			Item trash = TPlayer.trashItem;
 
-			bool CheckSlotStack(Item item, string warningMessage, bool checkNegative = true)
+			var itemWarning = GetString("Stack cheat detected. Remove item {0} ({1}) and then rejoin.");
+			var armorWarning = GetString("Stack cheat detected. Remove armor {0} ({1}) and then rejoin.");
+			var dyeWarning = GetString("Stack cheat detected. Remove dye {0} ({1}) and then rejoin.");
+			var itemDyeWarning = GetString("Stack cheat detected. Remove item dye {0} ({1}) and then rejoin.");
+			var piggyWarning = GetString("Stack cheat detected. Remove piggy-bank item {0} ({1}) and then rejoin.");
+			var safeWarning = GetString("Stack cheat detected. Remove safe item {0} ({1}) and then rejoin.");
+			var trashWarning = GetString("Stack cheat detected. Remove trash item {0} ({1}) and then rejoin.");
+			var forgeWarning = GetString("Stack cheat detected. Remove Defender's Forge item {0} ({1}) and then rejoin.");
+			var voidWarning = GetString("Stack cheat detected. Remove Void Vault item {0} ({1}) and then rejoin.");
+			var loadout1Warning = GetString("Stack cheat detected. Remove Loadout 1 item {0} ({1}) and then rejoin.");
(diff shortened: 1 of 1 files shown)


TShockAPI/TSPlayer.cs
```diff
@@ -486,335 +484,73 @@ public bool HasHackedItemStacks(bool shouldWarnPlayer = false)
...
-				else if (i < NetItem.Loadout3Dye.Item2)
-				{
-					var index = i - NetItem.Loadout3Dye.Item1;
-					Item item = new Item();
-					if (loadout3Dye[index] != null && loadout3Dye[index].type != 0)
-					{
-						item.netDefaults(loadout3Dye[index].type);
-						item.Prefix(loadout3Dye[index].prefix);
-						item.AffixName();
+			CheckSlotStack(trash, "Stack cheat detected. Remove trash item {0} ({1}) and then rejoin.", checkNegative: false);
```
ACaiCat: Why skip the negative stack check for the trash item?

TShockAPI/TSPlayer.cs
```diff
@@ -486,335 +484,73 @@ public bool HasHackedItemStacks(bool shouldWarnPlayer = false)
 			Item[] loadout2Dye = TPlayer.Loadouts[1].Dye;
 			Item[] loadout3Armor = TPlayer.Loadouts[2].Armor;
 			Item[] loadout3Dye = TPlayer.Loadouts[2].Dye;
-
 			Item trash = TPlayer.trashItem;
-			for (int i = 0; i < NetItem.MaxInventory; i++)
+
+			bool CheckSlotStack(Item item, string warningMessage, bool checkNegative = true)
```
ACaiCat: The return value doesn't seem to be used

TShockAPI/TSPlayer.cs
```diff
@@ -486,335 +484,73 @@ public bool HasHackedItemStacks(bool shouldWarnPlayer = false)
...
-						item.AffixName();
-						if (dye[index].stack > item.maxStack || dye[index].stack < 0)
-						{
-							check = true;
-							if (shouldWarnPlayer)
-							{
-								SendErrorMessage(GetString("Stack cheat detected. Remove dye {0} ({1}) and then rejoin.", item.Name, dye[index].stack));
-							}
-						}
+						SendErrorMessage(GetString(warningMessage, item.Name, item.stack));
```
ACaiCat: `GetString` should be used on the raw strings directly for strings extraction, not inside the helper method.

Xekep: Thanks, fair points.

1) About `checkNegative: false` for trash:
I left it this way on purpose to keep old behavior. Before refactor, trash only checked `stack > maxStack` (no negative check). I added a comment to make that explicit. If you want, I can change this too, but I’d prefer a separate PR since that’s behavior-changing.

2) About `CheckSlotStack` return value:
Yep, agreed - leftover from the first refactor. I changed it to `void` in `85e1f323`.

3) About `GetString`:
Also agreed. I moved messages to direct `GetString("...")` calls on raw literals, then pass the localized format into the helper. No nested `GetString` now. Included in `85e1f323`.

pacenadhif778:
i try your all pr and i get bug when use this pr player stuck in requesting tile data

pacenadhif778:
no plugin installed (need someone to test  it too) to validate my issue

hakusaro:
@greptile review

PR #3216, opened 2026-02-27
[perf] Clear per-second placement rollback cache

Xekep:
## Summary
- clear TilesCreated when TilePlaceThreshold is reset each second
- keep placement rollback tracking bounded to active threshold window

## Scope
- focused change in TShock.OnSecondUpdate only
- no unrelated file changes


hakusaro:
@greptile review

PR #3217, opened 2026-02-27
[perf] Prevent per-packet OnPlayerUpdate log spam

Xekep:
## Summary
- treat zero LastNetPosition as uninitialized baseline
- initialize from first valid update packet instead of logging each time

## Scope
- focused change in Bouncer.OnPlayerUpdate only
- no unrelated file changes


hakusaro:
@greptile review?

PR #3218, opened 2026-02-27
Fix SSC login deadlock when players are crowd controlled before auth

Xekep:
## Summary
- allow /login while CCed only when player is in SSC pre-login disabled state (IsDisabledForSSC)
- prevents login deadlock where server-applied webbed state blocks authentication

## Scope
- focused change in Commands.AttemptLogin only
- no unrelated file changes

### DIFF
--- a/TShockAPI/Commands.cs
+++ b/TShockAPI/Commands.cs
@@ -846,7 +846,9 @@ namespace TShockAPI
 				return;
 			}
 
-			if (args.TPlayer.CCed && Main.ServerSideCharacter)
+			// SSC pre-login flow keeps players disabled (webbed) until authentication.
+			// Do not block login for that temporary disabled state, or players can get stuck.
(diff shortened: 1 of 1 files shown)


hakusaro:
@greptile review?

hakusaro:
@greptile i have my doubts as to whether or not this is reachable in practice?

Xekep:
> @greptile i have my doubts as to whether or not this is reachable in practice?

This is reachable in practice; I reproduced it on a survival server with SSC enabled.

Steps:
1) Join without being logged in.
2) Player is SSC-disabled and gets Webbed (CC).
3) Run `/login <password>`.

Before this fix, login is rejected with “You cannot login whilst crowd controlled.”, causing an auth lock loop.  
The fix only bypasses this check when CC comes from SSC disable (`IsDisabledForSSC`), so normal CC behavior remains unchanged.

hakusaro:
I think we've effectively fixed the hole causing this, can you check if this patch is still necessary?

PR #3221, opened 2026-02-27
Add more AI value limits to certain projectiles

lost-werewolf:
Limit AI values on certain projectiles that require it.

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -3211,12 +3211,14 @@ namespace TShockAPI
 		private Dictionary<short, float> Projectile_MinValuesAI = new Dictionary<short, float> {
 			{ 611, -1 },
 
-			{ 950, 0 }
+			{ 950, 0 },
+			{ 502, 0 }
 		};
 		private Dictionary<short, float> Projectile_MaxValuesAI = new Dictionary<short, float> {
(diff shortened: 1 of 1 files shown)


hakusaro (commented):
How do we know what these numbers are? It would be better to tether them to a game value reference or use an enum.

lost-werewolf:
These numbers come from some analysis of how these projectiles are handled by the game:
Projectile ID 502 never exceeds ai[0] of 5 (technically, it does, but it's killed right before this happens) and always starts with ai[0] of 0. ai[0] can only increase, so negatives are not possible for this one.
Projectile ID 459 can only have ai[1] of 0.7f, 1.0f, and 1.3f (this is unused, but still present in the code).

lost-werewolf:
Providing screenshots from decompiled code here.
502 doesn't exceed ai[0] of 5 before being killed
image

459 can only have 0.7f, 1.0f, or 1.3f (unused) as ai[1]:
image
image

hakusaro:
So it would be better to have these in code, at the bare minimum as comments.

lost-werewolf:
Can do. The other projectiles lacked them so I didn't add any, but I'll add comments to the ones I added.

hakusaro (approved):

PR #3226, opened 2026-02-28
Fix first SSC login appearance reset for new players

Xekep:
Players reported that with SSC enabled, first-time players can get default appearance values after they log in.

This patch keeps SSC behavior the same, but if the seeded tsCharacter row has missing appearance fields, it fills them from the current player data at login and reloads PlayerData.

It is applied in all login paths (UUID login, password login, and /login).


lost-werewolf:
Was this an issue prior to #3199 ? Just curious.

hakusaro (approved):

PR #3233, opened 2026-02-28
Add Display Jar tile rect match and placement validation

ACaiCat:
**Fix Display Jar placement**

### DIFF
--- a/TShockAPI/Bouncer.cs
+++ b/TShockAPI/Bouncer.cs
@@ -3016,7 +3016,7 @@ namespace TShockAPI
 				return;
 			}
 
-			if (!args.Player.IsInRange(args.TileX, args.TileY, range: 13)) // To my knowledge, max legit tile reach with accessories.
+			if (!args.Player.IsInRange(args.TileX, args.TileY))
 			{
 				TShock.Log.ConsoleDebug(GetString("Bouncer / OnFoodPlatterTryPlacing rejected range checks from {0}", args.Player.Name));
 				args.Player.SendTileSquareCentered(args.TileX, args.TileY, 1);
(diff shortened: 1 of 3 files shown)


lost-werewolf:
This seems good. Could you also fix the Bouncer rejecting item removal from the jar when not using a pickaxe before I approve? This is about the same case as torches, right clicking removes the item without pickaxing the jar. Feel it would be good to squash here along with this match.

Edit: Attached image
image

Double edit: attached clip:
https://github.com/user-attachments/assets/1f692fca-02c7-4d5c-a1de-dd747cfe405c

ACaiCat:
Maybe we need a new GetDataHandler...
image

lost-werewolf:
>Maybe we need a new GetDataHandler...

If every packet sent by clients had a handler AND a hook to subscribe to, it would be great. Noticed once when making a plugin that I couldn't subscribe to the SpawnBoss handler, so I had to manually do so through OnGetData. Perhaps good for another PR.

ACaiCat:
@greptile-apps review

TShockAPI/Bouncer.cs
```diff
@@ -3013,6 +3025,58 @@ internal void OnFoodPlatterTryPlacing(object sender, GetDataHandlers.FoodPlatter
...
+				TShock.Log.ConsoleDebug(GetString("Bouncer / OnDisplayJarTryPlacing rejected permissions from {0}", args.Player.Name));
+				Item item = new Item();
+				item.netDefaults(args.ItemID);
+				args.Player.GiveItemCheck(args.ItemID, item.Name, args.Stack, args.Prefix);
+				args.Player.SendTileSquareCentered(args.TileX, args.TileY, 1);
+				args.Handled = true;
+				return;
+			}
+
+			if (!args.Player.IsInRange(args.TileX, args.TileY, range: 13)) // To my knowledge, max legit tile reach with accessories.
```
lost-werewolf: Thiss doesn't account for journey mode's extended reach. Perhaps better to simply leave the range argument as default.

lost-werewolf (changes requested):

TShockAPI/Bouncer.cs
ACaiCat: Removed

lost-werewolf (approved):
LGTM

PR #3237, opened 2026-03-02
fix: team change message isn't broadcast on spawn

lost-werewolf:
- Fix player spawn bypassing pvp with no team setting
- Prevent malicious client from fast-switching team with spawn packet
- Split up a check in HandlePlayerTeam to accomodate for team changes

As a side effect of the change to handling of team in the ``PlayerSpawn`` packet, it now must handle the vanilla logic on its own, to prevent malicious clients from changing their team too fast, and to also prevent this from bypassing the "pvpwithnoteam" pvp mode.

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -2907,6 +2907,7 @@ namespace TShockAPI
 			if (pvpMode == "pvpwithnoteam" && team != 0)
 			{
 				team = 0;
+				args.TPlayer.team = 0; // make sure to set it to 0 (no team)
 				teamCorrectNeeded = true;
 			}
 			// Malicious client likely trying to fast switch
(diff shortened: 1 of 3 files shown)


lost-werewolf:
@greptile review?

lost-werewolf:
The edge case should be resolved by [55a819d](https://github.com/Pryaxis/TShock/pull/3237/commits/55a819d45306879a9f526d7f4b01e0f40ca8d6b3), where I set the player's team to 0. Since it's already 0, there should be no edge case. Tested with a 3rd-party client using code execution to run code that sets my team and then spawns my player, I could not replicate anything.
To also fix another bug regarding PVP mode being switched while the server is running, and to prevent any more possible issues, ``TShock.OnSecondUpdate`` now checks it and appropriately sets the pvp status and team of the player.

hakusaro:
@greptile review

lost-werewolf:
I don't really see the edge case here. Since the player's team is always set to 0 when using ``pvpwithnoteam`` as the pvp mode now, it shouldn't ever be able to not be 0 when team correction is needed. Especially since now, every second, the player's team is checked and corrected to 0 if it's not.

hakusaro:
Please do not use conventional commits.

lost-werewolf:
My apologies. if you can clarify what you mean by that, I will avoid it in the future.

lost-werewolf:
Better off closing and redoing this PR to squash everything into one commit?

lost-werewolf:
Closing for now, will redo this in the near future when I have time.

Issue #3239, opened 2026-03-02
Allow the SaveWorker to break save attempt on timeout for a completed task


T3mp3lis:
**Is your feature request related to a problem? Please describe.**
The way SaveWorker is made right now, if the save task doesnt complete for any reason, the lock is never released, causing the server to hang. Players are booted from the server, console is unresponsive. If using docker, it cant even kill the process. I suspect it's the busy io delay in my setup that causes the issue, but having an option for a timeout on save attempt seems reasonable. 

**Describe the solution you'd like**
Add a 30-60 sec timeout for save task, and if it runs out, break the save attempt with an exception. It should not break the auto-saving fucntionality, since we reset the timer for each consecutive save in the queue anyway

**Describe alternatives you've considered**
I have made a wrapper script that pauses auto-saves through restapi during high load on my disk, but I couldnt predict all the spikes that my setup could have

**Additional context**
None

PR #3240, opened 2026-03-04
prevent SaveWorker lock from hanging server on slow IO

Reisenbug:
Addresses #3239 
`_saveLock` was previously held for the entire `WorldFile.SaveWorld()` call. If disk IO failed, the lock was never released, and SaveWorld(wait: true) would deadlock polling `saveQueueCount` through the same lock
Lock now only protects queue dequeue. A `volatile _isSaving` flag lets `SaveWorld(wait: true)` wait without acquiring the lock
#3239 Timeout not implemented: `WorldFile.SaveWorld()` has no `CancellationToken` support, so a timeout could only abandon the wait while IO continues, risking world file corruption if a new save starts concurrently

### DIFF
--- a/TShockAPI/SaveManager.cs
+++ b/TShockAPI/SaveManager.cs
@@ -42,6 +42,7 @@ namespace TShockAPI
 		private Object _saveLock = new Object();
 		private Queue<SaveTask> _saveQueue = new Queue<SaveTask>();
 		private Thread _saveThread;
+		private volatile bool _isSaving = false;
 		private int saveQueueCount { get { lock (_saveLock) return _saveQueue.Count; } }
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


Reisenbug:
@greptile review

T3mp3lis:
> https://github.com/Pryaxis/TShock/issues/3239 Timeout not implemented: WorldFile.SaveWorld() has no CancellationToken support, so a timeout could only abandon the wait while IO continues, risking world file corruption if a new save starts concurrently

I see, I tried fixing the issue myself, looks like unsuccessfully - 28dd40e3f09ff1b1442f21d39ec27aa0e52847cd

EDIT:
Though I couldn't really test it out, everyone who plays regularly has been busy and couldn't stress test with me. Triggering saves over and over did not cause the hang.

PR #3246, opened 2026-03-05
Add "keep player appearance" option to SSC

lost-werewolf:
This PR adds an option to SSC that retains the player's offline character appearance.
This does **NOT** keep any local items, just character data such as hair, skin color, clothing style, etc.


lost-werewolf:
Need feedback on if I should I include ``hairDye`` as local character appearance before doing so

ACaiCat (approved):

lost-werewolf:
> Need feedback on if I should I include `hairDye` as local character appearance before doing so

Still want to know if I should count the player's `hairDye` as "local character appearance" before I make any changes. Until then this PR isn't technically finished in my eyes

hakusaro:
It sounds like it to me.

lost-werewolf:
> It sounds like it to me.

Will do in a minute.

PR #3251, opened 2026-03-07
Fix stone wall conversion

lost-werewolf:
Fixes the ``WorldGenMock`` conversion for walls to replace uses of ``Stone`` (1) with ``StoneUnsafe`` (349) as 1.4.5 no longer use safe stone walls in conversions, so purifying evils no longer gets incorrectly rejected

### DIFF
--- a/TShockAPI/Handlers/SendTileRectHandler.cs
+++ b/TShockAPI/Handlers/SendTileRectHandler.cs
@@ -1231,9 +1231,9 @@ namespace TShockAPI.Handlers
 					}
 					break;
 				case 7:
-					if ((WallID.Sets.Conversion.Stone[wall] || WallID.Sets.Conversion.Ice[wall] || WallID.Sets.Conversion.Sandstone[wall]) && wall != 349)
+					if ((WallID.Sets.Conversion.Stone[wall] || WallID.Sets.Conversion.Ice[wall] || WallID.Sets.Conversion.Sandstone[wall]) && wall != WallID.StoneUnsafe)
 					{
-						tile.wall = 349; // wall 349 (StoneUnsafe) found in WorldGen.Convert when looking for WallID.Sets.Conversion.Stone, replaces use of wall 1 (Stone)
+						tile.wall = WallID.StoneUnsafe;
 					}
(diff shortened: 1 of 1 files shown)


ACaiCat (approved):

TShockAPI/Handlers/SendTileRectHandler.cs
```diff
@@ -1231,9 +1231,9 @@ private static void Convert(MockTile tile, int k, int l, int conversionType)
 					}
 					break;
 				case 7:
-					if ((WallID.Sets.Conversion.Stone[wall] || WallID.Sets.Conversion.Ice[wall] || WallID.Sets.Conversion.Sandstone[wall]) && wall != 1)
+					if ((WallID.Sets.Conversion.Stone[wall] || WallID.Sets.Conversion.Ice[wall] || WallID.Sets.Conversion.Sandstone[wall]) && wall != 349)
```
hakusaro: It would be really optimal to not use magic numbers anywhere in the codebase anymore. Can you reference some other point of reference or enumeration that explains what 349 is?

hakusaro (changes requested):

TShockAPI/Handlers/SendTileRectHandler.cs
lost-werewolf: comments added in https://github.com/Pryaxis/TShock/pull/3251/commits/053d40f22ef6b9cc80cd7933b14b6eca56952681.

hakusaro: Comments are not ideal, just use code references?

hakusaro (changes requested):

TShockAPI/Handlers/SendTileRectHandler.cs
lost-werewolf: Would be ideal to use ``WallID.StoneUnsafe`` instead of just 349? Clarification would be good before I make another change

lost-werewolf: Thanks for clarifying on contribution guidelines, making requested changes now.

lost-werewolf:
Apologies, redoing without conventional commits since I now properly grasp the concept of them (it flew over my head previously).

PR #3253, opened 2026-03-07
Fix unsafe stone wall failing to match in conversions

lost-werewolf:
Redo of #3251 to properly follow contributing guidelines.

``WorldGenMock`` now uses ``StoneWallUnsafe`` instead of ``Stone`` when simulating conversions, as 1.4.5 now uses this wall instead.


ACaiCat (approved):
The logic is the same as in #3251

lost-werewolf:
> The logic is the same as in #3251

I remade this PR to follow the contributing guidelines better without a mess of commits. Hopefully, avoided conventional commits properly since I actually didn't understand what they were at first.

PR #3255, opened 2026-03-07
Revamp the Nuget publish workflow

ZakFahey:
- The workflow now runs every time a release is created, rather than when a commit is pushed into the nuget-release branch.
- If there's a mismatch between the release tag and the project tag, the release tag will take precedence. If such a version override happens, there will be a workflow warning.
- It is recommended that for pre-releases, you don't update the version number in the csproj file. This will ensure that UpdateManager.cs continues to work as expected (it does not currently handle version numbers with a prerelease suffix, and otherwise we'd have to add logic to not alert for prerelease updates). But once you have a mainline release, the version number should be updated.

The goal here is to eliminate all human error that would cause the Nuget package to ever not be updated. If there is a GitHub release, the Nuget release will always be there, including for pre-releases.

The utility of the Nuget package is greatly diminished if it isn't consistently up-to-date. I was planning soon to update a bunch of plugins to the latest pre-release, many of which currently reference TShock using the Nuget package. Having the benefits of a package reference at one point, and then being forced to go back to a binary reference later on, and then just continuing to switch back and forth, makes for a less ergonomic updating experience.

Changelog updates: N/A

### DIFF
--- a/.github/workflows/ci-otapi3-nuget.yml
+++ b/.github/workflows/ci-otapi3-nuget.yml
@@ -1,12 +1,11 @@
-name: Deploy NuGet(OTAPI3)
+name: Publish NuGet
 
 on:
-  push:
-    branches: [ nuget-release ]
+  release:
+    types: [published]
 
 jobs:
-  build:
-
+  publish:
     runs-on: ubuntu-latest
     environment: release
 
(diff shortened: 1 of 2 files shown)


hakusaro:
I don’t have any intention of changing the nuget release cycle.

hakusaro:
For starters, we create pre-releases that are unstable. Second, with the advent of AI LLMs, updating your references should be trivial.

ZakFahey:
Is there any particular reason to avoid having prerelease Nuget packages for TShock prereleases? If it's tagged as a prerelease in Nuget, developers would know and understand that it is unstable already.

Also, plugins that reference TShock would usually be okay to reference a prerelease, because the main reason for doing so is to update against the API breaking changes. Between prerelease and main release I probably wouldn't expect breaking changes at least for package referencing, so getting a head start at updating plugins with a prerelease is often an okay final state even after the main release with all its bugfixes is out.

hakusaro:
Your changes don't tag anything as a pre-release do they?

I would prefer we discuss this set of practices in-general. If you open a PR without even discussing it, I think it leads to counterproductive "fighting against the current".

1. I'm not opposed in-principle to nuget pre-releases, but it has never been asked of us.
2. We don't really have tags that make logical sense. Maybe the tags are fine for nuget? I'm not sure. Pre-release tagging and versioning wasn't standardized.
3. I prefer to have a manual flow ala what OTAPI does, because it gives us control over what is published versus what is not.

I also consider build staining a general pre-requisite to this. I would like it if every build we release is stamped with the git version as well, so we know exactly what is where. Because we do not have build staining, I am reluctant to release anything that is not an official release to nuget.

hakusaro:
If you're interested in adding build staining, we could add automatic publishing to nuget for pre-releases afterwards, provided they are tagged as such*.

PR #3267, opened 2026-03-12
Don't push attestations to the container registry

### LINKED ISSUE #3242: Docker image tags are confusing and lack clear version mapping
### Is your feature request related to a problem? Please describe.
I'm always frustrated when trying to use TShock Docker images because the tags are confusing and don't correspond to official release versions. Currently, there are many `sha256-` prefixed tags (e.g., `sha256-128bde7551c...`) that are not human-readable, and tags like `general-devel` and `i18n` provide no clear indication of which TShock version they contain. This makes it difficult to select stable versions for deployment or understand what version is actually running.

### Describe the solution you'd like
I would like the Docker image tagging to follow a clear and consistent scheme that matches official TShock releases:
- Each official release version should have a corresponding Docker tag (e.g., `5.9.9.0`, `5.9.8.0`, etc.)
- `latest` tag should always point to the most recent stable release
- `devel` or `nightly` tags can be maintained for development builds
- Avoid using `sha256-` prefixed tags as primary identifiers in the tag list

### Describe alternatives you've considered
- Using `general-devel` as a workaround, but this is unpredictable and may not be stable
- Building custom Docker images from source, which adds unnecessary complexity for users who just want to run the official version
- Parsing commit hashes to guess versions, which is impractical and error-prone

### Additional context
Screenshot of current confusing tags:

Docker CI screenshot

other docker issue

Cai lmao

Community discussions (see #3131 and #2698) indicate that other users are also experiencing confusion with the current tagging scheme. Improving this would greatly enhance the user experience for Docker-based deployments.

timschumi:
While cool, those will spam the tag listing with unreadable and empty sha256 tags.

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -57,4 +57,3 @@ jobs:
         with:
           subject-name: ghcr.io/${{ github.repository }}
           subject-digest: ${{ steps.build.outputs.digest }}
-          push-to-registry: true


ACaiCat (approved):
Looks good

PR #3271, opened 2026-03-14
Add Missing Packet Handler ( WeaponRack, SyncItemsWithShimmer, SyncItemCannotBeTakenByEnemies and EntityLeash ) 

NightKLP:
it's almost the same as Item Frame function but it works which  just notice why this packet doesn't exist yet

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
 TShock, a server mod for Terraria
 Copyright (C) 2011-2019 Pryaxis & TShock Contributors
 
@@ -16,20 +16,21 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+using Microsoft.Xna.Framework;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.IO.Streams;
 using System.Linq;
-using Microsoft.Xna.Framework;
 using Terraria;
 using Terraria.DataStructures;
 using Terraria.GameContent.Tile_Entities;
 using Terraria.ID;
 using Terraria.Localization;
-using TShockAPI.Models.PlayerUpdate;
 using TShockAPI.Configuration;
+using TShockAPI.Models.PlayerUpdate;
 
 namespace TShockAPI
 {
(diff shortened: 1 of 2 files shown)


lost-werewolf (approved):
This looks great, thanks for the catch on this!

ACaiCat (approved):
Works as expected

NightKLP:
In the past days ive been **exploited by hackers in my server** and decided to investigate it ive build a **packet logger** ( good thing my server can handle it had so much memory that can handle it )
hackers just **spawn platinum coin out of tin air** to players which the packet logs shows it was **( SyncItemsWithShimmer and SyncItemCannotBeTakenByEnemies )** apparently

the packets i coded and unsure is:
- HandleSyncItemCannotBeTakenByEnemies

while this handler is not fully finish will be:
- TELeashedEntityAnchorPlaceItem

TShockAPI/GetDataHandlers.cs
```diff
@@ -4974,6 +5153,13 @@ Tuple<int, int> GetDyeSlotsForLoadoutIndex(int index)
 			return false;
 		}
 
+		private static bool HandleSyncItemCannotBeTakenByEnemies(GetDataHandlerArgs args)
+		{
+			args.Player.Kick(GetString("Exploit attempt detected!"));
+			TShock.Log.ConsoleDebug(GetString($"HandleSyncItemCannotBeTakenByEnemies: Player is trying to sync item that cannot be taken by hostile mobs; this is a crafted packet! - From {args.Player.Name}"));
```
lost-werewolf: This packet is used by clients to spawn items when using the Classy Cane or when a Lucky Coin is equipped. Rejecting it breaks those items.

TShockAPI/GetDataHandlers.cs
```diff
@@ -2499,6 +2548,67 @@ private static bool OnFoodPlatterTryPlacing(TSPlayer player, MemoryStream data,
 			FoodPlatterTryPlacing.Invoke(null, args);
 			return args.Handled;
 		}
+		
+		/// <summary>
+		/// For use in an SyncItemsWithShimmer event
+		/// </summary>
+		public class SyncItemsWithShimmerEventArgs : GetDataHandledEventArgs
```
lost-werewolf, line 2555: Item Drop (21), SyncItemsWithShimmer (145), and SyncItemCannotBeTakenByEnemies (148) all use the same handling, just with extra data for each. It would probably be best if all 3 of these used the same code and shared the same args. The packet-specific args are simply defaulted if the packet being read isn't a specific type. Check the MessageBuffer to see how it's handled.

lost-werewolf (changes requested):
If you can make Item Drop (21), SyncItemsWithShimmer (145), and SyncItemCannotBeTakenByEnemies (148) use the same args (expanding the Item Drop args) & handling, this should be good once you've finished, as the previous changes were great.

TShockAPI/GetDataHandlers.cs
NightKLP: alright thanks 👍

PR #3275, opened 2026-03-16
add handling for ChestSizeSync packet

lost-werewolf:
This adds handling and a permission for the chest size sync packet (155), due to a reported potential memory leak being caused by "chests and items" in the Pryaxis discord.

image

~~ReLogic, why do you add these kinds of things and then don't make sure only the server can send them?~~

### DIFF
--- a/TShockAPI/GetDataHandlers.cs
+++ b/TShockAPI/GetDataHandlers.cs
@@ -4986,7 +4986,7 @@ namespace TShockAPI
 			short id = args.Data.ReadInt16();
 			short newSize = args.Data.ReadInt16();
 
-			if (id < 0 || id > 8000) // chest is invalid
+			if (id < 0 || id >= 8000) // chest is invalid
 				return true;
 
 			Chest chest = Main.chest[id];
@@ -4994,6 +4994,20 @@ namespace TShockAPI
 			if (chest == null)
 			{
 				TShock.Log.ConsoleDebug(GetString("GetDataHandlers / HandleChestSizeSync rejected from null chest {0}", args.Player.Name));
+				return true;
+			}
+
+			if (args.Player.IsBeingDisabled())
+			{
+				TShock.Log.ConsoleDebug(GetString("GetDataHandlers / HandleChestSizeSync rejected from disabled {0}", args.Player.Name));
+				args.Player.SendData(PacketTypes.SyncChestSize, "", id, chest.item.Length);
+				return true;
+			}
+
+			if (args.Player.IsBouncerThrottled())
+			{
+				TShock.Log.ConsoleDebug(GetString("GetDataHandlers / HandleChestSizeSync rejected from throttled {0}", args.Player.Name));
+				args.Player.SendData(PacketTypes.SyncChestSize, "", id, chest.item.Length);
 				return true;
 			}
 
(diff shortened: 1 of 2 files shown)


TShockAPI/GetDataHandlers.cs
```diff
@@ -4980,6 +4981,46 @@ private static bool HandleDisplayJar(GetDataHandlerArgs args)
...
+				args.Player.SendData(PacketTypes.SyncChestSize, "", id, chest.item.Length);
+				return true;
+			}
+
+			if (newSize < 0) // size is invalid
+			{
+				TShock.Log.ConsoleDebug(GetString("GetDataHandlers / HandleChestSizeSync rejected from invalid size {0}", args.Player.Name));
+				args.Player.SendData(PacketTypes.SyncChestSize, "", id, chest.item.Length);
+				return true;
+			}
```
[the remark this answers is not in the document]
lost-werewolf: The lower bound is checked to prevent exceptions being thrown. Intentionally left the upper bound unchecked because of genuine use case for this, such as a chest that contains every single item in the game.

ACaiCat:
I have a question，can a normal player send resize packet?

lost-werewolf:
> I have a question，can a normal player send resize packet?

No, hence why I decided to permission lock it and reject by default otherwise. It has genuine use cases for people who would want it, since it can be easier to do something on a modified game client rather than adding a command to a plugin & then running that. After all, there's some settings that do allow modified clients to do stuff (such as invalid place styles)

TShockAPI/GetDataHandlers.cs
```diff
@@ -4980,6 +4981,60 @@ private static bool HandleDisplayJar(GetDataHandlerArgs args)
 			return false;
 		}
 
+		private static bool HandleChestSizeSync(GetDataHandlerArgs args)
+		{
+			short id = args.Data.ReadInt16();
+			short newSize = args.Data.ReadInt16();
+
+			if (id < 0 || id >= 8000) // chest is invalid
```
ACaiCat:
```suggestion
			if (id < 0 || id >= Main.maxChests) // chest is invalid
```

TShockAPI/GetDataHandlers.cs
```diff
@@ -4980,6 +4981,60 @@ private static bool HandleDisplayJar(GetDataHandlerArgs args)
...
+
+			if (args.Player.IsBouncerThrottled())
+			{
+				TShock.Log.ConsoleDebug(GetString("GetDataHandlers / HandleChestSizeSync rejected from throttled {0}", args.Player.Name));
+				args.Player.SendData(PacketTypes.SyncChestSize, "", id, chest.item.Length);
+				return true;
+			}
+
+			if (!args.Player.HasPermission(Permissions.resizechests))
+			{
```
ACaiCat, line 5015: We should consider logging or kicking the player — they must be cheating

TShockAPI/Permissions.cs
```diff
@@ -515,6 +515,9 @@ public static class Permissions
 
 		[Description("Player can send emotes.")]
 		public static readonly string sendemoji = "tshock.sendemoji";
+
+		[Description("Player can resize chests. Warning: Dangerous permission to grant, very easy to abuse.")]
+		public static readonly string resizechests = "tshock.resizechests";
```
ACaiCat: May be we should add it to `world` or `ignore` node

ACaiCat (approved):

ACaiCat:
@greptile-apps review

PR #3276, opened 2026-03-17
Revert "Filter out [ct:xxx] tags to prevent PE client crashes"

ACaiCat:
Reverts Pryaxis/TShock#3193

Terraria PE v1.4.5.6 totally resolves the CT tag crash bug


lost-werewolf (approved):
Mobile v1.4.5.6 patches all invalid ControlsTouch tags from my testing. Hopefully for good this time.

PR #3287, opened 2026-04-17
Make /give and /item not give more items than needed

JustULTA:
This makes previously unstackable items still being given in stacks of 1 if amount is not specified (all other items are still given in max amounts by default). Also makes it consistent with vanilla's journey mode item duplication.


ACaiCat (approved):
Looks good

PR #3294, opened 2026-06-21
Disable UUID auto-login for superadmin group

### LINKED ISSUE #3278: disable autologin for superadmin group
players who have previously logged in with a SuperAdmin account are automatically authenticated via UUID when they rejoin the server
this autologin happens when a player uses the same character name (matching the account username) and the same UUID, allowing them to bypass manual authentication with /login. This behavior can be a security risk for high-privilege acc i would like to request a feature where players in the SuperAdmin group are required to manually authenticate using /login, even if their UUID matches the stored account UUID and their character name is the same.

AdvenRanises:
Fixes #3278. Prevents superadmin accounts from being automatically authenticated via UUID, requiring manual /login instead.

### DIFF
--- a/TShockAPI/Configuration/TShockConfig.cs
+++ b/TShockAPI/Configuration/TShockConfig.cs
@@ -385,6 +385,10 @@ namespace TShockAPI.Configuration
 		[Description("Prevents users from being able to login with their client UUID.")]
 		public bool DisableUUIDLogin;
 
+            /// <summary>Prevents users in the superadmin group from being automatically logged in via UUID.</summary>
+            [Description("Prevents users in the superadmin group from being automatically logged in via UUID.")]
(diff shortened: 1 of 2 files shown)


AdvenRanises:
Fixed Greptile feedback: added early return and SendInfoMessage for superadmin UUID block.

SakuraIsayeki:
So as to not break current functionality, could it be possible to make this change appear behind a feature flag or option ? While I agree that UUID login presents risks, there should be a best effort made to preserve existing flows whilst allowing gradual/at-will security hardening. 

Thoughts @hakusaro ?

AdvenRanises:
Updated to use a config option as requested. Added DisableSuperAdminUUIDLogin (default: true) in TShockConfig.cs. Server owners can opt out by setting it to false if they want to preserve the old UUID auto-login behavior for superadmin accounts.

PR #3299, opened 2026-08-20
Rough first pass at a 1.4.5.7 port

ZakFahey:
All untested claudeslop but this is what I had time to do late last evening and before leaving for work today. Trying to get a first pass at something that might work. If you submit none of it and just use it for reference it won't hurt my feelings.


hakusaro:
Can you add a decomp to Pryaxis/Sources?

ZakFahey:
Closing in favor of https://github.com/Pryaxis/TShock/pull/3301

